PR #2, opened 2018-10-16
sysstat: allow sysstat_t to getattr on mountpoints

yuvalturg:
The sadc utility from sysstat that collects file systems statistics (-F
flag) will be denied to access /var/log/audit (getattr) if it is a
mountpoint

Signed-off-by: Yuval Turgeman <yturgema@redhat.com>


yuvalturg:
This is already covered in files_search_all_mountpoints, sorry.

PR #4, opened 2018-11-09
Bug fixes

bigon:
Some fixes for denials I saw on a default/minimal debian installation

The refpolicy definitely needs more test/love on debian, there are still a lot of denials I'm not sure how to fix in the minimal install


policy/modules/services/irqbalance.te
```diff
@@ -28,6 +28,7 @@ allow irqbalance_t self:capability { setpcap };
 dontaudit irqbalance_t self:capability sys_tty_config;
 allow irqbalance_t self:process { getcap getsched setcap signal_perms };
 allow irqbalance_t self:udp_socket create_socket_perms;
+allow irqbalance_t self:unix_stream_socket { accept listen };
```
pebenito: Should probably be:
```suggestion
allow irqbalance_t self:unix_stream_socket create_stream_socket_perms;
```
The other create_stream_socket_perms already come in from other interfaces, but this makes it explicit.

PR #8, opened 2018-12-30
Syncthing

alexminder:


policy/modules/apps/syncthing.te
```diff
@@ -67,3 +77,7 @@ optional_policy(`
 	# temporary hack for /run/NetworkManager/resolv.conf until we make this part of sysnet_dns_name_resolve()
 	networkmanager_read_pid_files(syncthing_t)
 ')
+
+tunable_policy(`syncthing_can_run_ifconfig',`
```
pebenito: This tunable block should actually be above the optional_policy.

policy/modules/apps/syncthing.te
```diff
@@ -67,3 +77,7 @@ optional_policy(`
...
+        sysnet_domtrans_ifconfig(syncthing_t)
```
pebenito: If syncthing isn't making any changes, it may work with a few additional, reasonable, perms, for syncthing to simply exec ifconfig without transition.

alexminder: ifconfig(iptorute2) without transition does not have all the required permissions to get route information from syncthing_t domain. I tried it.

PR #9, opened 2018-12-30
Postgres

alexminder:


policy/modules/kernel/filesystem.if
```diff
@@ -2337,7 +2337,7 @@ interface(`fs_rw_hugetlbfs_files',`
 		type hugetlbfs_t;
 	')
 
-	rw_files_pattern($1, hugetlbfs_t, hugetlbfs_t)
+	mmap_rw_files_pattern($1, hugetlbfs_t, hugetlbfs_t)
```
pebenito: Is the mmap definitely required in all cases? If not, the map should be separated into a different interface.

pebenito: There have been a lot of map additions to the policy lately. None which went to `hugetlbfs_t`, so I imagine this probably should be a separate interface called by the specific caller that you're concerned with (`postgresql_t`?)

PR #16, opened 2019-01-06
authlogin: allow callers of getpwent() to get dynamic users

fishilico:
On a system with systemd as init, nss-systemd module makes D-Bus calls to systemd.

cf. https://lore.kernel.org/selinux-refpolicy/87lg3y8ckd.fsf@gmail.com/T/

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -433,6 +433,11 @@ tunable_policy(`authlogin_nsswitch_use_ldap',`
 	sysnet_use_ldap(nsswitch_domain)
 ')
 
+ifdef(`init_systemd',`
+	# nss-systemd calls GetDynamicUsers() exposed by systemd over D-Bus
+	init_dbus_chat(nsswitch_domain)
+')
+
 optional_policy(`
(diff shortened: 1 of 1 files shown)


policy/modules/system/authlogin.te
```diff
@@ -433,6 +433,11 @@ tunable_policy(`authlogin_nsswitch_use_ldap',`
 	sysnet_use_ldap(nsswitch_domain)
 ')
 
+ifdef(`init_systemd',`
+	# nss-systemd calls GetDynamicUsers() exposed by systemd over D-Bus
+	init_dbus_chat(nsswitch_domain)
+')
+
```
pebenito: This block should go above the tunable_policy authlogin_nsswitch_use_ldap block. Also waiting pending other discussions about this change on list.

fishilico:
The issue regarding OpenSSH server has been fixed with https://github.com/SELinuxProject/refpolicy/commit/ef6c7f155e10deb50040da05c62883f0305829b7#diff-3e78b5c68b1436d3d5965d1f2f7850d8R281 and I agree with nsswitch_domain being too complex to add more accesses to it. Let's drop this patch.

PR #22, opened 2019-01-13
selinuxutil: allow restorecond to read symlinks

fishilico:
As restorecond dereferences symlinks when it encounters them in user home directories, allow this access.

### DIFF
--- a/policy/modules/system/selinuxutil.te
+++ b/policy/modules/system/selinuxutil.te
@@ -372,6 +372,7 @@ selinux_compute_user_contexts(restorecond_t)
 
 files_relabel_non_auth_files(restorecond_t )
 files_read_non_auth_files(restorecond_t)
+files_read_non_auth_symlinks(restorecond_t)
 auth_use_nsswitch(restorecond_t)
 
 logging_send_syslog_msg(restorecond_t)


stephensmalley:
This is a potential security risk.  You don't want to allow a domain to read/follow untrusted symlinks, as this exposes it to malicious symlink attacks.  In particular, in the case of restorecond, this could potentially be used to trick restorecond into relabeling a file outside the user's home directory to a context accessible to the user.

fishilico:
The AVC denials that I got was associated with a `type=SYSCALL` entry with `arch=c000003e syscall=137`, which is `statfs` syscall.

Here is what I see on a system with `restorecond` running with default configuration (as `root`, with `/root/*` in `/etc/selinux/restorecond.conf`) when using `strace` on this service and executing `ln -s target /root/symlink` on the system:
```
read(3, "\4\0\0\0\0\1\0\0\0\0\0\0\20\0\0\0symlink\0\0\0\0\0\0\0\0\0", 32768) = 32
lstat("/root/symlink", {st_mode=S_IFLNK|0777, st_size=6, ...}) = 0
statfs("/root/symlink", 0x7ffd6bb4d090) = -1 ENOENT (No such file or directory)
lstat("/root/symlink", {st_mode=S_IFLNK|0777, st_size=6, ...}) = 0
lgetxattr("/root/symlink", "security.selinux", "sysadm_u:object_r:user_home_t", 255) = 30
```
The first syscall is the inotify event received by `restorecond` and the third one triggers the symlink dereferencing. No other syscalls dereference the symlink so there is no security vulnerability, but I agree that this PR weakens the policy.

The code which performs this call to `statfs` is in libselinux, with comment "Ignore restoreconlast on in-memory filesystems": https://github.com/SELinuxProject/selinux/blob/a9f8a101fd2c475b3dcffbcc76b86355d956a095/libselinux/src/selinux_restorecon.c#L882-L886

The implementation of `statfs` uses `LOOKUP_FOLLOW` to resolve the path, which is why it reads symlinks (cf. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/statfs.c?h=v4.20#n85). The man page of `statfs` (http://man7.org/linux/man-pages/man2/statfs.2.html) gives another syscall which works on opened file descriptors, `fstatfs`, but migrating to this syscall would make `restorecond` open the files that may be relabeled. This would also require to weaken the policy... unless if `open(..., O_PATH)` does not check for `open` permission on the file (if this is the case, `fstatfs` can be used with a file descriptor opened with `O_PATH` since Linux 3.12 according to http://man7.org/linux/man-pages/man2/open.2.html).

Does `open(..., O_PATH)` leads to checking `file:open` permission?

stephensmalley:
First, I have some doubts as to whether restorecond.conf should contain /root/* or /root/.ssh/* by default.
Second, I have some doubts as to whether restorecond is truly necessary/useful still given support for name-based type transitions in policy (i.e. can't we just define name-based type transitions for all of these files to ensure proper labeling)?
Third, if the statfs() fails, the only thing that happens is that we leave setrestoreconlast unchanged, and in this case it is already false (not a directory), so no change in behavior even if we deny the access.  That makes it a dontaudit candidate.
Fourth, looking at the kernel code, it appears that open(..., O_PATH) returns before reaching the security_file_open() hook call and thus skips any permission checking on the final component, even open permission.  I don't think anyone thought about this when O_PATH was introduced. That could possibly be considered a bug but I don't think we can change it now as it would break existing userspace/policy that may be relying on this behavior. @pcmoore

doverride:
stephensmalley <notifications@github.com> writes:

> First, I have some doubts as to whether restorecond.conf should contain /root/* or /root/.ssh/* by default.
> Second, I have some doubts as to whether restorecond is truly necessary/useful still given support for name-based type transitions in policy
> (i.e. can't we just define name-based type transitions for all of these files to ensure proper labeling)?

Not all of it, no. I do exactly that and it's certainly worth it in my
experience, but there will be trade-offs if you rely solely on
that. Various files will have to have generic contexts as they cannot
have private labels due being created with random suffixes and then
being moved (example ~/.bash_history).

> Third, if the statfs() fails, the only thing that happens is that we leave setrestoreconlast unchanged, and in this case it is already false
> (not a directory), so no change in behavior even if we deny the access. That makes it a dontaudit candidate.
> Fourth, looking at the kernel code, it appears that open(..., O_PATH) returns before reaching the security_file_open() hook call and thus skips
> any permission checking on the final component, even open permission. I don't think anyone thought about this when O_PATH was introduced. That
> could possibly be considered a bug but I don't think we can change it now as it would break existing userspace/policy that may be relying on
> this behavior. @pcmoore
>
> —
> Reply to this email directly, view it on GitHub, or mute the thread.
>

-- 
Dominick Grift

stephensmalley:
systemd selinux code uses the idiom of (error handling omitted):
fd = open(path, O_NOFOLLOW|O_CLOEXEC|O_PATH);
fstat(fd, &st);
selabel_lookup_raw(label_hnd, &fcon, path, st.st_mode);
xsprintf(procfs_path, "/proc/self/fd/%i", fd);
setfilecon_raw(procfs_path, fcon);

fishilico:
The `xsprintf` + `setfilecon_raw` looks more complex that I could be, because `fsetfilecon_raw` exists. Anyway, using `open(..., O_PATH|O_NOFOLLOW)` here is not needed.

> Third, if the statfs() fails, the only thing that happens is that we leave setrestoreconlast unchanged, and in this case it is already false (not a directory), so no change in behavior even if we deny the access. That makes it a dontaudit candidate.

Thanks to this comment, I realized that the call to `statfs()` can be skipped if `setrestoreconlast` is already `false`... I will send a patch to selinux@ fixing this.

I agree with changing the `allow` to `dontaudit`.

stephensmalley:
> The `xsprintf` + `setfilecon_raw` looks more complex that I could be, because `fsetfilecon_raw` exists.

No, you can't perform operations like fgetfilecon (fgetxattr) or fsetfilecon (fsetxattr) on a descriptor opened with O_PATH; it will return EBADF.

pebenito:
While this is definitely something to switch to `dontaudit`, what concerns me is why there is `files_read_non_auth_files(restorecond_t)`. I glanced though the code and nothing jumped out at me.

PR #27, opened 2019-01-23
Postgres

alexminder:


policy/modules/kernel/filesystem.if
```diff
@@ -2340,6 +2340,25 @@ interface(`fs_rw_hugetlbfs_files',`
...
+## <summary>
+##      Read, map and write hugetlbfs files.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`fs_rmw_hugetlbfs_files',`
```
pebenito: this should be 'fs_mmap_read_hugetlbfs_files()`.

PR #29, opened 2019-01-23
sysnet_dns_name_resolve interface update for musl libc.

alexminder:
Add corenet_udp_bind_generic_node for sysnet_dns_name_resolve interface because with musl libc it required for address resolution.


pebenito:
I'm not accepting changes that only apply to musl at this time. See this thread, which would have the minimum needed patch to add musl: https://lore.kernel.org/selinux-refpolicy/20181118113738.29532-1-jason@perfinion.com/

PR #32, opened 2019-02-23
Allow systemd-journald to use kill(pid, 0) on every process

fishilico:
Since systemd 241, systemd-journald is using `kill(pid, 0)` in order to find dead processes and reduce its cache. The relevant commit is https://github.com/systemd/systemd/commit/91714a7f427a6c9c5c3be8b3819fee45050028f3 ("journald: periodically drop cache for all dead PIDs"). This commit added a call to `pid_is_unwaited(c->pid)`, which is a function implemented in https://github.com/systemd/systemd/blob/v241/src/basic/process-util.c#L936 :
```c
    bool pid_is_unwaited(pid_t pid) {
        /* Checks whether a PID is still valid at all, including a zombie */
        if (pid < 0)
                return false;
        if (pid <= 1) /* If we or PID 1 would be dead and have been waited for, this code would not be running */
                return true;
        if (pid == getpid_cached())
                return true;
        if (kill(pid, 0) >= 0)
                return true;
        return errno != ESRCH;
    }
```
This new code triggers the following AVC denials:

    type=AVC msg=audit(1550911933.606:332): avc:  denied  { signull }
    for  pid=224 comm="systemd-journal"
    scontext=system_u:system_r:syslogd_t
    tcontext=system_u:system_r:auditd_t tclass=process permissive=1

    type=AVC msg=audit(1550911933.606:333): avc:  denied  { signull }
    for  pid=224 comm="systemd-journal"
    scontext=system_u:system_r:syslogd_t
    tcontext=system_u:system_r:dhcpc_t tclass=process permissive=1

    type=AVC msg=audit(1550911933.606:334): avc:  denied  { signull }
    for  pid=224 comm="systemd-journal"
    scontext=system_u:system_r:syslogd_t
    tcontext=system_u:system_r:sshd_t tclass=process permissive=1

### DIFF
--- a/policy/modules/system/logging.if
+++ b/policy/modules/system/logging.if
@@ -638,6 +638,11 @@ interface(`logging_send_syslog_msg',`
 	# will write to the console.
 	term_write_console($1)
 	term_dontaudit_read_console($1)
+
+	ifdef(`init_systemd',`
+		# Allow systemd-journald to check whether the process died
+		allow syslogd_t $1:process signull;
+	')
 ')
 
 ########################################


doverride:
I would narrow this down to only apply to `logging_send_syslog_msg()` callers. By adding this to `logging_send_syslog_msg()` (should probably add systemd related stuff in an ifdef_systemd block):

`allow syslogd_t $1:process signull;`

Maybe even introduce a type attribute: logging_send_syslog_msg_type, associate callers with the type attribute and then use the typeattribute to associate common rules like the rule allowing journald to send null signals to all clients (this might be more efficient, if it works)

doverride:
The above is what I implemented and it seems to work for me.
```
$ d2sesearch -A -t journald.client_subj_type_attribute -dt
allow journald.subj journald.client_subj_type_attribute:process signull;
```

doverride:
The problem with implementing the suggested `logging_send_syslog_msg_type` type attribute is that then  it is impossible to call `logging_send_send_syslog_msg()` with a type attribute argument. Because refpolicy cannot associate type attributes with type attributes. So this is really a shitty trade off to consider:

1. if you use `domain_signull_all_domains(syslogd_t)` then this is very efficient but arguably too broad. One could argue: it's worth the trade off.
2. if you add `allow syslogd_t $1:process signull;` to `logging_send_syslog_msg()` then this is specific, and it makes sense when you read it. However it is very inefficient. In other words: expensive.
3. If you implement a type attribute and associate rules common to syslog users, and syslog users with the type attribute, then you can have the best of both worlds. The problem is, as mentioned above, that llogging_send_syslog_msg() can only be called with types as the argument. Efficient, but potentially painful and disruptive.

doverride:
This is one of the major reason's I ,reluctantly, decided to move away from module policy. It would not allow me to do the right thing due to technical limitations, while there was an alternative that would.

fishilico:
For option 2, here are some statistics on my policy which is very close to refpolicy:

* `sesearch -A -s syslogd_t -p signull` (after `logging_send_syslog_msg()` is modified) returns 468 lines.
* `seinfo -x -a domain` returns 803 types.

This means that option 1 adds a permission which is not needed for approx. 42% of types associated to `domain` attribute.

What is considered inefficient/expensive in the policy? I have several metrics to my mind:
* the number of types in the policy (which is not modified here) ;
* the number of attributes that the kernel needs to expand when loading a policy ;
* the number of rules processed by userspace tools (and for example returned by `sesearch`) ;
* the size of the compiled policy file (which is quite hard to predict, as the access vector tables are stored using sparse bitmaps).

doverride:
Good question. I would say that all of the above should be taken into consideration. Using too many type attributes can have consequences as well (Google faced this and contributed a patch that allows one to force expansion of type attributes)

doverride:
The nice thing about CIL is that it stimulates the use of type attributes, and it is optimized for this scenario. I personally have yet to reach the point where my extensive use of type attributes is a botlleneck. But if that were ever the case I would decide to force expansion of specified type attributes to tune things.

Another nice benefit of using type attribute to the fullest is that it make policy analysis much nicer (provided that the policy analysis suite is optimized as well)

doverride:
RHEL6 was a really good example of where things really got out of hand. The footprint of the policy was huge, load and compile times were high (semodule -i woud take ages). And people started contributing all kinds of work arounds to stop the bleeding (bzipping pp's for example)

doverride:
Option 2 also adds an extra 467 lines of policy compared to option 1 which adds 1 lines.

That looks pretty expensive to me no matter what metrics you take

doverride:
Another drawback of using type attributes (in module policy as well as CIL) is that you cannot associate with type attributes in booleans. So that would mean that you would not be able to call `logging_send_syslog_msg()` conditionally.

pebenito:
The correct way is the rule in `logging_send_syslog_msg()`. In my opinion, this is not too expensive.

fishilico:
Thank you for your feedback. I have modified my commit to implement option 2.

PR #38, opened 2019-03-21
logging: Reduce the scope of systemd-journald accesses

fishilico:
`systemd-journald` only requests the states of processes that are sending messages to logs. Make the policy more restrictive in this regard.

### DIFF
--- a/policy/modules/system/logging.if
+++ b/policy/modules/system/logging.if
@@ -640,8 +640,11 @@ interface(`logging_send_syslog_msg',`
 	term_dontaudit_read_console($1)
 
 	ifdef(`init_systemd',`
-		# Allow systemd-journald to check whether the process died
-		allow syslogd_t $1:process signull;
+		# Allow systemd-journald to read the state of its clients and to check whether they died
+		allow syslogd_t $1:process {getattr signull};
+		allow syslogd_t $1:dir search_dir_perms;
+		allow syslogd_t $1:file read_file_perms;
+		allow syslogd_t $1:lnk_file read_lnk_file_perms;
 	')
 ')
 
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -485,8 +485,6 @@ dev_read_urand(syslogd_t)
 dev_rw_kmsg(syslogd_t)
 
 domain_use_interactive_fds(syslogd_t)
-# Allow access to /proc/ information for journald
-domain_read_all_domains_state(syslogd_t)
 
 files_read_etc_files(syslogd_t)
 files_read_usr_files(syslogd_t)
@@ -540,9 +538,6 @@ ifdef(`init_systemd',`
 	dev_read_kmsg(syslogd_t)
 	dev_read_urand(syslogd_t)
 	dev_write_kmsg(syslogd_t)
-
-	domain_getattr_all_domains(syslogd_t)
-	domain_read_all_domains_state(syslogd_t)
 
 	init_create_runtime_dirs(syslogd_t)
 	init_daemon_pid_file(syslogd_var_run_t, dir, "syslogd")


doverride:
Turns out that its not as simple as this unfortunately (same with the signull actually)

journald is not only sending null signals to journal clients, it is actually sending null signals to atleast all systemd system daemons, maybe even processes forked by systemd system daemon as well (not sure).. Systemd catches stdout and stderror, and that ends up in journal.

With the state reading it seems to get even more complicated as journald also reads state of systemd user daemon processes even if theyre not journal clients

fishilico:
> all systemd system daemons

Most daemons use `systemd_log_parse_environment()` or `logging_send_syslog_msg()` in refpolicy, so I though this case was covered.

> maybe even processes forked by systemd system daemon as well (not sure).. Systemd catches stdout and stderror, and that ends up in journal.

Indeed. When a daemon consists in a program launched as `my_daemon_t` with stdout/stderr caught by journald, I expect the policy to contain a rule such as `logging_send_syslog_msg(my_daemon_t)`, but if this daemon launches a process with a type transition, the new type does not have to have `logging_send_syslog_msg(new_type_t)` (and adding such rules everywhere looks inefficient).

> With the state reading it seems to get even more complicated as journald also reads state of systemd user daemon processes even if theyre not journal clients

I forgot about them. In order to test the refpolicy, I am currently using a virtual machine with a system which is quite minimal (base install + some services). It looks useful to configure a user daemon process on this system in order to cover its access. Do you happen to know one which is simple to enable?

doverride:
> > all systemd system daemons
> 
> Most daemons use `systemd_log_parse_environment()` or `logging_send_syslog_msg()` in refpolicy, so I though this case was covered.

I mean systemd system daemon in general (e.g. The commands called with Exec...= in all the service units)

There are few system daemons that arent also journal clients though in practice, but i encountered some. for example rngd.

> 
> > maybe even processes forked by systemd system daemon as well (not sure).. Systemd catches stdout and stderror, and that ends up in journal.
> 
> Indeed. When a daemon consists in a program launched as `my_daemon_t` with stdout/stderr caught by journald, I expect the policy to contain a rule such as `logging_send_syslog_msg(my_daemon_t)`, but if this daemon launches a process with a type transition, the new type does not have to have `logging_send_syslog_msg(new_type_t)` (and adding such rules everywhere looks inefficient).

Yes, however do note that i am not sure if the stdout/stderr of the forked process ends up with systemd/journald. As a matter of fact, I don't think it does. But i am not sure. (I havent encountered this yet)

> 
> > With the state reading it seems to get even more complicated as journald also reads state of systemd user daemon processes even if theyre not journal clients
> 
> I forgot about them. In order to test the refpolicy, I am currently using a virtual machine with a system which is quite minimal (base install + some services). It looks useful to configure a user daemon process on this system in order to cover its access. Do you happen to know one which is simple to enable?

I don't know if refpolicy supports this functionality yet. But assuming that your systemd --user instance runs with unconfined_u:unconfined_r:unconfined:s0, any user daemon will run with that as well and that will cause systemd-journald to read the state of unconfined_t processes, which atleast indicates that it reads state of systemd --user daemons

I did test this though on dssp2, i have for example redshift configured as a systemd --user daemon and it does not log but it does print to stdout:

```
Mar 21 10:00:10 brutus audit[1066]: AVC avc:  denied  { search } for  pid=1066 comm="systemd-journal" name="9852" dev="proc" ino=63133 scontext=sys.id:sys.role:journald.subj:s0 tcontext=wheel.id:wheel.role:redshift.subj:s0 tclass=dir permissive=0
Mar 21 10:00:10 brutus redshift[9852]: Location: 51.80 N, 4.67 E
```

doverride:
So What I did in dssp2:

1. i just allowed systemd-journald to read state of all domains (This is a typical example of compromise)
2. I allowed systemd-journald to send null signals to init daemons (to catch init daemons that arent also journal clients)
3. I allowed systemd-journald to send null signals to journal clients

Yes 2 and 3 overlap most of the time. It was either this or making systemd daemons that dont actually log to journal, journal clients anyway. I choose the former.

doverride:
Example (and also an example of how easy it is to test these things with my dssp2 live cd):
https://www.youtube.com/watch?v=sOwqUyCqkPo

doverride:
This example might be a bit better: https://www.youtube.com/watch?v=0b-AsQDQmP8

pebenito:
@fishilico did you have any changes to the PR? (I'm not requesting any)

pebenito:
ping @fishilico

doverride:
I would not reduce the scope here. As a matter of fact, remember that journald "signull" thing we went through a while ago? Eventually i ended up broadening that scope by allowing journald to send null signal to all domains after all. Things are just that entangled.

It turns out (i believe) that journald reads state and sends null signals to all processes forked by pid, which is essentially everything. Just because that might not be immediately appearant, eventually youll find that they will surface one by one.

pebenito:
Closing. @fishilico feel free to reopen if you want to further discuss this.

fishilico:
Hi, sorry for the delay. I have been very busy lately, and I will not have time to work on testing this before June/July. The work done by @doverride indicates that my PR was too restrictive, so I am fine with not merging it.
Thanks

PR #39, opened 2019-03-22
init: Revise conditions in init_startstop_service().

pebenito:
The role transition for directly executing an init script should only
happen if direct_sysadm_daemon is enabled and only on non-systemd
systems.

@perfinion 
@bigon 
@fishilico 
@bachradsusi

### DIFF
--- a/policy/modules/system/init.if
+++ b/policy/modules/system/init.if
@@ -1728,28 +1728,29 @@ interface(`init_startstop_service',`
 		role system_r;
 	')
 
-	ifndef(`direct_sysadm_daemon',`
-		ifdef(`distro_gentoo',`
-			# for OpenRC
-			seutil_labeled_init_script_run_runinit($1, $2, $4)
-		',`
-			# rules for sysvinit / upstart
-			init_labeled_script_domtrans($1, $4)
-			domain_system_change_exemption($1)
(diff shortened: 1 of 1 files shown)


perfinion:
@pebenito Hm this looks like it changes things around for Gentoo? I want to be able to have both OpenRC and systemd enabled for Gentoo. The migration path requires rebuilding the policies to enable systemd first when still on OpenRC and the user should be able to reboot between both without any trouble. 

Thats why I originally had the ifdef's for systemd and openrc in separate blocks. Is there a better way to do this?

pebenito:
I'm ok putting it back outside the systemd ifdef. This is why I explicitly requested the review. :)

pebenito:
I'd prefer that all 4 people mentioned state that there are no concerns before I merge this.

bachradsusi:
No concerns from my side. According to @wrabcak  neither Fedora nor Red Hat Enterprise Linux would be affected by this change.

etbe:
I think it would be good to support both systemd and non-systemd on Debian. This change does make the policy less ugly which is a good thing.

pebenito:
@etbe I'm taking to mean there are no objections?

pebenito:
I decided to simply flatten it out.

PR #40, opened 2019-04-03
Qt library version 5 requires unprivileged users to write xserver_tmp_t

gtrentalancia:
The Qt library version 5 requires to write xserver_tmp_t files upon starting up applications (tested on version 5.12.1).
---
 policy/modules/services/xserver.if |    1 +
 1 file changed, 1 insertion(+)

### DIFF
--- a/policy/modules/services/xserver.if
+++ b/policy/modules/services/xserver.if
@@ -148,6 +148,9 @@ interface(`xserver_role',`
 	allow $2 xserver_t:shm rw_shm_perms;
 	allow $2 xserver_tmpfs_t:file rw_file_perms;
 
+	# XCB Event Queue: used by the Qt library for example
+	allow $2 xserver_tmp_t:file rw_file_perms;
+
 	allow $2 iceauth_home_t:file manage_file_perms;
 	allow $2 iceauth_home_t:file { relabelfrom relabelto };
 


policy/modules/services/xserver.if
```diff
@@ -146,6 +146,7 @@ interface(`xserver_role',`
 
 	# Communicate via System V shared memory.
 	allow $2 xserver_t:shm rw_shm_perms;
+	allow $2 xserver_tmp_t:file rw_file_perms;
```
pebenito: What errors do you get if you don't allow this?

I'd like to determine if this makes sense with the other shm accesses, or if it should go somewhere else in the interface.

gtrentalancia:
There is no error produced in the terminal or GUI.

The applications simply refuse to start or hhe GUI simply gets stuck during initialization as in a "frozen" GUI.

I hope this helps...

Regards,

Guido

gtrentalancia:
Please note that this problem might be due to a more general libxcb pattern rather than merely a Qt related issue.

In any case it is due to operations on the XCB Event Queue.

Regards,

Guido

pebenito:
In that case, please separate from the "Communicate via System V shared memory" block, and add a comment so we can keep track of that.

gtrentalancia:
> In that case, please separate from the "Communicate via System V shared memory" block, and add a comment so we can keep track of that.

Done, thanks.

PR #41, opened 2019-04-17
Support CD ripper programs (e.g. cdda2wav)

gtrentalancia:
Extend the cdrecord apps module in order to support CD ripper
programs such as cdda2wav (including CDDB lookup functionality).

Update the cdrecord local policy with a few missing critical
permissions.

Add the file context for the readcd program executable file.

---
 policy/modules/apps/cdrecord.fc |    6 +
 policy/modules/apps/cdrecord.if |   29 ++++++++
 policy/modules/apps/cdrecord.te |  120 ++++++++++++++++++++++++++++++++++++-
 policy/modules/roles/staff.te      |    4 +
 policy/modules/roles/sysadm.te     |    4 +
 policy/modules/roles/unprivuser.te |    4 +
 6 files changed, 163 insertions(+), 4 deletions(-)

### DIFF
--- a/policy/modules/apps/cdrecord.if
+++ b/policy/modules/apps/cdrecord.if
@@ -49,3 +49,32 @@ interface(`cdrecord_exec',`
 	corecmd_search_bin($1)
 	can_exec($1, cdrecord_exec_t)
 ')
+
+########################################
+## <summary>
+##	Role access for cdripper.
+## </summary>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	User domain for the role.
+##	</summary>
+## </param>
+#
+interface(`cdripper_role',`
+	gen_require(`
+		attribute_role cdripper_roles;
+		type cdripper_t, cdripper_exec_t;
+	')
+
+	roleattribute $1 cdripper_roles;
+
+	domtrans_pattern($2, cdripper_exec_t, cdripper_t)
+
+	allow $2 cdripper_t:process { ptrace signal_perms };
+	ps_process_pattern($2, cdripper_t)
+')
--- a/policy/modules/apps/cdrecord.te
+++ b/policy/modules/apps/cdrecord.te
@@ -15,7 +15,20 @@ policy_module(cdrecord, 2.6.1)
 ## </desc>
 gen_tunable(cdrecord_read_content, false)
 
+## <desc>
+##	<p>
+##	Determine whether the cdripper
+##	programs can read and write various
+##	content, including user home
+##	directories, user temporary
+##	directories, nfs, samba, devices
+##	and untrusted content files
+##	</p>
+## </desc>
+gen_tunable(cdripper_rw_content, false)
+
 attribute_role cdrecord_roles;
+attribute_role cdripper_roles;
 
 type cdrecord_t;
 type cdrecord_exec_t;
@@ -24,13 +37,18 @@ typealias cdrecord_t alias { auditadm_cdrecord_t secadm_cdrecord_t };
 userdom_user_application_domain(cdrecord_t, cdrecord_exec_t)
 role cdrecord_roles types cdrecord_t;
 
+type cdripper_t;
+type cdripper_exec_t;
+userdom_user_application_domain(cdripper_t, cdripper_exec_t)
+role cdripper_roles types cdripper_t;
+
 ########################################
 #
-# Local policy
+# Cdrecord local policy
 #
 
-allow cdrecord_t self:capability { dac_override ipc_lock setuid sys_nice sys_rawio };
-allow cdrecord_t self:process { getcap getsched setrlimit setsched sigkill };
+allow cdrecord_t self:capability { dac_override ipc_lock setuid sys_nice sys_rawio sys_resource };
+allow cdrecord_t self:process { getcap getsched setrlimit setsched sigkill signal };
 allow cdrecord_t self:unix_stream_socket { accept listen };
 
 corecmd_exec_bin(cdrecord_t)
@@ -48,6 +66,7 @@ term_list_ptys(cdrecord_t)
 
 storage_raw_read_removable_device(cdrecord_t)
 storage_raw_write_removable_device(cdrecord_t)
+storage_read_scsi_generic(cdrecord_t)
 storage_write_scsi_generic(cdrecord_t)
 
 logging_send_syslog_msg(cdrecord_t)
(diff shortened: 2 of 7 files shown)


policy/modules/apps/cdrecord.if
```diff
@@ -49,3 +49,32 @@ interface(`cdrecord_exec',`
...
+		type cdripper_t, cdripper_exec_t;
+	')
+
+	roleattribute $1 cdripper_roles;
+
+	domtrans_pattern($2, cdripper_exec_t, cdripper_t)
+
+	allow $2 cdripper_t:process { ptrace signal_perms };
+	ps_process_pattern($2, cdripper_t)
+')
```
doverride: I see no reason for a `cdripper_role()`. You should probably use a `cdrecord_domtrans_cdripper()` and `cdrecord_run_cdripper()` instead

policy/modules/apps/cdrecord.te
```diff
@@ -15,7 +15,20 @@ policy_module(cdrecord, 2.6.1)
...
+##	Determine whether the cdripper
+##	programs can read various content,
+##	including user home directories,
+##	user temporary directories, nfs,
+##	samba, devices and untrusted
+##	content files
+##	</p>
+## </desc>
+gen_tunable(cdripper_rw_content, false)
+
```
doverride: I suppose this is debatable. I think this is overdoing it. I would personally probably just allow it to atleast manage generic home content and maybe also generic tmp content unconditionally and get it over with.

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+# Cdripper (e.g. cdda2wav) local policy
+#
+
+allow cdripper_t self:capability { ipc_lock sys_nice setgid setuid dac_override sys_rawio };
+allow cdripper_t self:process { getcap getsched setcap setrlimit setsched sigkill };
+allow cdripper_t self:fifo_file rw_fifo_file_perms;
+allow cdripper_t self:tcp_socket create_socket_perms;
+allow cdripper_t self:udp_socket create_socket_perms;
+allow cdripper_t cdripper_exec_t:file execute_no_trans;
+
```
doverride: `can_exec(cdripper_t, cdripper_exec_t)`

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+########################################
+#
+# Cdripper (e.g. cdda2wav) local policy
+#
+
+allow cdripper_t self:capability { ipc_lock sys_nice setgid setuid dac_override sys_rawio };
+allow cdripper_t self:process { getcap getsched setcap setrlimit setsched sigkill };
+allow cdripper_t self:fifo_file rw_fifo_file_perms;
+allow cdripper_t self:tcp_socket create_socket_perms;
+allow cdripper_t self:udp_socket create_socket_perms;
```
doverride: The above is probably redundant is it is probably already enclosed with the below `sysnet_dns_name_resolve(cdrippter_t)` call, however I would probably call `auth_use_nsswitch(cdripper_t)` instead (which by the way also implies `sysnet_dns_name_resolve(cdripper_t)`

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+allow cdripper_t cdripper_exec_t:file execute_no_trans;
+
+corecmd_exec_bin(cdripper_t)
+corecmd_exec_shell(cdripper_t)
+
+corenet_tcp_connect_all_unreserved_ports(cdripper_t)
```
doverride: That seems pretty broad but ok, atleast its narrowed down to `all unreserved` only

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+
+dev_list_all_dev_nodes(cdripper_t)
```
doverride: Not your fault but:

This interface does not make sense to me (at least its name, i have not looked it up so i do not know what it implies).  "List" is a directory permission. There is only one directory type in "/dev" which is "device_t" so to say list all dev nodes seems weird. You don't "list dev nodes"

I think this should probably have been splitted in at least two interfaces with saner names, for example:
```
dev_list(cdripper_t)
dev_getattr_all_dev_nodes(cdripper_t)
```
maybe even a `dev_read_lnk_files(cdripper_t)` added.

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+dev_read_sysfs(cdripper_t)
+
+domain_interactive_fd(cdripper_t)
```
doverride: What made you add this? This interface is very broad and probably should not be used here

policy/modules/apps/cdrecord.te
```diff
@@ -113,3 +132,98 @@ tunable_policy(`use_nfs_home_dirs',`
...
+	userdom_dontaudit_list_user_home_dirs(cdripper_t)
+	userdom_dontaudit_write_user_home_content_files(cdripper_t)
+')
+
+tunable_policy(`use_nfs_home_dirs',`
+	files_search_mnt(cdripper_t)
+	fs_manage_nfs_files(cdripper_t)
+	fs_manage_nfs_symlinks(cdripper_t)
+')
+
```
doverride: Let me get this straight: I what scenario would you be able to use "cdripper" without setting `cdripper_rw_content`? Were will you be ripping to?

policy/modules/apps/cdrecord.if
gtrentalancia: The two roles are independent of each other, in other words cdrecord_role() does not imply cdripper_role() and viceversa.

People might want to enable one but not the other.

So this is a nack.

policy/modules/apps/cdrecord.te
gtrentalancia: This reflects the current cdrecord role / local policy.

If cdrecord_t does not currently allow to even *read* home content, why should cdripper_t do so ?

This is another nack.

policy/modules/apps/cdrecord.te
gtrentalancia: ACK.

policy/modules/apps/cdrecord.te
gtrentalancia: Partial ACK.

I prefer not to further extend the permission set by involving NSSWITCH when all is needed is a DNS lookup for the CDDB server.

But thanks for pointing out the rest, useful !

policy/modules/apps/cdrecord.te
gtrentalancia: No longer needed, see above. ACK.

policy/modules/apps/cdrecord.te
gtrentalancia: NACK. Already there and in common use.

policy/modules/apps/cdrecord.te
gtrentalancia: ACK. Removed the unneeded interface calls (inherited from the existing cdrecord local policy).

policy/modules/apps/cdrecord.te
gtrentalancia: Don't know ++ don't care.

This is the way the cdrecord module was designed from the beginning and the aim of this patch is to enable the CD ripper functionality, not change the current module design.

This is a nack. But feel free to submit a separate patch, if you care so much about this issue and have time to spend developing and testing it.

policy/modules/apps/cdrecord.te
gtrentalancia: Actually, it cannot be removed...

gtrentalancia: But replaced with a new CDDB port definition in corenetwork.

policy/modules/apps/cdrecord.if
doverride: Why did you decide to append this cdripper domain to the existing cdrecord module?

These are two fundamentally different things. CD record writes to the removable device and reads from your home dir (example: burning ~/fedora.iso to a CD) CD ripper does the opposite. It reads the CD and writes it to your home directory.

gtrentalancia: Because they both belong to the same package which is cdrtools, so it doesn't make sense to create to different modules for programs that belong to the same package !

policy/modules/apps/cdrecord.te
doverride: cdrecord is allowed to read generic user home content files unconditionally, because that is what it does. it enables you to write a home content file to a cd.

cdripper should be allowed to manage generic user home content files unconditionally, because that is what it does. it enables you to rip a cd to a home content file.

policy/modules/apps/cdrecord.te
doverride: I am not asking you to interfere with the existing policy, but you cannot compare apples and oranges

policy/modules/apps/cdrecord.te
gtrentalancia: This is a NACK again.

cdrecord is NOT allowed to read generic user home content unconditionally, but only when the **cdrecord_read_content** is turned **ON** !

The equivalent (not the same) applies to the CD ripper program introduced in this patch:

tunable_policy(`cdripper_rw_content',`
        userdom_manage_user_home_content_files(cdripper_t)
        userdom_manage_user_home_content_symlinks(cdripper_t)
',`
        userdom_dontaudit_list_user_home_dirs(cdripper_t)
        userdom_dontaudit_manage_user_home_content_files(cdripper_t)
')

policy/modules/apps/cdrecord.if
doverride: Ok, but i think you might have misunderstood the point i was trying to make.

When i say that you should use a `cdrecord_run_cdripper()` instead of `cdripper_role()`, i am not referring to the type cdrecord_t, but instead i am referring to the cdrecord module. the cdripper_t type is declared in the cdrecord module and so when you create new interfaces for cdripper, you should prepend the name with cdrecord_ to indicate what module the interface is part of.

So even if you would for some reason stick with `cdripper_role()` you would (probably) still need to rename it to `cdrecord_cdripper_role()`.

Howver there is no need for a "role" here. And you should not argue that you need a role for cdripper because cdrecord uses one. cdripper is not cdrecord and vice versa.

policy/modules/apps/cdrecord.te
doverride: Have a look:

https://github.com/SELinuxProject/refpolicy/blob/master/policy/modules/apps/cdrecord.te#L58

policy/modules/apps/cdrecord.te
gtrentalancia: This is a pointless discussion !

First thing, I am not comparing different things !

Also, as already explained, I don't know where it is going to rip, but most importantly, **I don't care** !

It could rip to anywhere the underlying policy allows the calling domain to write, that's it.

If you want an example based on the current policy, then if the calling domain is _user_ it could write to USB removable media if the _user_write_removable_ boolean is turned ON.

pebenito: I agree with @doverride . There needs to be a default location to store the output. The home dir is the most obvious location. `cdripper_rw_content` should be removed and writes to the home content should be unconditional.

policy/modules/apps/cdrecord.if
pebenito: The name is definitely wrong, as @doverride noted.

gtrentalancia:
I'm afraid, I don't agree with you that the cdripper program should be allowed unconditionally to write the user home directory (and/or the other locations such as NFS mounts).

Either you make a subsequent commit with a different author or I'll close the Pull Request in a few hours.

doverride:
> I'm afraid, I don't agree with you that the cdripper program should be allowed unconditionally to write the user home directory (and/or the other locations such as NFS mounts).
> 
> Either you make a subsequent commit with a different author or I'll close the Pull Request in a few hours.

It is probably more constructive if you give a good counter argument. I have been reading `man cdda2wav` and this part struck me:

``
cdda2wav is able to read parts of an audio CD or multimedia CDROM (containing audio parts) directly digitally. These parts can be written to a file, a pipe, or to a sound device.

cdda2wav stands for CDDA to WAV (where CDDA stands for compact disc digital audio and WAV is a sound sample format introduced by MS Windows). It allows copying CDDA audio data from the CDROM drive into a file in WAV or other formats. 
```

So it seems that it is not a fact that it writes to a file. Instead you can also write directly to a sound device or pipe the output into some other program.

This may (or may not) be a valid reason to make this conditional.

Regardless of the above though, there are some other outstanding requests that you have not addressed

gtrentalancia:
I support the current cdrecord design and even more importantly the following **FUNDAMENTAL** security policy design rule:

http://oss.tresys.com/pipermail/refpolicy/2017-April/009325.html

I will **NEVER** come to a different compromise on user data confidentiality or integrity.

doverride:
I admire your "energy", but probably best to never say never. Especially when it comes to SELinux where compromise is inevitable. I use to be like this as well, Then I spent a decade confining the user session, and found out the hard way that you have to give and take. Find a balance and settle for a balance between security and usability.

gtrentalancia:
It's a matter of roles in the very first place !

It's the Security Policy Administrator that should decide the right balance between usability and security and set up the security strategy: it takes between 5 and 10 minutes to set up the booleans on a new system and virtually nothing to replicate them in a clustered environment.

It's the Security Policy Administrator that should take the responsibility for allowing or denying operations involving user data confidentialtity or integrity, NOT the Security Policy Designer.

The Security Policy is **General,** **Flexible** and **Customizable,** otherwise there is not point in having it.

The Security Policy Designer lays down the strategic infrastructure, but it's the Security Policy Administrator which ultimately decides the actual strategy and **takes responsibility for it**.

doverride:
You would end up with many "can_write_content" booleans depending on the number of confined user applications, Booleans are expensive. I prefer to treat generic content as "generic", and to not worry to much about giving access to generic content. The administrator can associate a private context with content he wants to protect instead. Generic content is meant to be shared and to be widely accessible. Even if what you are proposing was practical, then booleans governing access to generic content are still assuming that all users that execute the app with a domain transition are equal. It is just too expensive in my view. Generic content is not worth it in my moderated view.

PR #43, opened 2019-04-23
Various device labels

pebenito:


policy/modules/kernel/devices.fc
```diff
@@ -63,6 +63,7 @@
 /dev/mcelog		-c	gen_context(system_u:object_r:kmsg_device_t,mls_systemhigh)
 /dev/mei		-c	gen_context(system_u:object_r:mei_device_t,s0)
 /dev/mem		-c	gen_context(system_u:object_r:memory_device_t,mls_systemhigh)
+/dev/memory_bandwidth	-c	gen_context(system_u:object_r:memory_device_t,mls_systemhigh)
```
doverride: Did you have any particular reason to label this that way? I think this is one of the PM quality of service framework nodes:

/dev/cpu_dma_latency
/dev/network_latency
/dev/network_throughput
/dev/memory_bandwidth

Not sure if it is appropriate to label these this way. Not sure if its a given that something operating on memory_bandwith should also able to operate on mem.

pebenito: This is a good point, I didn't realize they were all related that way. The existing netcontrol_device_t will need to be renamed too.

PR #46, opened 2019-04-24
systemd: Add initial policy for systemd --user.

pebenito:
This is just a start; it does not cover all uses.

### DIFF
--- a/policy/modules/system/init.if
+++ b/policy/modules/system/init.if
@@ -652,6 +652,44 @@ interface(`init_domtrans',`
 
 ########################################
 ## <summary>
+##	Execute init (/sbin/init) with a domain transition
+##	to the provided domain.
+## </summary>
+## <desc>
+##	Execute init (/sbin/init) with a domain transition
+##	to the provided domain.  This is used by systemd
+##	to execute the systemd user session.
+## </desc>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	New domain.
+##	</summary>
+## </param>
+#
+interface(`init_pgm_spec_user_daemon_domain',`
+	gen_require(`
+		type init_t, init_exec_t;
+	')
+
+	domain_type($1)
+	domain_entry_file($1, init_exec_t)
+
+	spec_domtrans_pattern(init_t, init_exec_t, $1)
+
+	allow init_t $1:process { setsched rlimitinh noatsecure };
+
+	ifdef(`init_systemd',`
+		allow $1 init_t:unix_stream_socket { getattr read write ioctl };
+	')
+')
+
+########################################
+## <summary>
 ##	Execute the init program in the caller domain.
 ## </summary>
 ## <param name="domain">
@@ -668,6 +706,26 @@ interface(`init_exec',`
 
 	corecmd_search_bin($1)
 	can_exec($1, init_exec_t)
+')
+
+########################################
+## <summary>
+##	Allow the init program to be an entrypoint
+##	for the specified domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`init_pgm_entrypoint',`
+	gen_require(`
+		type init_exec_t;
+	')
+
+	allow $1 init_exec_t:file entrypoint;
 ')
 
 ########################################
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -1,5 +1,62 @@
 ## <summary>Systemd components (not PID 1)</summary>
 
+#########################################
+## <summary>
+##	Template for systemd --user per-role domains.
+## </summary>
+## <param name="prefix">
+##	<summary>
+##	Prefix for generated types
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	The user role.
+##	</summary>
+## </param>
+## <param name="userdomain">
+##	<summary>
+##	The user domain for the role.
+##	</summary>
+## </param>
+#
+template(`systemd_role_template',`
+	gen_require(`
+		attribute systemd_user_session_type, systemd_log_parse_env_type;
+		type systemd_user_runtime_t, systemd_user_runtime_notify_t;
+	')
+
+	#################################
+	#
+	# Declarations
+	#
+	type $1_systemd_t, systemd_user_session_type, systemd_log_parse_env_type;
+	init_pgm_spec_user_daemon_domain($1_systemd_t)
+	domain_user_exemption_target($1_systemd_t)
+	ubac_constrained($1_systemd_t)
+	role $2 types $1_systemd_t;
+
+	#################################
+	#
+	# Local policy
+	#
+
+	allow $3 systemd_user_runtime_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $3 systemd_user_runtime_t:file { manage_file_perms relabel_file_perms };
+	allow $3 systemd_user_runtime_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
+	allow $3 systemd_user_runtime_t:fifo_file { manage_fifo_file_perms relabel_fifo_file_perms };
+	allow $3 systemd_user_runtime_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
+
+	allow $3 systemd_user_runtime_notify_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
+
+	# This domain is per-role because of the below transitions.
+	# See the sytemd --user section of systemd.te for the
+	# remainder of the rules.
+	allow $1_systemd_t $3:process { setsched rlimitinh };
+	corecmd_shell_domtrans($1_systemd_t, $3)
+	corecmd_bin_domtrans($1_systemd_t, $3)
+')
+
 ######################################
 ## <summary>
 ##   Make the specified type usable as an
@@ -904,4 +961,58 @@ interface(`systemd_getattr_updated_runtime',`
 	')
 
 	getattr_files_pattern($1, systemd_update_run_t, systemd_update_run_t)
+')
+
+########################################
+## <summary>
+##	Search keys for the all systemd --user domains.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_search_all_user_keys',`
+	gen_require(`
+		attribute systemd_user_session_type;
+	')
+
+	allow $1 systemd_user_session_type:key search;
+')
+
+########################################
+## <summary>
+##	Create keys for the all systemd --user domains.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_create_all_user_keys',`
+	gen_require(`
+		attribute systemd_user_session_type;
+	')
+
+	allow $1 systemd_user_session_type:key create;
+')
+
+########################################
+## <summary>
+##	Write keys for the all systemd --user domains.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_write_all_user_keys',`
+	gen_require(`
+		attribute systemd_user_session_type;
+	')
+
+	allow $1 systemd_user_session_type:key write;
 ')
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -22,6 +22,7 @@ gen_tunable(systemd_nspawn_labeled_namespace, false)
 
 attribute systemd_log_parse_env_type;
 attribute systemd_tmpfiles_conf_type;
+attribute systemd_user_session_type;
 
 type systemd_activate_t;
 type systemd_activate_exec_t;
@@ -201,6 +202,12 @@ init_system_domain(systemd_update_done_t, systemd_update_done_exec_t)
 
 type systemd_update_run_t;
 files_type(systemd_update_run_t)
+
+type systemd_user_runtime_notify_t;
+userdom_user_runtime_content(systemd_user_runtime_notify_t)
+
+type systemd_user_runtime_t;
+userdom_user_runtime_content(systemd_user_runtime_t)
 
 #
 # Unit file types
@@ -1132,3 +1139,44 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
 seutil_read_file_contexts(systemd_update_done_t)
 
 systemd_log_parse_environment(systemd_update_done_t)
+
+#########################################
+#
+# User session (systemd --user) local policy
+#
+
+allow systemd_user_session_type self:capability { dac_read_search sys_resource };
+dontaudit systemd_user_session_type self:capability dac_override;
+allow systemd_user_session_type self:process setfscreate;
+allow systemd_user_session_type self:udp_socket create_socket_perms;
+allow systemd_user_session_type self:unix_stream_socket create_stream_socket_perms;
+allow systemd_user_session_type self:netlink_kobject_uevent_socket { bind create getattr setopt };
+
+allow systemd_user_session_type systemd_user_runtime_t:dir manage_dir_perms;
+allow systemd_user_session_type systemd_user_runtime_t:sock_file { create write };
+userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, dir)
+
+allow systemd_user_session_type systemd_user_runtime_notify_t:sock_file create;
+type_transition systemd_user_session_type systemd_user_runtime_t:sock_file systemd_user_runtime_notify_t "notify";
+
+dev_write_sysfs_dirs(systemd_user_session_type)
+dev_read_sysfs(systemd_user_session_type)
+
+files_read_etc_files(systemd_user_session_type)
+files_list_usr(systemd_user_session_type)
+
+fs_getattr_cgroup(systemd_user_session_type)
+fs_rw_cgroup_files(systemd_user_session_type)
+fs_manage_cgroup_dirs(systemd_user_session_type)
+
+init_signal(systemd_user_session_type)
+
+kernel_read_kernel_sysctls(systemd_user_session_type)
+
+mount_list_runtime(systemd_user_session_type)
+
+storage_getattr_fixed_disk_dev(systemd_user_session_type)
+
+# for systemd to read udev status
+udev_read_pid_files(systemd_user_session_type)
+udev_list_pids(systemd_user_session_type)
--- a/config/appconfig-mcs/default_contexts
+++ b/config/appconfig-mcs/default_contexts
@@ -1,4 +1,5 @@
 system_r:crond_t:s0		user_r:user_t:s0 staff_r:staff_t:s0 sysadm_r:sysadm_t:s0 unconfined_r:unconfined_t:s0 user_r:cronjob_t:s0 staff_r:cronjob_t:s0 sysadm_r:cronjob_t:s0 system_r:system_cronjob_t:s0 unconfined_r:unconfined_cronjob_t:s0
+system_r:init_t:s0		user_r:user_systemd_t:s0 staff_r:staff_systemd_t:s0 sysadm_r:sysadm_systemd_t:s0 unconfined_r:unconfined_t:s0
 system_r:local_login_t:s0	user_r:user_t:s0 staff_r:staff_t:s0 sysadm_r:sysadm_t:s0 unconfined_r:unconfined_t:s0
 system_r:remote_login_t:s0	user_r:user_t:s0 staff_r:staff_t:s0 unconfined_r:unconfined_t:s0
 system_r:sshd_t:s0		user_r:user_t:s0 staff_r:staff_t:s0 sysadm_r:sysadm_t:s0 unconfined_r:unconfined_t:s0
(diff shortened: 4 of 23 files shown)


policy/modules/system/systemd.te
```diff
@@ -1132,3 +1137,39 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+#
+
+allow systemd_user_session_type self:capability { dac_override sys_resource };
+allow systemd_user_session_type self:process setfscreate;
+allow systemd_user_session_type self:udp_socket create_socket_perms;
+allow systemd_user_session_type self:unix_stream_socket create_stream_socket_perms;
+allow systemd_user_session_type self:netlink_kobject_uevent_socket { bind create getattr setopt };
+
+allow systemd_user_session_type systemd_user_runtime_t:sock_file { create write };
+userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, sock_file)
```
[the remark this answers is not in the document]
pebenito: To make sure I understand, you're suggesting the filetrans should be `userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, dir, "systemd")`?

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1137,39 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+
+mount_list_runtime(systemd_user_session_type)
+
+storage_getattr_fixed_disk_dev(systemd_user_session_type)
+
+# for systemd to read udev status
+udev_read_pid_files(systemd_user_session_type)
+udev_list_pids(systemd_user_session_type)
+
+userdom_manage_user_runtime_dirs(systemd_user_session_type)
```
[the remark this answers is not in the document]
doverride: I know Stephen said that selinux is not optimized for name-based type transitions. but they provide real benefits. particularly for unconfined_t. With these you can to a large extend ensure that the labels of objects of unconfined users are consistent with the labels of  confined users. So that if you ever remap a user from unconfined to user_t for example, that in theory, you do not have to relabel the users home. But that means that you have to use name-based type transitions consistently where possible and that you add the type transition rules to unconfined as well. In dssp2 i have a huge list of these:

https://github.com/DefenSec/dssp2-standard/blob/master/policy/system/u/unconfined.cil

Expensive, yes, but it makes the labels so much more consistent

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1137,39 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
 seutil_read_file_contexts(systemd_update_done_t)
 
 systemd_log_parse_environment(systemd_update_done_t)
+
+#########################################
+#
+# User session (systemd --user) local policy
+#
+
+allow systemd_user_session_type self:capability { dac_override sys_resource };
```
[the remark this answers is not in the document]
pebenito: I don't think that worked, but will try again.

policy/modules/system/systemd.te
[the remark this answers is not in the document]
doverride: If you do not do the above then a `restorecon -RvF /home` will never be completely silent when you have a system with confined and unconfined users alike

policy/modules/system/init.if
```diff
@@ -650,6 +650,35 @@ interface(`init_domtrans',`
...
+## </param>
+#
+interface(`init_pgm_spec_daemon_domain',`
+	gen_require(`
+		type init_exec_t;
+	')
+
+	init_spec_daemon_domain($1, init_exec_t)
+')
+
```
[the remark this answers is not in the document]
pebenito: Good point about `system_r` especially.

policy/modules/system/systemd.te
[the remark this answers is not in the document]
doverride: Also note that users will/should be able to "control" units in /etc/systemd/user /usr/lib/systemd/user, So you may want to differentiate between system-wide system-units and user-units

policy/modules/system/systemd.te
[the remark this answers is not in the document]
doverride: I works here

policy/modules/system/systemd.te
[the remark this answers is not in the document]
doverride: > To make sure I understand, you're suggesting the filetrans should be `userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, dir, "systemd")`?

Yes. create /run/user/UID/systemd with a name-based type transition

policy/modules/system/systemd.if
```diff
@@ -1,5 +1,51 @@
...
+	#################################
+	#
+	# Local policy
+	#
+	# This domain is per-role because of the below transitions.
+	# See the sytemd --user section of systemd.te for the
+	# remainder of the rules.
+	corecmd_shell_domtrans($1_systemd_t, $3)
+	corecmd_bin_domtrans($1_systemd_t, $3)
+')
```
doverride: At some point you may want to consider a "home_bin_t" (or eqquivalent) for ~/.local/share/bin
Then transition here on that as well.

policy/modules/system/systemd.te
```diff
@@ -202,6 +203,10 @@ init_system_domain(systemd_update_done_t, systemd_update_done_exec_t)
 type systemd_update_run_t;
 files_type(systemd_update_run_t)
 
+type systemd_user_runtime_t;
+files_type(systemd_user_runtime_t)
```
doverride: Remember that systemd-logind wants to delete all of /run/user/UID when a user logs out. So you may want to create a userdom_user_runtime_content_type, so that you can use that to allow systemd-logind to delete all userdom_user_runtime_content_type

doverride: I would be surprise if refpolicy has not yet implemented this?

doverride: https://github.com/SELinuxProject/refpolicy/blob/master/policy/modules/system/userdomain.te#L78

so you want something like: userdom_user_runtime_content(system_user_runtime_t)
so that systemd-logind can delete it when you log out

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1137,39 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
 
 systemd_log_parse_environment(systemd_update_done_t)
+
+#########################################
+#
+# User session (systemd --user) local policy
+#
+
+allow systemd_user_session_type self:capability { dac_override sys_resource };
+allow systemd_user_session_type self:process setfscreate;
```
doverride: This is fortunate by the way, this means that systemd --user will automatically transtition where needed as long as you have the fc-specs set up properly (this does not include socket activation though: setsockfscreate)

if does however (fortunately) do setsockcreate

doverride: So i suppose the name-based auto type transition rules i mentioned earlier arent as important. As a matter of fact if you have your fc specs set up systemd should just create stuff with proper contexts automatically.

The thing is: we probably don't /shouldnt specify fc specs for content in /run/user/UID (i know i don't but i am not sure if refpolicy has any fc specs for content in /run/user/UID

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1137,39 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+allow systemd_user_session_type self:netlink_kobject_uevent_socket { bind create getattr setopt };
+
+allow systemd_user_session_type systemd_user_runtime_t:sock_file { create write };
+userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, sock_file)
+
+dev_write_sysfs_dirs(systemd_user_session_type)
+dev_read_sysfs(systemd_user_session_type)
+
+files_read_etc_files(systemd_user_session_type)
+files_list_usr(systemd_user_session_type)
```
doverride: I wonder what is it listing  here?

policy/modules/system/systemd.te
pebenito: This is just the early structure for the policy. I didn't have any socket activating user services to test on the system I'm working on.

policy/modules/system/systemd.if
doverride: allow callers to manage and relabel systend_runtime_user_t here (also allow callers to create systemd dir with systemd_runtime_user_t name-based type transition)

policy/modules/system/systemd.te
doverride: > This is fortunate by the way, this means that systemd --user will automatically transtition where needed as long as you have the fc-specs set up properly (this does not include socket activation though: setsockfscreate)
> 
> if does however (fortunately) do setsockcreate

Hmm, actually i might be wrong here. Maybe it does do the setsockcreate properly. The thing here is that when you add fc-specs at runtime that might affect systemd, then you need to do a `systemctl (--user) daemon-reexec. You see, systemd loads the ffc-spec when it starts and then it caches them. So any specs you add at runtime will not be acknowledged by the running systemd instance

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1136,42 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+allow systemd_user_session_type self:capability { dac_read_search sys_resource };
+dontaudit systemd_user_session_type self:capability dac_override;
+allow systemd_user_session_type self:process setfscreate;
+allow systemd_user_session_type self:udp_socket create_socket_perms;
+allow systemd_user_session_type self:unix_stream_socket create_stream_socket_perms;
+allow systemd_user_session_type self:netlink_kobject_uevent_socket { bind create getattr setopt };
+
+allow systemd_user_session_type systemd_user_runtime_t:dir manage_dir_perms;
+allow systemd_user_session_type systemd_user_runtime_t:sock_file { create write };
+userdom_user_runtime_filetrans(systemd_user_session_type, systemd_user_runtime_t, dir)
```
doverride: I see that you opted for a non-name-based type transtiion here. Not sure if that is a good idea.

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1136,42 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+dev_read_sysfs(systemd_user_session_type)
+
+files_read_etc_files(systemd_user_session_type)
+files_list_usr(systemd_user_session_type)
+
+fs_getattr_cgroup(systemd_user_session_type)
+fs_rw_cgroup_files(systemd_user_session_type)
+fs_manage_cgroup_dirs(systemd_user_session_type)
+
+init_dgram_send(systemd_user_session_type)
```
doverride: this is probably a dupe? systemd_user_session_type is already `systemd_log_parse_env_type` which implies `logging_send_syslog_msg()` which implies `init_dgram_send()` ?

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1136,42 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+init_signal(systemd_user_session_type)
```
doverride: You will probably end up with `init_signull()` and `init_sigchld()` here as well as i noted above (although i suppose the init_sigchld() and maybe even the `init_signull()` is probably allready implied by `init_pgm_spec_user_daemon_domain()`

policy/modules/system/systemd.te
```diff
@@ -1132,3 +1136,42 @@ files_var_filetrans(systemd_update_done_t, systemd_update_run_t, file, ".updated
...
+
+allow systemd_user_session_type self:capability { dac_read_search sys_resource };
+dontaudit systemd_user_session_type self:capability dac_override;
+allow systemd_user_session_type self:process setfscreate;
+allow systemd_user_session_type self:udp_socket create_socket_perms;
+allow systemd_user_session_type self:unix_stream_socket create_stream_socket_perms;
+allow systemd_user_session_type self:netlink_kobject_uevent_socket { bind create getattr setopt };
+
+allow systemd_user_session_type systemd_user_runtime_t:dir manage_dir_perms;
+allow systemd_user_session_type systemd_user_runtime_t:sock_file { create write };
```
doverride: Not sure if you want this added complexity but systemd maintains two sockets, for two different purposes.. "private" and "notify", With a name-based type transttion you would be able to diferentiate between the two.

policy/modules/system/init.if
```diff
@@ -650,6 +650,44 @@ interface(`init_domtrans',`
...
+	domain_type($1)
+	domain_entry_file($1, init_exec_t)
+
+	spec_domtrans_pattern(init_t, init_exec_t, $1)
+
+	allow init_t $1:process rlimitinh;
+
+	ifdef(`init_systemd',`
+		allow $1 init_t:unix_stream_socket { getattr read write ioctl };
+	')
```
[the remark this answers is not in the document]
doverride: Never mind the above,, i see you call `init_signal(systemd_user_session_type)` below

policy/modules/system/systemd.if
```diff
@@ -1,5 +1,65 @@
...
+	manage_lnk_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	manage_fifo_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	manage_sock_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+
+	relabel_dirs_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	relabel_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	relabel_lnk_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	relabel_fifo_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+	relabel_sock_files_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t)
+
```
doverride: nit, but you can make this more compact by leveraging permission sets instead:
```
allow $3 systemd_user_runtime_t:dir { manage_dir_perms relabel_dir_perms };
allow $3 systemd_user_runtime_t:file { manage_file_perms relabel_file_perms };
allow $3 systemd_user_runtime_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
allow $3 systemd_user_runtime_t:fifo_file { manage_fifo_file_perms relabel_fifo_file_perms };
allow $3 systemd_user_runtime_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
```

policy/modules/system/systemd.te
pebenito: This is just the initial policy. If we eventually find that it has to be name-based, then it can be added then.

policy/modules/system/init.if
```diff
@@ -650,6 +650,44 @@ interface(`init_domtrans',`
...
+	gen_require(`
+		type init_t, init_exec_t;
+	')
+
+	domain_type($1)
+	domain_entry_file($1, init_exec_t)
+
+	spec_domtrans_pattern(init_t, init_exec_t, $1)
+
+	allow init_t $1:process rlimitinh;
```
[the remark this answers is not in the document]
pebenito: I can buy the setsched, but what environment is being filtered such that noatsecure is required?

policy/modules/system/systemd.te
pebenito: I am aware of them. I'm open to having them separated in a future change.

policy/modules/system/systemd.te
pebenito: No, `logging_send_syslog_msg()` includes `kernel_dgram_send()`. Though this may be a system-specific thing. I'll remove it for now.

policy/modules/system/systemd.te
pebenito: I take that back. If you elaborate on this point, I could add it in this PR.

policy/modules/system/init.if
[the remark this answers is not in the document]
doverride: Ok then for now do the setsched and just keep in the back of your mind that if you see behavior that cannot be explained, that it may be the noatsecure. I am afraid I can no longer find any justification in my git history

policy/modules/system/systemd.te
doverride: This might explain a bit about the notify socket:

https://www.freedesktop.org/software/systemd/man/systemd-notify.html

So just because a service unit may have Type=notify or just because systemd-notify is used to notify systemd of status changes, that does not mean that these processes also have to connecto to systemd with the private socket

policy/modules/system/init.if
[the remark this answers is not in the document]
doverride: Actually, i suppose you can try it out. do a systemd-run --user env with and without the noatsecure? See if theres any changes.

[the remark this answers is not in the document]
doverride: i will try this see if i can refresh my memory....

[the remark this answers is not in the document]
doverride: I think its would make sense though that systemd --system cleans the environment when it executes systemd --user

policy/modules/system/systemd.te
pebenito: Are you suggesting three types, one for each socket plus the general `systemd_user_runtime_t`, or would it be sufficient to do a separate type just for notify?

policy/modules/system/init.if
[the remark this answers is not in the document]
doverride: I tried it out and without the noatsecure access the systemd-user instance cannot even be spawned in enforcing mode.

policy/modules/system/systemd.te
doverride: I just have a private type "systemd_user_runtime_private_t" for the "private" socket. The notify socket is just generic systemd_user_runtime_t just as the directory. That would be enough to differentiate between the notify socket and the private socket.

pebenito: I would have thought the notify socket would be the one with the different type, since that would theoretically be the one other domains interact with.

doverride:
```
$ ls -Z /run/user/1000/systemd/*
       wheel.id:wheel.role:systemd.user.tmpfs.user_tmpfs_file:s0 /run/user/1000/systemd/notify
     wheel.id:wheel.role:systemd.user.private.user_tmpfs_file:s0 /run/user/1000/systemd/private

/run/user/1000/systemd/transient:
wheel.id:wheel.role:systemd.user.unit_runtime.user_tmpfs_file:s0 dbus-:1.2-ca.desrt.dconf@0.service
wheel.id:wheel.role:systemd.user.unit_runtime.user_tmpfs_file:s0 dbus-:1.8-org.a11y.atspi.Registry@0.service

```
So something like this:

1. /run/user/UID/systemd:  systemd_user_runtime_t
2. /run/user/UID/systemd/notify: systemd_user_runtime_t
3. /run/user/UID/systemd/private: systemd_user_runtime_private_t
4. /run/user/UID/systemd/transient(/.*)? systemd_user_runtime_unit_t

doverride: or simplified:

    /run/user/UID/systemd(/.*)?: systemd_user_runtime_t
    /run/user/UID/systemd/private: systemd_user_runtime_private_t
    /run/user/UID/systemd/transient(/.*)? systemd_user_runtime_unit_t

doverride: > I would have thought the notify socket would be the one with the different type, since that would theoretically be the one other domains interact with.

I suppose. I dont think it matter much. Just as long as you can differentiate between private and notify.

doverride: But yes I you like this better:

```
/run/user/UID/systemd(/.*)? systemd_user_runtime_t
/run/user/UID/systemd/notify systemd_user_runtime_notify_t
/run/user/UID/systemd/transient(/.*)? systemd_user_runtime_unit_t
```

policy/modules/system/systemd.if
```diff
@@ -1,5 +1,59 @@
...
+## <param name="userdomain">
+##	<summary>
+##	The user domain for the role.
+##	</summary>
+## </param>
+#
+template(`systemd_role_template',`
+	gen_require(`
+		attribute systemd_user_session_type, systemd_log_parse_env_type;
+		type systemd_user_runtime_t;
```
doverride: You now should require the systemd_user_runtime_notify_t type

policy/modules/system/systemd.if
```diff
@@ -1,5 +1,59 @@
...
+	#################################
+	#
+	# Local policy
+	#
+
+	allow $3 systemd_user_runtime_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $3 systemd_user_runtime_t:file { manage_file_perms relabel_file_perms };
+	allow $3 systemd_user_runtime_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
+	allow $3 systemd_user_runtime_t:fifo_file { manage_fifo_file_perms relabel_fifo_file_perms };
+	allow $3 systemd_user_runtime_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
```
doverride: allow $3 systemd_user_runtime_notify_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };

policy/modules/system/systemd.if
```diff
@@ -1,5 +1,59 @@
...
+
+	# This domain is per-role because of the below transitions.
+	# See the sytemd --user section of systemd.te for the
+	# remainder of the rules.
+	corecmd_shell_domtrans($1_systemd_t, $3)
+	corecmd_bin_domtrans($1_systemd_t, $3)
```
doverride: I see that you did not add the setsched rlimitinh access for systemd --user to commands executed with a domain transition back to the calling shell domain (no problem just keep in mind that resource limits  may or may not work reliably).

`allow $1_systemd_t $3:process { setsched rlimitinh };`

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Properties

Example: `systemd-run --user -p LimitNPROC=1 ulimit -a`

pebenito: This was an accidental omission. I have it in my test policy but missed committing it.

doverride:
This was all i could find. Looks good now as far as i am concerned.

doverride:
It is going to be really interesting though to see how and if this will work out. Especially when dbus comes into play. The issue is that afaik you cannot reliably declare types in optional policy and that is what these templates that dbus and now systemd_user use are all about. We'll see what happens when the two meet.

pebenito:
No need to relook at this last push, I just moved some lines.

PR #61, opened 2019-08-17
entropyd: make haveged service work on Debian

fishilico:
On Debian, haveged fails to start with "haveged: Couldn't open random device: Permission denied". strace shows:

    openat(AT_FDCWD, "/dev/random", O_RDWR) = -1 EACCES (Permission denied)

audit.log has:

    type=AVC msg=audit(1566048720.132:1338): avc:  denied  { search }
    for  pid=20235 comm="haveged" name="/" dev="tmpfs" ino=76666
    scontext=system_u:system_r:entropyd_t
    tcontext=system_u:object_r:tmpfs_t tclass=dir permissive=0

With systemd, /dev is a temporary filesystem (tmpfs_t), so haveged needs the search permission to it in order to open /dev/random.

While at it, add a file context of haveged's systemd unit file.

### DIFF
--- a/policy/modules/services/entropyd.fc
+++ b/policy/modules/services/entropyd.fc
@@ -1,5 +1,7 @@
 /etc/rc\.d/init\.d/((audio-entropyd)|(haveged))	--	gen_context(system_u:object_r:entropyd_initrc_exec_t,s0)
 
+/usr/lib/systemd/system/haveged.*\.service	--	gen_context(system_u:object_r:entropyd_unit_t,s0)
+
 /usr/bin/audio-entropyd	--	gen_context(system_u:object_r:entropyd_exec_t,s0)
 /usr/bin/haveged	--	gen_context(system_u:object_r:entropyd_exec_t,s0)
 
--- a/policy/modules/services/entropyd.te
+++ b/policy/modules/services/entropyd.te
@@ -21,6 +21,9 @@ init_daemon_domain(entropyd_t, entropyd_exec_t)
 type entropyd_initrc_exec_t;
 init_script_file(entropyd_initrc_exec_t)
 
+type entropyd_unit_t;
+init_unit_file(entropyd_unit_t)
+
 type entropyd_var_run_t;
 files_pid_file(entropyd_var_run_t)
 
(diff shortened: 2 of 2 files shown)


fishilico:
There has been some discussion on the IRC channel about `fs_search_tmpfs(entropyd_t)`. As it is caused by `PrivateDevices=yes` in the unit files, it is specific to systemd and should at least be in a ``ifdef(`init_systemd'`` block.

Moreover, as this issue is shared by all services that use `PrivateDevices=yes`, an attribute could be introduced with two rules, instead of discovering all the services that need these rules one at a time:
```
fs_search_tmpfs(systemd_system_private_devices_daemon_type)
fs_read_tmpfs_symlinks(systemd_system_private_devices_daemon_type)
```

For information, I also took a look at systemd's source code in order to find out whether it would be possible to get the `/dev` directory labeled correctly in the mount namespace of services using `PrivateDevices=yes`. There is a function, `mount_private_dev`, that uses `mkdir_p_label` to create `/run/systemd/unit-root/dev` (https://github.com/systemd/systemd/blob/v242/src/core/namespace.c#L753), but the label disappears under the mounted directory when `mount` a few lines later. The directory which is mount-moved comes from a tmpfs created on a temporary dir with pattern `/tmp/namespace-dev-XXXXXX` (https://github.com/systemd/systemd/blob/v242/src/core/namespace.c#L669). This is why /dev gets labeled as `tmpfs_t` in the namespace. This contrast with `/tmp` when `PrivateTmp=yes`, as the `/tmp` folder is then bind-mounted with a directory which is created with label `tmp_t` thanks to a file context pattern in the policy for `/tmp/systemd-private-[^/]+/tmp`.

In summary: changing systemd in order to get it to label `/dev` as `device_t` is not as simple as adding some rules to the policy and would require changes in the options of the `mount` call. I do not want to spend time motivating this change before systemd's developers, so I am looking for other solutions that boil down into allowing `tmpfs_t` access to services in a way which is similar to `device_t`.

pebenito:
Sadly systemd is incredibly complicated. Since it is such a pain to try to get something upstreamed into systemd, this is not a horrible compromise. It simply needs to be documented in the policy. I think there is a potential to have interfaces that mirror the systemd unit options. For example, systemd_PrivateDevices(entropyd_t), systemd_PrivateTmp(), etc.

fishilico:
This Pull Request has been split between #63 and #64.

PR #63, opened 2019-08-18
systemd: introduce a new attribute for services using PrivateDevices=yes

fishilico:
When a service is configured with `PrivateDevices=yes`, its `/dev` directory has label `tmpfs_t`. This requires allowing more accesses in order for the service to use `/dev`.

This is related to https://github.com/SELinuxProject/refpolicy/pull/61

### DIFF
--- a/policy/modules/services/entropyd.te
+++ b/policy/modules/services/entropyd.te
@@ -64,6 +64,12 @@ miscfiles_read_localization(entropyd_t)
 userdom_dontaudit_use_unpriv_user_fds(entropyd_t)
 userdom_dontaudit_search_user_home_dirs(entropyd_t)
 
+ifdef(`init_systemd', `
+	optional_policy(`
+		systemd_PrivateDevices(entropyd_t)
+	')
+')
+
 tunable_policy(`entropyd_use_audio',`
 	dev_read_sound(entropyd_t)
 	dev_write_sound(entropyd_t)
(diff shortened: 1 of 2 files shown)


fishilico:
I changed the name of the interface and removed the attribute.

fishilico:
My patch currently breaks building with `SYSTEMD=n`, as
```
optional_policy(`
	systemd_PrivateDevices(entropyd_t)
')
```

is reduced to:
```
optional {
}
```

The issue here is that `systemd_PrivateDevices` does not require anything but a compile-time boolean to be set. In order for this to work, dropping the `optional_policy` is possible (calling `systemd_PrivateDevices` does not make a module depends on `systemd` module).

How would you like this to be fixed?

pebenito:
Simple fix. Remove the ifdef. That check should be on caller side.

pebenito:
To clarify, since this is the systemd module, all the interfaces should be assuming init_systemd=y. If that check is necessary it should be done on caller side.

fishilico:
Thanks, I moved the ifdef to the caller side.

policy/modules/services/entropyd.te
```diff
@@ -83,3 +83,9 @@ optional_policy(`
 optional_policy(`
 	udev_read_db(entropyd_t)
 ')
+
+ifdef(`init_systemd', `
```
pebenito: This should move to line 66.

fishilico: Moved, thanks.

PR #82, opened 2019-09-01
cups: add a slash to match /opt/brother/Printers/

fishilico:
The pattern `/opt/brother/Printers(.*/)?inf(/.*)?` matches the content of directories such as `/opt/brother/Printersinf/`, which seems buggy. On several systems, `/opt/brother/Printers/` is a directory that contains directories named as printer models.

Add a `/` before `(.*/)?` in order to make sure subdirectories of `/opt/brother/Printers` named `inf` are matched by the pattern.

### DIFF
--- a/policy/modules/services/cups.fc
+++ b/policy/modules/services/cups.fc
@@ -18,7 +18,7 @@
 
 /etc/printcap.*	--	gen_context(system_u:object_r:cupsd_rw_etc_t,s0)
 
-/opt/brother/Printers(.*/)?inf(/.*)?	gen_context(system_u:object_r:cupsd_rw_etc_t,s0)
+/opt/brother/Printers/(.*/)?inf(/.*)?	gen_context(system_u:object_r:cupsd_rw_etc_t,s0)
 /opt/gutenprint/ppds(/.*)?	gen_context(system_u:object_r:cupsd_rw_etc_t,s0)
 
 /usr/bin/hp-[^/]+	--	gen_context(system_u:object_r:hplip_exec_t,s0)
--- a/policy/modules/services/cups.fc
+++ b/policy/modules/services/cups.fc
@@ -18,7 +18,7 @@
 
 /etc/printcap.*	--	gen_context(system_u:object_r:cupsd_rw_etc_t,s0)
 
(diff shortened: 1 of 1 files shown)


pebenito:
Shouldn't it be `/opt/brother/Printers/([^/]+/)?inf(/.*)?`

fishilico:
I agree and I have added a commit in order to keep in the git history why this was changed, in case this broke something somewhere.

By the way, no pattern in the policy uses `/([^/]+/)?`, so testing/check_fc_files.py will need to be modified, with something like:

```patch
--- a/testing/check_fc_files.py
+++ b/testing/check_fc_files.py
@@ -129,6 +129,10 @@ def analyze_fc_file(fc_path):
                 print(f"{prefix}using (.*/)? without a previous slash could be a bug in {path} as it can match the empty string, please use /(.*/)? instead")  # noqa
                 retval = False
 
+            if re.search(r'[^/]\(\[\^/\]\+/\)\?', path):
+                print(f"{prefix}using ([^/]+/)? without a previous slash could be a bug in {path} as it can match the empty string, please use /([^/]+/)? instead")  # noqa
+                retval = False
+
             if re.search(r'[^/]\(\.\*/\)\*', path):
                 print(f"{prefix}using (.*/)* without a previous slash could be a bug in {path} as it can match the empty string, please use /(.*/)* instead")  # noqa
                 retval = False
@@ -202,6 +206,8 @@ def analyze_fc_file(fc_path):
                 retval = False
 
             # Remove optional directories and filename parts
+            reduced_path = reduced_path.replace('/([^/]+/)?', '/')
+            reduced_path = reduced_path.replace('(/[^/]+)?/', '/')
             reduced_path = reduced_path.replace('[^/]*', '')
             reduced_path = reduced_path.replace('[^/]+', '∞')
             reduced_path = reduced_path.replace('[^/-]+', '∞')
```

How would you want to proceed for this modification? I can either add a new commit to https://github.com/SELinuxProject/refpolicy/pull/74 or wait for it to be merged and create a later pull request with this and the Travis-CI configuration to run the script during CI tests.

pebenito:
Please add new commit to #74

PR #86, opened 2019-09-04
Display a list of useful policy modules from the .fc files

fishilico:
It happens that refpolicy users want to reduce the number of modules that are used on their systems to a minimal set. Doing so by hand is prone to errors, because one might forget that a piece of software is installed on the target system, in which case files and processes will be mislabeled.

In order to help detecting such issues, I wrote a script a few years ago which converted the regular expressions in `.fc` files to `glob()` patterns (aka. the patterns used by several shells to match paths). As this issue came again recently on IRC (#selinux channel), I cleaned up the script and suggest adding it to the main repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,3 +100,4 @@ script:
   - make DESTDIR=${HOME}/tmp install-src
   - make DESTDIR=${HOME}/tmp install-docs
   - make DESTDIR=${HOME}/tmp install-appconfig
+  - ./testing/get_useful_modules.py
--- a/testing/get_useful_modules.py
+++ b/testing/get_useful_modules.py
@@ -1,0 +1,673 @@
+#!/usr/bin/env python3
+# -*- coding:UTF-8 -*-
+# Copyright (C) 2014-2019 Nicolas Iooss
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


pebenito:
I'm struggling with this. I'm not sure that this belongs in refpolicy since it's not actually required to build refpolicy.

fishilico:
I do not have a strong opinion about whether to include this script in refpolicy or not. The main reason why I created this Pull Request was to make it easier for refpolicy users to choose which modules to load, and I agree it is not required at all to build refpolicy or to find bugs in it.

Here is a list of Pros/Cons about the script in this Pull Request.

Pros:
* It helps people who want to reduce the size of their SELinux policy.
* It contains code that parses and expands the patterns in `.fc` files that can be re-used in other projects.
* It helps detecting issues such as unintended missing `optional_policy` blocks (like https://github.com/SELinuxProject/refpolicy/pull/87).

Cons:
* It is tied to refpolicy and its `.fc` files. A possible "version 2" of this script could include parsing `.pp` and `.cil` files. This would make it less likely to "belong to refpolicy".
* It adds code to maintain, which is not directly related to building refpolicy.
* It implements heuristics for converting regular expressions to `glob` patterns, that can be broken.

Anyway, if you reject this Pull Request, I will keep this script in my public home-files repository in order to be able to easily share it with people who may be interested with it. So it won't be lost.

PR #87, opened 2019-09-04
portage: really make consoletype module optional

fishilico:
All callers of `consoletype_exec()` put it in an `optional_policy()` block but `portage`. This makes `consoletype` module mandatory when module `portage` is loaded, even when `consoletype` is not installed.

Fix this issue by introducing an `optional_policy()` block.

### DIFF
--- a/policy/modules/admin/portage.te
+++ b/policy/modules/admin/portage.te
@@ -130,14 +130,16 @@ miscfiles_read_localization(gcc_config_t)
 
 userdom_use_user_terminals(gcc_config_t)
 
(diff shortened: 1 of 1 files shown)


pebenito:
@perfinion your thoughts? Is this actually optional for portage?

perfinion:
I'm good with this change. adding optional_policy makes things easier to load so this would just make upgrades / whatever easier and I don't see any downsides.

I did a quick look at gcc-config and I cant find a call to consoletype but I'd be hesitant to remove the rule completely.

PR #118, opened 2019-10-03
re-implement fc_sort in python

cgzones:
`fc_sort` is the only/last build tool that requires a C compiler

Re-implement it in python, so that gcc dependencies can be dropped

The output of the C and the python version differ slightly in the order of equally specific file contexts

old:
```
    /.*		system_u:object_r:default_t
    /sys(/.*)?		system_u:object_r:sysfs_t
    /mnt(/[^/]*)		-l	system_u:object_r:mnt_t
    /mnt(/[^/]*)?		-d	system_u:object_r:mnt_t
    /opt/.*		system_u:object_r:usr_t
    /var/.*		system_u:object_r:var_t
    /usr/.*		system_u:object_r:usr_t
    /srv/.*		system_u:object_r:var_t
    /tmp/.*		<<none>>
    /run/.*		<<none>>
    /dev/.*		system_u:object_r:device_t
    /etc/.*		system_u:object_r:etc_t
```
new:
```
    /.*		system_u:object_r:default_t
    /sys(/.*)?		system_u:object_r:sysfs_t
    /mnt(/[^/]*)		-l	system_u:object_r:mnt_t
    /mnt(/[^/]*)?		-d	system_u:object_r:mnt_t
    /dev/.*		system_u:object_r:device_t
    /etc/.*		system_u:object_r:etc_t
    /opt/.*		system_u:object_r:usr_t
    /run/.*		<<none>>
    /srv/.*		system_u:object_r:var_t
    /tmp/.*		<<none>>
    /usr/.*		system_u:object_r:usr_t
    /var/.*		system_u:object_r:var_t
```


support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+            if char == '\\':
+                skip_escaped = True
+
+            if not self.meta:
+                self.stem_len += 1
+
+            self.str_len += 1
+
+
+def compare_fcds(a, b):
```
pebenito: My preference would have this as part of the FileContext class, then the sorted() use would work naturally. I think sorted uses `__lt__` and `__eq__` and maybe `__ne__`.

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+    parser.add_argument('infile', metavar='INFILE', nargs=1, type=Path,
+                        help='input file of the original file context definitions')
+    parser.add_argument('outfile', metavar='OUTFILE', nargs='?', type=Path, default=None,
+                        help='output file for the sorted file context definitions')
+    args = parser.parse_args(argv)
+
+    file_context_definitions = []
+
+    # Parse the input file
+    with args.infile[0].open('r') as fd:
```
pebenito: If you remove the nargs=1 then you can just do `args.infile`.

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+            file_context_definitions.append(FileContext(path, ftype, context))
+
+    # Sort
+    file_context_definitions.sort(key=functools.cmp_to_key(compare_fcds))
+
+    # Print output, either to file or if no output file given to stdout
+    fd = args.outfile.open('w') if args.outfile else sys.stdout
+    for fcd in file_context_definitions:
+        if fcd.file_type:
+            fd.write('%s\t\t%s\t%s\n' % (fcd.path, fcd.file_type, fcd.context,))
```
pebenito: I'd prefer the part that creates the string representation be the FileContext `__str__` method. Also, use `.format()` instead of `%`

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+
+            path, ftype, context = matches.group('path', 'ftype', 'context')
+
+            file_context_definitions.append(FileContext(path, ftype, context))
+
+    # Sort
+    file_context_definitions.sort(key=functools.cmp_to_key(compare_fcds))
+
+    # Print output, either to file or if no output file given to stdout
+    fd = args.outfile.open('w') if args.outfile else sys.stdout
```
pebenito: please use a context manager: `with open(filename, "w") as fd:`. Then you don't need the `close()`.

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+    if not a.file_type and b.file_type:
+        return -1
+    if not b.file_type and a.file_type:
+        return 1
+
+    # If none of the above conditions were satisfied, then a and b are equally specific
+    return 0
+
+
+def main(argv=None):
```
pebenito: This script is so small, please just put all of this function in `if __name__ == '__main__':`

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+            # Ignore comments and empty lines
+            if not line or line.startswith('#'):
+                continue
+
+            matches = re.match(r'^(?P<path>\S+)\s+(?P<ftype>-.)?\s*(?P<context>.+)$', line)
+            if matches is None:
+                print("%s:%d: unable to parse a file context line: %s" % (args.infile[0], lineno, line,))
+                exit(1)
+
+            path, ftype, context = matches.group('path', 'ftype', 'context')
```
pebenito: I think this part should be part of the FileContext `__init__`.

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for char in self.path:
+            if skip_escaped:
+                skip_escaped = False
+                continue
+
+            if char in ('.', '^', '$', '?', '*', '+', '|', '[', '(', '{',):
+                self.meta = True
+            if char == '\\':
+                skip_escaped = True
```
pebenito: It might be simpler to use `str.find()`. Then when you get an index, you do index-1 to see if it is escaped. That will simultaneously allow you to determine the stem length too. If you enumerate() too, you can exit the loop once the charater count > stem_len.

I also don't think that the `self.meta` is really needed. `If stem_len:` would be true if there are metacharacters.

support/fc_sort.py
```diff
@@ -0,0 +1,153 @@
...
+
+            if not self.meta:
+                self.stem_len += 1
+
+            self.str_len += 1
```
pebenito: `len(self.path)`

support/fc_sort.py
cgzones: done

support/fc_sort.py
cgzones: done

support/fc_sort.py
cgzones: done

support/fc_sort.py
cgzones: done

support/fc_sort.py
cgzones: done

support/fc_sort.py
cgzones: but then one would need pass the filename and line number for error output too

support/fc_sort.py
cgzones: that would be a functional change, cause previously escaped character did not count

support/fc_sort.py
cgzones: I think that would complicate the logic: one would need to rescan with `str.find()` if `index-1` is an escape character

pebenito: Ah yes, you're right of course.

support/fc_sort.py
pebenito: Ah, yes you're right of course.

support/fc_sort.py
pebenito: Ok.

support/fc_sort.py
pebenito: The __init__ would have the matches = re.match(). Raise ValueError if matches == None. If not, then do the matches.group(). Then caller would catch ValueError, print error message.

cgzones: done

PR #119, opened 2019-10-03
travis: force the use of python3.5

cgzones:
python3.5 is the oldest available one on bionic (though refpolicy requires only 3.4)

Also print the python3 (not 2) version during build


cgzones:
this might need an update, since the new check_fc_files linter by Nicolas needs a newer python version

PR #121, opened 2019-10-05
dbus: allow systemd services to watch the DBUS system socket

fishilico:
Since systemd 243 (commit https://github.com/systemd/systemd/commit/8a5cd31e5fa0b1313a196b902e4b3c4603e7dfdf), several systemd's services use `inotify_add_watch()` if connecting to the DBUS system socket failed, in order to wait for the socket to appear.

On Debian 10, this produces logs such as:

    type=AVC msg=audit(1570213116.564:18): avc:  denied  { read } for
    pid=255 comm="systemd-resolve" name="system_bus_socket" dev="tmpfs"
    ino=13104 scontext=system_u:system_r:systemd_resolved_t
    tcontext=system_u:object_r:system_dbusd_runtime_t tclass=sock_file
    permissive=1

    type=SYSCALL msg=audit(1570213116.564:18): arch=c000003e syscall=254
    success=yes exit=4 a0=f a1=55d645b99960 a2=2000d84 a3=1 items=0
    ppid=1 pid=255 auid=4294967295 uid=103 gid=104 euid=103 suid=103
    fsuid=103 egid=104 sgid=104 fsgid=104 tty=(none) ses=4294967295
    comm="systemd-resolve" exe="/usr/lib/systemd/systemd-resolved"
    subj=system_u:system_r:systemd_resolved_t key=(null)

    type=PROCTITLE msg=audit(1570213116.564:18):
    proctitle="/lib/systemd/systemd-resolved"

(syscall 254 on x86-64 is `inotify_add_watch()` and mode `a2=2000d84` means `IN_DELETE_SELF|IN_MOVE_SELF|IN_ATTRIB|IN_CREATE|IN_MOVED_TO|IN_DONT_FOLLOW`).

As this does not extend to all DBUS users, but only to those which are using `libsystemd-shared-243.so`, introduce a new interface to allow this access instead of adding more statements to `dbus_system_bus_client()`.

### DIFF
--- a/policy/modules/services/dbus.if
+++ b/policy/modules/services/dbus.if
@@ -438,6 +438,34 @@ interface(`dbus_spec_session_domain',`
 
 ########################################
 ## <summary>
+##	Watch the creation of the Unix socket
+##	of the DBUS system bus and its parent
+##	directories with inotify.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dbus_watch_system_bus',`
+	gen_require(`
+		type system_dbusd_runtime_t;
+	')
+
+	files_list_pids($1)
+
+	# Do not use list_dirs_pattern() and read_sock_files_pattern(), as
+	# inotify_add_watch() only needs the read permission.
+
+	# For /run/dbus
+	allow $1 system_dbusd_runtime_t:dir {search_dir_perms read};
+	# For /run/dbus/system_bus_socket
+	allow $1 system_dbusd_runtime_t:sock_file read;
+')
+
+########################################
+## <summary>
 ##	Acquire service on the DBUS system bus.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/services/ntp.te
+++ b/policy/modules/services/ntp.te
@@ -141,6 +141,7 @@ ifdef(`init_systemd',`
 
 	dbus_system_bus_client(ntpd_t)
 	dbus_connect_system_bus(ntpd_t)
+	dbus_watch_system_bus(ntpd_t)
 	init_dbus_chat(ntpd_t)
 	init_get_system_status(ntpd_t)
 	init_list_unit_dirs(ntpd_t)
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -699,6 +699,7 @@ systemd_log_parse_environment(systemd_networkd_t)
 optional_policy(`
 	dbus_system_bus_client(systemd_networkd_t)
 	dbus_connect_system_bus(systemd_networkd_t)
+	dbus_watch_system_bus(systemd_networkd_t)
 
 	systemd_dbus_chat_hostnamed(systemd_networkd_t)
 ')
@@ -993,6 +994,7 @@ systemd_read_networkd_runtime(systemd_resolved_t)
 optional_policy(`
 	dbus_connect_system_bus(systemd_resolved_t)
 	dbus_system_bus_client(systemd_resolved_t)
+	dbus_watch_system_bus(systemd_resolved_t)
 ')
 
 #########################################


fishilico:
Actually there have been some works in the kernel to make `inotify` no longer checks for `read` permission, but for new ones introduced by commit https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ac5656d8a4cdd93cd2c74355ed12e5617817e0e7 (merged in 5.4-rc1, not yet available as a distribution kernel):
```c
#define COMMON_FILE_PERMS COMMON_FILE_SOCK_PERMS, "unlink", "link", \
    "rename", "execute", "quotaon", "mounton", "audit_access", \
	"open", "execmod", "watch", "watch_mount", "watch_sb", \
	"watch_with_perm", "watch_reads"
```

Before accepting this PR, a naming convention needs to be chosen for the future interfaces that will provide rules for these accesses.

Moreover, there are at least two options in order to handle these new permissions:

* put permissions `read` and `watch_*` both in a `dbus_watch_...` interface, even though `read` will not be needed with Linux>=5.4
* create two interfaces: a "clean" one with only `watch_*`, and another one for the "`read` meaning `watch` because inotify is used".

Another option might consist in introducing a tunable or a boolean meaning "enable rules for kernels older than 5.4". I do not know whether this would be more acceptable than the other suggestions.

Last but not least, a possible option may be "ignore these reads now, and just start addressing them only kernels that support them properly" (and reject this Pull Request). I am fine with this option too.

doverride:
On Sat, Oct 05, 2019 at 01:23:45PM -0700, Nicolas Iooss wrote:
> Actually there have been some works in the kernel to make `inotify` no longer checks for `read` permission, but for new ones introduced by commit https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ac5656d8a4cdd93cd2c74355ed12e5617817e0e7 (merged in 5.4-rc1, not yet available as a distribution kernel):
> ```c
> #define COMMON_FILE_PERMS COMMON_FILE_SOCK_PERMS, "unlink", "link", \
>     "rename", "execute", "quotaon", "mounton", "audit_access", \
> 	"open", "execmod", "watch", "watch_mount", "watch_sb", \
> 	"watch_with_perm", "watch_reads"
> ```
> 
> Before accepting this PR, a naming convention needs to be chosen for the future interfaces that will provide rules for these accesses.
> 
> Moreover, there are at least two options in order to handle these new permissions:
> 
> * put permissions `read` and `watch_*` both in a `dbus_watch_...` interface, even though `read` will not be needed with Linux>=5.4
> * create two interfaces: a "clean" one with only `watch_*`, and another one for the "`read` meaning `watch` because inotify is used".
> 
> Another option might consist in introducing a tunable or a boolean meaning "enable rules for kernels older than 5.4". I do not know whether this would be more acceptable than the other suggestions.
> 
> Last but not least, a possible option may be "ignore these reads now, and just start addressing them only kernels that support them properly" (and reject this Pull Request). I am fine with this option too.


The introduction of the fsnotify perms in Linux 5.4 will not make the current "read" events related to fsnotify go away in my tests, and so regardless of kernel, these permissions will still need to be added.

The question then remains how to implement fsnotify (naming convention). In my dssp2 policy i decided to treat every "watch" permission separately.
I could have introduced macros like "*.watchr_dirs"  that would imply both the read as well as the watch on the dir, but i decided not to.
I am just going to settle for now on separate "*.list_dirs" and "*.watch_dirs" eventhough it seems obvious that the "list" is related to the "watch"
The thing is that sometimes theres a watch on a object that is either already read for another reason, or is already managed.
In that scenario i would end up appending a separate "*.watch.*" anyway. So i figure that i might as well consistently not group watch-related perms.

> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/SELinuxProject/refpolicy/pull/121#issuecomment-538685635

-- 
Dominick Grift

stephensmalley:
Just to confirm what @doverride said, inotify_add_watch will still check read permission; the watch* permissions are in addition to the existing inode_permission() check, not a replacement for it. FWIW, Android policy has added watch and watch_reads permissions to its macros for reading files and directories so that they are allowed whenever read is allowed, but limiting watch_with_perm, watch_mount, and watch_sb to only where required.

doverride:
On Mon, Oct 07, 2019 at 06:42:12AM -0700, stephensmalley wrote:
> Just to confirm what @doverride said, inotify_add_watch will still check read permission; the watch* permissions are in addition to the existing inode_permission() check, not a replacement for it. FWIW, Android policy has added watch and watch_reads permissions to its macros for reading files and directories so that they are allowed whenever read is allowed, but limiting watch_with_perm, watch_mount, and watch_sb to only where required.

Yes i noticed that too. I do not do that in dssp2. all watch perms need to be allowed on a case by case basis and the reads are treated as seperate accesses as well
That way i get least privilege and have the flexibility to mix and match. I still have to see how that approach will pan out in the long term but for now it seems do-able

> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/SELinuxProject/refpolicy/pull/121#issuecomment-539017134

-- 
Dominick Grift

pebenito:
@stephensmalley is there a revocation issue for the watches, similar to mmaps? If so I'd probably go in the direction that we took the `map` permission. If not, then I'd probably lean towards the Android style of handling this.

stephensmalley:
> @stephensmalley is there a revocation issue for the watches, similar to mmaps? If so I'd probably go in the direction that we took the `map` permission. If not, then I'd probably lean towards the Android style of handling this.

Yes, permission is only checked when the watch is set, so if the file is relabeled or policy changes after the watch is set, the watcher will still get notifications for as long as it likes.

pebenito:
The watch access vectors are in the policy now. I'm leaning towards names like dbus_watch_system_bus_socket(). We should take a similar direction as with map; because of the revocation issue, I don't want unnecessary watches littering the policy.

fishilico:
This PR has been overseeded by commit f064342f413f ("systemd: Add filesystem watches."), which added:
```
dbus_watch_system_bus_runtime_dirs(systemd_networkd_t)
dbus_watch_system_bus_runtime_named_sockets(systemd_networkd_t)

dbus_watch_system_bus_runtime_dirs(systemd_resolved_t)
dbus_watch_system_bus_runtime_named_sockets(systemd_resolved_t)
```

I am closing this PR, as my commits are no longer fully relevant. If I encounter issues again, I will open a new one.

PR #122, opened 2019-10-11
Fix several stuff to run GNOME in debian unstable

bigon:


policy/modules/admin/alsa.fc
```diff
@@ -4,6 +4,7 @@ HOME_DIR/\.asoundrc				--	gen_context(system_u:object_r:alsa_home_t,s0)
 /etc/asound\.conf				--	gen_context(system_u:object_r:alsa_etc_t,s0)
 
 /run/alsa(/.*)?						gen_context(system_u:object_r:alsa_runtime_t,s0)
+/run/alsactl.pid				--	gen_context(system_u:object_r:alsa_runtime_t,s0)
```
fishilico: This dot needs to be escaped.
This is why Travis-CI test failed, https://travis-ci.org/SELinuxProject/refpolicy/jobs/596576535 contains:
```
/home/travis/build/SELinuxProject/refpolicy/policy/modules/admin/alsa.fc:7:
unescaped dot still present /run/alsactl.pid after being reduced to /run/alsactl.pid
(suggestion: use \. to match a dot, or a charset like [^/]))
```

bigon: OK this is fixed now

pebenito:
From the list discussion it seems this should be closed(?)

doverride:
On Thu, Oct 31, 2019 at 12:13:02AM -0700, Chris PeBenito wrote:
> From the list discussion it seems this should be closed(?)

AFAIK this series can be merged with the exception of the colord related patch

> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/SELinuxProject/refpolicy/pull/122#issuecomment-548243633

-- 
Dominick Grift

bigon:
Let me update the merge request to remove that one

bigon:
I also removed 8200aa8c74d8f6a7fc0fd7db735557d775dad4d3 because this is actually only needed in debian, in other distro `colord-sane` is labeled `bin_t` as they are located in `/usr/libexec`, so maybe labeling it `bin_t` is a better idea?

bigon:
Sorry for the mess with git...

PR #124, opened 2019-10-24
Add watch access vectors.

pebenito:
Signed-off-by: Chris PeBenito <chpebeni@linux.microsoft.com>

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,3 +1,6 @@
+Reference Policy has the following runtime requirements:
+	* Linux kernel >= 2.6.34
+
 Reference Policy has the following build requirements:
 	* SELinux userspace 2.8
 	* Python >= 3.4
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -28,6 +28,9 @@ common file
 	swapon
 	quotaon
 	mounton
+	open
+	audit_access
+	execmod
 }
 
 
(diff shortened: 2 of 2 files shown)


pebenito:
If I move the permissions to the common, is the breakage only for kernels that don't have the dynamic class/perm mapping? i.e. super old kernels?

stephensmalley:
Yes, kernels < 2.6.33, < RHEL 6.  I guess refpolicy already broke compatibility with those kernels when it added the map permission to common file and socket.

stephensmalley:
NB Dominick noted that changes to common/class file also broke old crond still shipping in some distros (Debian) that were still using the hardcoded FILE__ENTRYPOINT definition.  However, that too would have broken with the map permission change, so nothing new there, and I believe a patch was submited for cron.

pebenito:
Thanks Stephen for jogging my memory.

INSTALL
```diff
@@ -1,3 +1,6 @@
+Reference Policy has the following runtime requirements:
+	* Linux kernel >= 2.6.34
+
```
stephensmalley: 2.6.33

policy/flask/access_vectors
```diff
@@ -28,6 +28,9 @@ common file
 	swapon
 	quotaon
 	mounton
+	open
+	audit_access
+	execmod
```
stephensmalley: kernel and fedora have:
audit_access
open
execmod

PR #125, opened 2019-10-28
bugfix/123: allow xserver_t self:capability chown

mrl5:
on Funtoo 1.4 when xdm is started by OpenRC it additionally wants to

type=AVC msg=audit(1571776002.581:399): avc: denied { chown } for pid=6225 comm="X" capability=0 scontext=system_u:system_r:xserver_t tcontext=system_u:system_r:xserver_t tclass=capability permissive=0
type=AVC msg=audit(1571776002.729:400): avc: denied { chown } for pid=6225 comm="X" capability=0 scontext=system_u:system_r:xserver_t tcontext=system_u:system_r:xserver_t tclass=capability permissive=0

this does not occur if this is run by root: /etc/init.d/xdm start

more info:
- https://bugs.funtoo.org/browse/FL-6779
- https://bugs.gentoo.org/698112

### DIFF
--- a/policy/modules/services/xserver.te
+++ b/policy/modules/services/xserver.te
@@ -632,8 +632,7 @@ allow xserver_t input_xevent_t:x_event send;
 # execheap needed until the X module loader is fixed.
 # NVIDIA Needs execstack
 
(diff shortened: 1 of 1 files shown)


mrl5:
... or perhaps it should be done as a boolean

pebenito:
Since this may be a distro-specific rule, please work this through the Gentoo folks.

PR #129, opened 2019-11-18
Fix SyntaxError

alexa-v2:
This fixes the following error while building on Arch Linus with python version 3.8.0:

```
Updating policy/booleans.conf and policy/modules.conf
python3 -t -t -E -W error support/sedoctool.py -b policy/booleans.conf -m policy/modules.conf -x doc/policy.xml
  File "support/sedoctool.py", line 269
    if desc.data is not '':
       ^
SyntaxError: "is not" with a literal. Did you mean "!="?
make: *** [Makefile:405: conf.intermediate] Error 1
==> ERROR: A failure occurred in prepare().
    Aborting...
```

### DIFF
--- a/support/sedoctool.py
+++ b/support/sedoctool.py
@@ -266,7 +266,7 @@ def format_html_desc(node):
 	desc_buf = ''
 	for desc in node.childNodes:
 		if desc.nodeName == "#text":
-			if desc.data is not '':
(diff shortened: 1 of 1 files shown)


pebenito:
Thanks for the patch, but I fixed this in a slightly different way. See 0bfd138.

PR #151, opened 2020-01-03
kernel/corecommands: fix the label of xfce4 helpers

bauen1:
Even if they live under `/usr/lib/$triplet/xfce4/` instead of `/usr/lib/xfce4/`

Sources:
https://packages.debian.org/buster/amd64/tumbler/filelist
https://packages.debian.org/buster/all/libexo-helpers/filelist
https://packages.debian.org/buster/amd64/libexo-1-0/filelist
https://packages.debian.org/buster/amd64/libexo-2-0/filelist (for some reason it does not list exo-helper-2 but `dpkg -S` and `dpkg -L` says it is the package responsible for it)
https://packages.debian.org/buster/amd64/xfce4-panel/filelist
https://packages.debian.org/buster/amd64/xfce4-session/filelist
https://packages.debian.org/buster/amd64/xfconf/filelist
https://packages.debian.org/buster/amd64/xfwm4/filelist
https://packages.debian.org/buster/all/libexo-helpers/filelist


pebenito:
I'd prefer to add these to a `distro_debian` block.

cc: @bigon

bigon:
I would indeed put that in a  `distro_debian` block

The debian policy now officially supports `/usr/libexec/` so it means that in the (long) term some of these files might be moved to `/usr/libexec/xfce4`

pebenito:
Is Debian's `$(libdir)` `/usr/lib/x86_64-linux-gnu`?

bigon:
Both `/usr/lib` and `/usr/lib/x86_64-linux-gnu` are supported for the libraries

The later is the [multiarch](https://wiki.debian.org/Multiarch/Implementation#What_does_the_end_result_look_like.3F) path

pebenito:
I don't have the multiarch path on my Fedora or Gentoo systems, so I think it should go in a distro_debian block.

PR #155, opened 2020-01-08
Initial 5.4 fixes

pebenito:


policy/modules/kernel/files.if
```diff
@@ -4872,6 +4908,24 @@ interface(`files_delete_usr_dirs',`
 	delete_dirs_pattern($1, usr_t, usr_t)
 ')
 
+########################################
+## <summary>
+##	Wwatch generic directories in /usr.
```
bauen1: Typo

policy/modules/system/systemd.if
```diff
@@ -420,6 +420,7 @@ interface(`systemd_use_passwd_agent',`
 		type systemd_passwd_runtime_t;
 	')
 
+	allow $1 systemd_passwd_runtime_t:dir watch;
```
doverride: Just a note for consideration here. Systemd watches all path units by default. The event above is for one of the systemd-ask-password-*.path units i believe:

systemctl cat systemd-ask-password-console.path

You might oonsider handling these on a higher level for self documenting policy. Something like: `systemd_path_unit_type(systemd_passwd_runtime_t)`

This is what i implemented in my policy
 4195               (allow subj daemon_path_obj_type_attribute manage_dir_perms)     |
 4196               (allow subj daemon_path_obj_type_attribute (all_files (watch)))```
```

Comment (doverride) [reply policy/modules/system/systemd.if]: indepedent of the above though:`systemd-tty-ask-passwd--agent` also watches this

Comment (doverride) [reply policy/modules/system/systemd.if]: This is currently the only common path unit though:
```
$ grep -r systemd.system.daemon_path_obj
policy/systemd/a/ask_password.cil:              (call systemd.system.daemon_path_obj_type (runtime_file)))
```
But still its easier to manage and maintain path unit paths this way in the longer term

pebenito: Makes sense, thanks for pointing it out.

PR #156, opened 2020-01-12
usermanage: add groupadd to lookup dynamic users from systemd

fishilico:
On a Debian 10 test virtual machine, when installing packages adds a group, the following AVC occurs:

    type=USER_AVC msg=audit(1578863991.588:575): pid=381 uid=104
    auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t
    msg='avc:  denied  { send_msg } for msgtype=method_call
    interface=org.freedesktop.systemd1.Manager
    member=LookupDynamicUserByName dest=org.freedesktop.systemd1
    spid=13759 tpid=1 scontext=unconfined_u:unconfined_r:groupadd_t
    tcontext=system_u:system_r:init_t tclass=dbus permissive=1
    exe="/usr/bin/dbus-daemon" sauid=104 hostname=? addr=? terminal=?'

Allow `groupadd` to call DBUS method `LookupDynamicUserByName()`.

### DIFF
--- a/policy/modules/admin/usermanage.te
+++ b/policy/modules/admin/usermanage.te
@@ -278,6 +278,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	systemd_use_nss(groupadd_t)
+')
+
+optional_policy(`
 	unconfined_use_fds(groupadd_t)
 ')
 
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -49,6 +49,9 @@ init_system_domain(systemd_binfmt_t, systemd_binfmt_exec_t)
 type systemd_binfmt_unit_t;
 init_unit_file(systemd_binfmt_unit_t)
 
+type systemd_conf_t;
+files_config_file(systemd_conf_t)
+
 type systemd_gpt_generator_t;
 type systemd_gpt_generator_exec_t;
 init_system_domain(systemd_gpt_generator_t, systemd_gpt_generator_exec_t)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -1,5 +1,6 @@
 /etc/\.updated				--	gen_context(system_u:object_r:systemd_update_run_t,s0)
 
+/etc/systemd/dont-synthesize-nobody	--	gen_context(system_u:object_r:systemd_conf_t,s0)
 /etc/udev/hwdb\.bin			--	gen_context(system_u:object_r:systemd_hwdb_t,s0)
 
 /run/log/journal(/.*)?				gen_context(system_u:object_r:systemd_journal_t,s0)
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -81,6 +81,34 @@ interface(`systemd_log_parse_environment',`
 
 ######################################
 ## <summary>
+##   Allow domain to use systemd's Name Service Switch (NSS) module.
+##   This module provides UNIX user and group name resolution for dynamic users
+##   and groups allocated through the DynamicUser= option in systemd unit files
+## </summary>
+## <param name="domain">
+##   <summary>
+##     Domain allowed access
+##   </summary>
+## </param>
+#
+interface(`systemd_use_nss',`
+	gen_require(`
+		type systemd_conf_t;
+	')
+
+	# Get attributes of /etc/systemd/dont-synthesize-nobody
+	files_search_etc($1)
+	allow $1 systemd_conf_t:file getattr;
+
+	optional_policy(`
+		dbus_system_bus_client($1)
+		# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
+		init_dbus_chat($1)
+	')
+')
+
+######################################
+## <summary>
 ##   Allow domain to be used as a systemd service with a unit
 ##   that uses PrivateDevices=yes in section [Service].
 ## </summary>


doverride:
Nicolas Iooss <notifications@github.com> writes:

> On a Debian 10 test virtual machine, when installing packages adds a group, the following AVC occurs:
>
> type=USER_AVC msg=audit(1578863991.588:575): pid=381 uid=104
> auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t
> msg='avc:  denied  { send_msg } for msgtype=method_call
> interface=org.freedesktop.systemd1.Manager
> member=LookupDynamicUserByName dest=org.freedesktop.systemd1
> spid=13759 tpid=1 scontext=unconfined_u:unconfined_r:groupadd_t
> tcontext=system_u:system_r:init_t tclass=dbus permissive=1
> exe="/usr/bin/dbus-daemon" sauid=104 hostname=? addr=? terminal=?'
>
> Allow groupadd to call DBUS method LookupDynamicUserByName().

Probably better to deal with this on glibc-nss level (this is caused by
nss_systemd)

>
> ------------------------------------------------------------------------------------------------------------------------------------------------
>
>
>   https://github.com/SELinuxProject/refpolicy/pull/156
>
> Commit Summary
>
> * usermanage: add groupadd to lookup dynamic users from systemd
>
> File Changes
>
> * M policy/modules/admin/usermanage.te (4) 
>
> Patch Links:
>
> * https://github.com/SELinuxProject/refpolicy/pull/156.patch
> * https://github.com/SELinuxProject/refpolicy/pull/156.diff
>
> —
>

-- 
Dominick Grift

fishilico:
> Probably better to deal with this on glibc-nss level (this is caused by
nss_systemd)

When a similar issue was last raised on the mailing list a few months ago, @pebenito preferred not to extend `auth_use_nsswitch()` nor `auth_use_pam()` in this regard (https://lore.kernel.org/selinux-refpolicy/bb485d61-aa34-5f53-e129-2c9cdb4958cc@ieee.org/).
I share this preference too, because even though the DBUS call is caused by nss-systemd, I only see few domains actually needing a new rule to interact with ``init_t`` over DBUS just for nsswitch.

doverride:
Ok but might be better to create some sort of higher level interface for this or are you going to comment every individual call?

In my policy theres a pretty solid way to split this:
1. systemd generally creates objects with dynamic uids in "private" dirs that unpriv processes cannot access.
2. unpriv processes can not read state of processes associated with the system role (unless a role or type is exempted)
3. the above practically means that users and user session processes dont need access to this as they can read system state and can't access "private".
4. the dynamic uid resolving only happens on demand, so if it cannot other wise resolve a uid then it will resort to dynamic uid resolving.

I have two kind of "auth_use_nsswitch" type interfaces, one "base" and one "full" The base only provides access to nsswitch.conf , passwd and sssd. where possible i use this (confined user shells for example) other nss clients get the whole deal incl. nss_systemd

fishilico:
I agree that having a new interface sounds nice. As I don't know what "full" will contain besides accesses needed by nss-systemd, I suggest creating a new systemd-centered interface named `auth_use_nss_systemd` with an attribute `nss_systemd_type`.
Here is a non-tested set of rules that could be written (as extracted from a related PR, https://github.com/SELinuxProject/refpolicy/pull/98/files#diff-7994cfa189eab5c427add49ed258d5eaR69):
```
ifdef(`init_systemd', `
	# For /run/systemd/dynamic-uid/
	init_list_pids(nss_systemd_type)

	optional_policy(`
		dbus_system_bus_client(nss_systemd_type)
		# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
		init_dbus_chat(nss_systemd_type)
	')
')
```
What about this?

pebenito:
I'm fine with a systemd interface like:
```
interface(`systemd_use_nss',`
	dbus_system_bus_client($1)
	# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
	init_dbus_chat($1)
	# For /run/systemd/dynamic-uid/
	init_list_pids($1)
')
```

doverride:
> I agree that having a new interface sounds nice. As I don't know what "full" will contain besides accesses needed by nss-systemd, I suggest creating a new systemd-centered interface named `auth_use_nss_systemd` with an attribute `nss_systemd_type`.
> Here is a non-tested set of rules that could be written (as extracted from a related PR, https://github.com/SELinuxProject/refpolicy/pull/98/files#diff-7994cfa189eab5c427add49ed258d5eaR69):
> 
> ```
> ifdef(`init_systemd', `
> 	# For /run/systemd/dynamic-uid/
> 	init_list_pids(nss_systemd_type)
> 
> 	optional_policy(`
> 		dbus_system_bus_client(nss_systemd_type)
> 		# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
> 		init_dbus_chat(nss_systemd_type)
> 	')
> ')
> ```
> 
> What about this?

I wouldnt add the "init_list_pids($1)" as that seems to be used only by dbus.
Instead i would allow callers of the interface to get attributes of "/etc/systemd/dont-synthesize-nobody"

https://github.com/systemd/systemd/blob/200427cf7c8d5184bf0762b83a6084cc05e94dd4/src/basic/user-util.c#L825

fishilico:
> I'm fine with a systemd interface like:
> 
> ```
> interface(`systemd_use_nss',`
> 	dbus_system_bus_client($1)
> 	# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
> 	init_dbus_chat($1)
> 	# For /run/systemd/dynamic-uid/
> 	init_list_pids($1)
> ')
> ```

I am encountering some issues while implementing this interface with an `optional_policy` around calls to `dbus` module:
```
interface(`systemd_use_nss',`
	optional_policy(`
		dbus_system_bus_client($1)
		# For GetDynamicUser(), LookupDynamicUserByName()... of org.freedesktop.systemd1.Manager
		init_dbus_chat($1)
	')
')
```
This leads to this not being compilable:
```
ifdef(`init_systemd',`
	optional_policy(`
		systemd_use_nss(groupadd_t)
	')
')
```
... because there is an `optional_policy` block without requirements. This should be fixed if a statement is added in order to allow getting the attributes of `/etc/systemd/dont-synthesize-nobody`. How should this file be labeled? `etc_t`, `systemd_conf_dont_synthesize_nobody_t` or something else?

doverride:
The labeling for systemd that i have is pretty complicated

```
drwxr-xr-x.   6 root root sys.id:sys.role:systemd.generic_config.config_file:s0      4096 Dec 22 10:35 .
drwxr-xr-x. 124 root root sys.id:sys.role:files.generic_config.config_file:s0       12288 Jan 11 12:04 ..
-rw-r--r--.   1 root root sys.id:sys.role:bootchart.config.config_file:s0             720 May 31  2016 bootchart.conf
-rw-r--r--.   1 root root sys.id:sys.role:sd_coredump.config.config_file:s0           615 Dec 15 11:47 coredump.conf
-rw-r--r--.   1 root root sys.id:sys.role:importd.cert.cert_file:s0                  4091 Aug  2 08:25 import-pubring.gpg
-rw-r--r--.   1 root root sys.id:sys.role:journald.config.config_file:s0             1038 Oct 19 18:47 journald.conf
-rw-r--r--.   1 root root sys.id:sys.role:logind.config.config_file:s0               1042 Dec 22 10:32 logind.conf
drwxr-xr-x.   2 root root sys.id:sys.role:networkd.config.config_file:s0             4096 Dec 22 10:35 network
-rw-r--r--.   1 root root sys.id:sys.role:networkd.config.config_file:s0              584 Dec 15 11:47 networkd.conf
drwxr-xr-x.   2 root root sys.id:sys.role:nspawn.config.config_file:s0               4096 Aug 11 13:19 nspawn
-rw-r--r--.   1 root root sys.id:sys.role:sd_pstore.config.config_file:s0             529 Dec 15 11:47 pstore.conf
-rw-r--r--.   1 root root sys.id:sys.role:resolved.config.config_file:s0              764 Dec 22 10:32 resolved.conf
-rw-r--r--.   1 root root sys.id:sys.role:sd_sleep.config.config_file:s0              790 Dec 15 11:47 sleep.conf
drwxr-xr-x.  20 root root sys.id:sys.role:systemd.system.generic_units.unit_file:s0  4096 Dec 22 10:35 system
-rw-r--r--.   1 root root sys.id:sys.role:systemd.generic_config.config_file:s0      1762 Dec 22 10:32 system.conf
-rw-r--r--.   1 root root sys.id:sys.role:timesyncd.config.config_file:s0             677 Dec 22 10:32 timesyncd.conf
drwxr-xr-x.   4 root root sys.id:sys.role:systemd.user.generic_units.unit_file:s0    4096 Dec 22 10:35 user
-rw-r--r--.   1 root root sys.id:sys.role:systemd.generic_config.config_file:s0      1185 Dec 22 10:32 user.conf
```
If you compare it to /etc (etc_t) then /etc/systemd is systemd_conf_t (generic systemd config) so is /etc/systemd/dont_synthesize_nobody, but most other stuff in there has private labels (for confined administration)

doverride:
types maintained under systemd.generic_config.config_file in dssp2:
```
[root@brutus ~]# seinfo -xasystemd.config_file.obj_type_attribute

Type Attributes: 1
   attribute systemd.config_file.obj_type_attribute;
        bootchart.config.config_file
        journal_remote.config.config_file
        journal_upload.config.config_file
        journald.config.config_file
        logind.config.config_file
        networkd.config.config_file
        nspawn.config.config_file
        portable.config.config_file
        resolved.config.config_file
        sd_coredump.config.config_file
        sd_pstore.config.config_file
        sd_sleep.config.config_file
        systemd.generic_config.config_file
        systemd.system.control_config.config_file
        timedated.config.config_file
        timesyncd.config.config_file
[root@brutus ~]# 
```

fishilico:
I updated this PR introduce interface `systemd_use_nss` and use it.

policy/modules/system/systemd.te
```diff
@@ -22,6 +22,7 @@ gen_tunable(systemd_nspawn_labeled_namespace, false)
 
 attribute systemd_log_parse_env_type;
 attribute systemd_tmpfiles_conf_type;
+attribute systemd_use_nss_type;
```
doverride: redundant?

policy/modules/system/systemd.te
```diff
@@ -49,6 +50,9 @@ init_system_domain(systemd_binfmt_t, systemd_binfmt_exec_t)
 type systemd_binfmt_unit_t;
 init_unit_file(systemd_binfmt_unit_t)
 
+type systemd_conf_t;
+files_type(systemd_conf_t)
```
doverride: I suppose its a `files_config_file()`

fishilico:
Thanks for your review! `systemd_use_nss_type` slipped through my proof-reading and using `files_type` came for looking at how other modules defined their `..._conf_t` types: `grep 'files_type.*_conf_t'` yields some results:
```
policy/modules/admin/portage.te:63:files_type(portage_conf_t)
policy/modules/system/modutils.te:20:files_type(modules_conf_t)
policy/modules/system/sysnetwork.te:39:files_type(net_conf_t)
policy/modules/services/squid.te:40:files_type(squid_conf_t)
policy/modules/services/bind.te:37:files_type(named_conf_t)
policy/modules/services/nsd.te:16:files_type(nsd_conf_t)
policy/modules/services/kerberos.te:38:files_type(krb5_conf_t)
policy/modules/services/kerberos.te:50:files_type(krb5kdc_conf_t)
policy/modules/services/zebra.te:21:files_type(zebra_conf_t)
policy/modules/services/hostapd.te:13:files_type(hostapd_conf_t)
policy/modules/services/gpm.te:16:files_type(gpm_conf_t)
policy/modules/services/glusterfs.te:13:files_type(glusterd_conf_t)
policy/modules/services/nis.te:49:files_type(ypserv_conf_t)
```

Some of these matches could also be converted to `files_config_file`.

Anyway, I updated my commits.

policy/modules/admin/usermanage.te
```diff
@@ -277,6 +277,12 @@ optional_policy(`
 	rpm_rw_pipes(groupadd_t)
 ')
 
+ifdef(`init_systemd',`
```
pebenito: I don't think this needs to be in a init_systemd block; the optional should suffice.

fishilico: Ok. I removed the `ifdef(init_systemd)` block.

PR #159, opened 2020-01-14
Remove unused permissions

stephensmalley:
Remove unused permissions from SELinux.  These have all been long-since removed from the kernel or were never used by it in the first place.


stephensmalley:
This fails travis-ci because it introduces refpolicywarn() statements in interfaces that only allowed the removed/unused permissions to indicate that the entire interface is now deprecated and the interface is still called by various modules.  Not sure how you want to handle.  Could remove all callers but leave the interface with deprecation warning, remove the interface and all callers, ...

stephensmalley:
Updated to make sure all obsolete interfaces generate warnings, use dollarszero as appropriate, and remove all callers in upstream refpolicy.

stephensmalley:
@wrabcak @WOnder93 Would be good to apply this to Fedora policy too at some point.  Shouldn't have any real compatibility issues, except possibly requiring rebuilding third party modules and audit2allow-generated modules that may have picked up the dead permissions in their require blocks as a result of using the refpolicy policy_module() interface that generates requires on all the kernel classes/perms defined.

PR #160, opened 2020-01-14
init: Revise systemd bind mounts.

pebenito:
Add interface similar to files_mountpoint() and add a conditional which
allows mount on non_security_file_type.

Signed-off-by: Chris PeBenito <chpebeni@linux.microsoft.com>


doverride:
This will not address making device nodes InaccessiblePath=

systemd-run -p InaccessiblePaths=/dev/kvm true

PR #161, opened 2020-01-15
Rename obsolete netlink_firewall_socket and netlink_ip6fw_socket classes

stephensmalley:
The implementation for NETLINK_FIREWALL and NETLINK_IP6_FW protocols
was removed from the kernel in commit
d16cf20e2f2f13411eece7f7fb72c17d141c4a84 ("netfilter: remove ip_queue
support") circa Linux 3.5.  Consequently, kernels >= 3.5 should never
perform permission checks on these classes although they remained
defined in the SELinux kernel classmap until the netlink classes
were updated by
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6c6d2e9bde1c1c87a7ead806f8f5e2181d41a652
circa Linux v4.2.

Removing these class definitions would break legacy userspace that relies
upon stable values for the userspace security class definitions since it
will perturb those values by removing classes that preceded them.  dbus-daemon
in particular is known to break if its dbus class changes at runtime,
which could occur upon a policy reload that removes these classes.
Fixing this requires ensuring that dbus-daemon looks up the appropriate
class value on each use or upon policy reload, via userspace interfaces
such as selinux_check_access(), string_to_security_class(), and/or
selinux_set_callback(SELINUX_CB_POLICYLOAD, ...) with a callback function
that remaps the class value if needed.  Other userspace policy enforcers
are believed to have been updated in recent versions but older versions
may break upon such a change.

Hence, this change renames these classes with obsolete_ prefixes and
removes all rules referencing them from refpolicy, thereby preserving
the class numbering for subsequent classes while making it clear that
these classses are no longer meaningful for modern kernels.

This change does however create a potential compatibility break for
kernels < 3.5, since the policy will cease to define the kernel class
names and therefore the kernel will handle permission checks on the
class based on the handle_unknown setting in policy.  For most
Linux distributions, this will default to allow and therefore avoid
breaking userspace but will fail open.  For kernels < 2.6.33 (i.e.
the dynamic class/perm discovery support), the presence of a class
in policy with the same number but a different name than the kernel
class will cause the policy load to fail entirely.

Signed-off-by: Stephen Smalley <sds@tycho.nsa.gov>


stephensmalley:
Caveat: This change almost certainly breaks RHEL6 due to both kernel and userspace compatibility and may break RHEL7 userspace (if there are any lingering users of the fixed class/perm definitions in RHEL7 userspace).

stephensmalley:
An unscientific look at RHEL7 suggests that everything has been converted except that dbus-daemon doesn't yet handle a policy reload at runtime that changes its class definition, so a policy update on RHEL7 that drop these classes could cause dbus-daemon to start always failing checks.

stephensmalley:
@WOnder93 FYI wrt RHEL7 comments above

stephensmalley:
This might also be true on RHEL8. dbus in Fedora was fixed not by patching dbus-daemon but by switching to dbus-broker.   Upstream, there is a patch in dbus git that ostensibly fixes the problem but no released version yet. https://gitlab.freedesktop.org/dbus/dbus/issues/198

doverride:
> This might also be true on RHEL8. dbus in Fedora was fixed not by patching dbus-daemon but by switching to dbus-broker. Upstream, there is a patch in dbus git that ostensibly fixes the problem but no released version yet. https://gitlab.freedesktop.org/dbus/dbus/issues/198

I think that dbus patch only partially fixes it. Last time i tried it did not work for me (ie messing with classorder at still confuses dbus and makes it block all.

stephensmalley:
> I think that dbus patch only partially fixes it. Last time i tried it did not work for me (ie messing with classorder at still confuses dbus and makes it block all.

Is that with dbus-daemon, dbus-broker or both?

In any event, since this seems like it could be a breaking change for RHEL8 and RHEL7, maybe we want to go with the alternative I proposed in the commit message of just renaming the classes and removing rules referencing them?  That would keep SECCLASS_DBUS and other userspace classes with the same values but avoid user confusion over the obsolete classes.  That presumes we are ok dropping kernel compatibility for kernels < 3.5, including RHEL6.

doverride:
> > I think that dbus patch only partially fixes it. Last time i tried it did not work for me (ie messing with classorder at still confuses dbus and makes it block all.
> 
> Is that with dbus-daemon, dbus-broker or both?
> 
> In any event, since this seems like it could be a breaking change for RHEL8 and RHEL7, maybe we want to go with the alternative I proposed in the commit message of just renaming the classes and removing rules referencing them? That would keep SECCLASS_DBUS and other userspace classes with the same values but avoid user confusion over the obsolete classes. That presumes we are ok dropping kernel compatibility for kernels < 3.5, including RHEL6.

dbus-broker is good. It's just dbus.

stephensmalley:
> > I think that dbus patch only partially fixes it. Last time i tried it did not work for me (ie messing with classorder at still confuses dbus and makes it block all.
> 
> Is that with dbus-daemon, dbus-broker or both?
> 
> In any event, since this seems like it could be a breaking change for RHEL8 and RHEL7, maybe we want to go with the alternative I proposed in the commit message of just renaming the classes and removing rules referencing them? That would keep SECCLASS_DBUS and other userspace classes with the same values but avoid user confusion over the obsolete classes. That presumes we are ok dropping kernel compatibility for kernels < 3.5, including RHEL6.

Or we could just defer this to the distant future when dbus-daemon and RHEL6 are fully dead and gone.

pebenito:
I think the renaming is probably the best choice. I don't have a problem breaking kernel < 3.5 and RHEL6, but I think other distros are using dbus-daemon. @perfinion @bigon?

WOnder93:
@stephensmalley I don't think you need to worry about refpolicy changes vs. RHEL. AFAIK, we don't ship refpolicy itself as a supported option. RHEL policy is based on the Fedora fork of refpolicy, which is not being actively synced/rebased with upstream refpolicy. (And even if it wasn't this way, we could just simply not backport this commit into RHEL and I don't think it would cause any big issues with later backports.)

@bachradsusi, @wrabcak, am I right?

wrabcak:
> 
> @bachradsusi, @wrabcak, am I right?

Yes,  you're right.

stephensmalley:
Updated to rename the two classes with obsolete_ prefixes instead of removing them, while still removing all rules referencing them from refpolicy.  Preserves userspace class values for compatibility with dbus-daemon and others. Potential compatibility break for kernels < 3.5 (depending on handle_unknown), definite compatibility break for kernels < 2.6.33 (those without dynamic class/perm support).  Kernels < 4.2 might also produce class not defined warnings on loading policies with this change because they still had the classes in the kernel classmap even though they were no longer being used by anything.

perfinion:
Gentoo still uses dbus-daemon yeah. dbus-broker is packaged on gentoo but its not the default. I didnt even realize it existed till just now.

pebenito:
@WOnder93 @wrabcak @bachradsusi While refpolicy is not officially supported by on RHEL, there are systems that use it. Because of that, I keep RHEL compatibility as much as reasonable.

bigon:
I just see this now, but indeed this should be with https://gitlab.freedesktop.org/dbus/dbus/commit/a442601cb2e14f6ff3111fe5a86ebdf4d0dee436 and https://gitlab.freedesktop.org/dbus/dbus/commit/6072f8b24153d844a3033108a17bcd0c1a967816, but note there is no (stable) release with these.

The remaining issue was that libselinux is not allowing to flush the class cache, see @stephensmalley mail from a while ago, https://marc.info/?l=selinux&m=152181091619149&w=2. I don't think that a way to do this has been added yet?

stephensmalley:
> I just see this now, but indeed this should be with https://gitlab.freedesktop.org/dbus/dbus/commit/a442601cb2e14f6ff3111fe5a86ebdf4d0dee436 and https://gitlab.freedesktop.org/dbus/dbus/commit/6072f8b24153d844a3033108a17bcd0c1a967816, but note there is no (stable) release with these.
> 
> The remaining issue was that libselinux is not allowing to flush the class cache, see @stephensmalley mails from a while ago, https://marc.info/?l=selinux&m=152181091619149&w=2 and https://marc.info/?l=selinux&m=152181125219220&w=2. I don't think that a way to do this has been added yet?

Thanks for the reminder.  Just posted a patch for libselinux to automatically flush the class cache upon receiving a policyload notification, so dbus-daemon should pick that up automatically once libselinux is updated.  Also exported the interface (renamed to have a selinux_ prefix) so that other applications can explicitly call it if needed but that shouldn't be necessary for dbus-daemon AFAICT. BTW still not sure why dbus-daemon couldn't just use selinux_check_access() but maybe that isn't sufficiently performant for its usage.  Looks like dbus-daemon has an avc entry reference but it isn't really using it in a meaningful way so that could just be dropped.

PR #162, opened 2020-01-16
Add requires to interfaces that reference types or attributes without…

dburgener:
… requiring them

### DIFF
--- a/policy/modules/services/procmail.if
+++ b/policy/modules/services/procmail.if
@@ -90,7 +90,7 @@ interface(`procmail_read_home_files',`
 #
 interface(`procmail_relabel_home_files',`
 	gen_require(`
-		type ppp_home_t;
+		type ppp_home_t, procmail_home_t;
 	')
 
 	userdom_search_user_home_dirs($1)
--- a/policy/modules/services/procmail.if
+++ b/policy/modules/services/procmail.if
@@ -90,7 +90,7 @@ interface(`procmail_read_home_files',`
 #
 interface(`procmail_relabel_home_files',`
 	gen_require(`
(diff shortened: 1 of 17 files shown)


dburgener:
I've updated to add the Signed-off-by.  Thanks!

policy/modules/services/procmail.if
```diff
@@ -90,7 +90,7 @@ interface(`procmail_read_home_files',`
 #
 interface(`procmail_relabel_home_files',`
 	gen_require(`
-		type ppp_home_t;
+		type ppp_home_t, procmail_home_t;
```
pebenito: In this case it should drop `ppp_home_t` as unused.

policy/modules/services/xserver.if
```diff
@@ -22,6 +22,7 @@ interface(`xserver_restricted_role',`
 		type user_fonts_t, user_fonts_cache_t, user_fonts_config_t;
 		type iceauth_t, iceauth_exec_t, iceauth_home_t;
 		type xauth_t, xauth_exec_t, xauth_home_t;
+		type xdm_t, xdm_tmp_t;
```
pebenito: `xserver_exec_t` appears to be unused.

policy/modules/services/zabbix.if
```diff
@@ -109,7 +109,7 @@ interface(`zabbix_read_pid_files',`
 #
 interface(`zabbix_agent_tcp_connect',`
 	gen_require(`
-		type zabbix_agent_t;
+		type zabbix_t, zabbix_agent_t;
```
pebenito: Seems that `zappix_agent_t` should be removed.

PR #166, opened 2020-01-21
Fix situations where require blocks in interfaces listed types not ac…

dburgener:
…tually referenced by that interface

Signed-off-by: Daniel Burgener <dburgener@tresys.com>

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -2312,7 +2312,7 @@ interface(`userdom_manage_user_home_content_files',`
 #
 interface(`userdom_dontaudit_manage_user_home_content_dirs',`
 	gen_require(`
-		type user_home_dir_t, user_home_t;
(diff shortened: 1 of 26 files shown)


policy/modules/system/userdomain.if
```diff
@@ -2312,10 +2312,10 @@ interface(`userdom_manage_user_home_content_files',`
 #
 interface(`userdom_dontaudit_manage_user_home_content_dirs',`
 	gen_require(`
-		type user_home_dir_t, user_home_t;
+		type user_home_t, user_home_dir_t;
 	')
 
-	dontaudit $1 user_home_t:dir manage_dir_perms;
+	dontaudit $1 { user_home_t user_home_dir_t }:dir manage_dir_perms;
```
pebenito: Actually the interface already had the intended effect, so the `user_home_dir_t` should be dropped from the require.

PR #167, opened 2020-01-22
Allow user domains to read/write the wireless devices (e.g. rfkill)

gtrentalancia:
Allow userdomain to read and write the wireless devices (for
example for querying their state, enabling and/or disabling
them using userspace tools such as "rfkill" from util-linux):
this is tunable policy.

See also:
https://wireless.wiki.kernel.org/en/users/documentation/rfkill

Make the same access tunable throughout the rest of the policy
as it is generally controlled directly and interactively by
the user.

--
 policy/modules/admin/blueman.te           |    5 ++++-
 policy/modules/apps/wm.te                 |    5 ++++-
 policy/modules/kernel/devices.te          |    9 +++++++++
 policy/modules/services/bluetooth.te      |    5 ++++-
 policy/modules/services/hostapd.te        |    5 ++++-
 policy/modules/services/networkmanager.te |    5 ++++-
 policy/modules/system/systemd.te          |    5 ++++-
 policy/modules/system/userdomain.if       |    4 ++++
 8 files changed, 37 insertions(+), 6 deletions(-)

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -605,6 +605,7 @@ template(`userdom_common_user_template',`
 	dev_read_sound($1_t)
 	dev_read_sound_mixer($1_t)
 	dev_write_sound_mixer($1_t)
+	dev_rw_wireless($1_t)
 
 	files_exec_etc_files($1_t)
 	files_search_locks($1_t)


policy/modules/system/userdomain.if
```diff
@@ -605,6 +605,7 @@ template(`userdom_common_user_template',`
 	dev_read_sound($1_t)
 	dev_read_sound_mixer($1_t)
 	dev_write_sound_mixer($1_t)
+	dev_rw_wireless($1_t)
```
pebenito: I'm on the fence. We may want to make a tunable for this.

gtrentalancia: In view of the fact that wireless networks expose to more risks compared to a wired interface coming from a socket in a secured building, it does make sense to make it tunable...

gtrentalancia: I am thinking of calling it "user_manage_rf_devices", but "user_manage_wireless_devices" is also good to me...

gtrentalancia: Actually: "allow_user_manage_rf_devices", defaulting to "true" since the risk is probably medium rather than high, thus giving precedence to usability.

gtrentalancia:
However, if this is made tunable for user domain, then it would need to be made tunable also for other domains that are run/controlled interactively by the user!

At the moment the following are dealing with the dev_rw_wireless() interface:

apps/wm.te
system/systemd.te
admin/blueman.te
services/networkmanager.te
services/bluetooth.te
services/hostapd.te

so I suppose at least the following ones also need to have the new tunable check:

apps/wm.te

as they are directly and interactively controlled by the user.

However, the following other domains might also need to have the new tunable check in virtue of the fact that they are controlled by the user by the means of getting started as init services which implies init_daemon_domain() or getting started from other interactive userspace tools (such as ifconfig and so on):

system/systemd.te
services/networkmanager.te
services/bluetooth.te
services/hostapd.te

So, the issue of making it tunable is possibly a bit more complicated than just adding the tunable here.

policy/modules/system/userdomain.if
pebenito: I wasn't specifically requesting the change, I was undecided. With what you pointed out, the original change was ok.

gtrentalancia:
A better option might probably be to only allow both read/write permissions for administrative users, while giving read-only permissions to common users (by defining a new device.if interface).

The risk is mostly related to Denial Of Service, i.e. an unprivileged process disabling the wireless network.

What do you say ?

PR #168, opened 2020-01-23
Escape unescaped . in file context regular expression.

dburgener:
Signed-off-by: Daniel Burgener <dburgener@tresys.com>


dburgener:
Looks like this is a false positive.

PR #169, opened 2020-01-23
Update the pulseaudio application module

gtrentalancia:
Update the pulseaudio application module with a few user domain file read and management permissions.
---
 policy/modules/apps/pulseaudio.te |    1 +
 1 file changed, 1 insertions(+)


gtrentalancia:
In the original version (now amended), here was a misplaced interface userdom_read_user_data() which is not in refpolicy.

But apart from that, userdom_manage_user_tmp_files() is common in other application modules such as mozilla, evolution, mplayer (and they are not daemons), even when they have their own tmp data type.

PR #170, opened 2020-01-25
Remove obsolete gentoo specific rule

Jarel1337:
Looking at all audit versions in gentoo (2.8.5 to 2.6.4) every single one of them has `var/log/audit` as a directory and not as a file.

Tested on gentoo.

### DIFF
--- a/policy/modules/system/logging.fc
+++ b/policy/modules/system/logging.fc
@@ -63,10 +63,6 @@ ifdef(`distro_suse', `
 /var/log/audit(/.*)?		gen_context(system_u:object_r:auditd_log_t,mls_systemhigh)
 /var/log/syslog-ng(/.*)? 	gen_context(system_u:object_r:syslogd_runtime_t,mls_systemhigh)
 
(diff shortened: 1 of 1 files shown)


fishilico:
Please be aware that you are removing a `ifndef`, not an `ifdef`: the pattern you are removing was being applied on every distribution but Gentoo.

That being said, it seems that `/var/log/audit.log` is the historical path to the audit log file and its context has been introduced in 2005 with commit eb7f9a34cbec ("move audit to logging"). Two months later, commit a28f6db57649 ("add in some rules from NSA CVS to make targeted policy work") added a label for `/var/log/audit(/.*)?`. In 2006, commit 14b1684aae4f ("gentoo testing fixes.") introduced the `ifndef(distro_gentoo)` block. I guess that it was causing something to break on Gentoo?

Anyway, I agree with removing the file context rule for `/var/log/audit.log`. If a distribution is still using it, it can be added again within a proper `ifdef` block.

PR #172, opened 2020-01-26
sysadm: allow usage of inherited pipes from role changing domains

bauen1:
Allows `staff_r:staff_t`, `auditadm_r:auditadm_t`, secadm_r:secadm_t` access to inherited pipes (sudo with pipes).
e.g:
```bash
# id -Z
staff_u:staff_r:staff_t:s0
# sudo echo 1234 | cat -
1234
```

### DIFF
--- a/policy/modules/roles/sysadm.if
+++ b/policy/modules/roles/sysadm.if
@@ -9,14 +9,24 @@
 ##	Role allowed access.
 ##	</summary>
 ## </param>
+## <param name="domain">
+##  <summary>
+##  Domain allowed access.
+##  </summary>
+## </param>
 ## <rolecap/>
 #
 interface(`sysadm_role_change',`
 	gen_require(`
+		type sysadm_t;
 		role sysadm_r;
 	')
 
 	allow $1 sysadm_r;
+
+	tunable_policy(`sysadm_allow_rw_inherited_fifo', `
+		allow sysadm_t $2:fifo_file rw_inherited_fifo_file_perms;
+	')
 ')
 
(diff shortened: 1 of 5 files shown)


policy/modules/roles/sysadm.if
```diff
@@ -236,3 +236,22 @@ interface(`sysadm_rw_pipes',`
...
+##  </summary>
+## </param>
+#
+interface(`sysadm_rw_inherited_pipe_from', `
+	gen_require(`
+		type sysadm_t;
+	')
+
+	allow sysadm_t $1:fifo_file rw_inherited_fifo_file_perms;
+')
```
pebenito: I'm not completely convinced that we want to have this in general. If we do, it would be better to add this onto the `sysadm_role_change()` interface by adding the source domain as a parameter. Maybe the the fifo_file rule should be in a conditional block (boolean) in the interface too.

policy/modules/roles/sysadm.te
```diff
@@ -12,6 +12,14 @@ policy_module(sysadm, 2.15.3)
 ## </desc>
 gen_tunable(allow_ptrace, false)
 
+## <desc>
+## <p>
+## Allow sysadm to read/write to fifo files inherited from
+## a domain allowed to change role.
+## </p>
+## </desc>
+gen_tunable(sysadm_allow_rw_inherited_fifo, true)
```
pebenito: Since this allows the sysadm to possibly be influenced by unpriv processes, this should default to `false`.

PR #178, opened 2020-02-09
Restrict down wireless device permissions to prevent DoS attack

gtrentalancia:
This patch improves a previous commit by restricting down
the permissions to write the wireless device in order to
prevent a possible Denial of Service (DoS) attack from an
unprivileged process bringing down the wireless interfaces.

Only administrative users can now enable/disable the wireless
interfaces, while normal users can only read their status.

---
 policy/modules/kernel/devices.if    |   18 ++++++++++++++++++
 policy/modules/system/userdomain.if |    3 ++-
 2 files changed, 20 insertions(+), 1 deletion(-)

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -5065,6 +5065,24 @@ interface(`dev_write_watchdog',`
 
 ########################################
 ## <summary>
+##	Read the wireless device.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dev_read_wireless',`
+	gen_require(`
+		type device_t, wireless_device_t;
+	')
+
+	read_chr_files_pattern($1, device_t, wireless_device_t)
+')
+
+########################################
+## <summary>
 ##	Read and write the the wireless device.
 ## </summary>
 ## <param name="domain">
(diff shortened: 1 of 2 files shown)


gtrentalancia:
See also:

https://github.com/SELinuxProject/refpolicy/pull/167#issuecomment-578438194

pebenito:
I think your point was good in the other PR, where you said that other domains that a user can control can control the wireless too. It doesn't seem like this provides much value as-is because of that.

gtrentalancia:
user_t controlling the so-called "wireless button" (rfkill) as in current refpolicy and as recently introduced by https://github.com/SELinuxProject/refpolicy/pull/167 exposes the network to a potential Denial of Service (DoS) attack by a malicious process running as user_t (quite easily achieved) bringing down the wireless network connectivity.

The physical user can still control the wireless interface through the other domains such as the window manager (wm_t) domain, for example by clicking on a wireless network button eventually provided by the window manager, however the *key point* here is that this latter scenario is very unlikely to cause DoS because user_t would not normally be able to label, relabel or install (malicious) binaries as wm_exec_t !

Another way of seeing this is that user_t is the default domain for potentially malicious scripts or binaries that the physical user might for example create deliberately or otherwise unintentionally download from the network: in such scenario the malicious script or binary (or the malicious user) would be able to bring down the wireless network, thus causing Denial of Service.

Given the above, I would strongly recommend the additional patch in this pull request !

pebenito:
The concern is not about relabeling wm_exec_t, it is the fact that user_t processes have a lot of direct access to the processes (IPC, parent process, configuration files, etc.) which have the rfkill access.

gtrentalancia:
As already explained the current situation is not optimal from a security point of view as it potentially allows Denial of Service (DoS), so this patch is highly recommended.

gtrentalancia:
Wise. Also reflects default 644 permissions on the rfkill device.

PR #179, opened 2020-02-10
Consider iwd equivalent to NetworkManager etc.

topimiettinen:


topimiettinen:
It's not so simple, needs more thought...

PR #182, opened 2020-02-12
Consider iwd equivalent to NetworkManager etc.

topimiettinen:


topimiettinen:
Perhaps `/dev/rfkill` should have a dedicated label instead of `event_device_t`, though that could affect its other users.

topimiettinen:
`/dev/rfkill` as `event_device_t` was indeed a bug in my setup, it's now `wireless_device_t`.

I'll drop this part.

topimiettinen:
Pushed a fixed version. Also, iwd reads a config file `/etc/iwd/main.conf` if present but it's usually not needed.

PR #186, opened 2020-02-15
Remove old exception

Jarel1337:
This exception goes back 14 years to commit 85c20af3c10867e3d62afa4326ea3d0f5dec4e41 and 11a0508edeaa33538abdb74bb5ba033742a77a07.
The tts exception is covered by a distro agnostic rule further up, and the udev rule doesn't even work (it's supposed to be /lib/udev/ not /usr/lib/udev on gentoo) so I seriously doubt anyone is going to miss them.


pebenito:
@perfinion do you have an opinion?

pebenito:
Please add signed-off-by.

PR #187, opened 2020-02-15
Portage update

Jarel1337:
Portage changed the default location for it's repos a while ago: https://github.com/gentoo/portage/commit/ef8469befbd7b61f1bae8f21ffd95e0c0306a4f8


pebenito:
cc @perfinion

policy/modules/admin/portage.fc
```diff
@@ -8,6 +8,7 @@
 /usr/bin/glsa-check	--	gen_context(system_u:object_r:portage_exec_t,s0)
 /usr/bin/layman	--	gen_context(system_u:object_r:portage_fetch_exec_t,s0)
 /usr/bin/sandbox	--	gen_context(system_u:object_r:portage_exec_t,s0)
+/usr/bin/emerge --	gen_context(system_u:object_r:portage_exec_t,s0)
```
perfinion: hmm this already exists in the gentoo policy. Its actually a bit useless, in gentoo there is also `/usr/lib/python-exec/python[0-9]\.[0-9]*/emerge --      gen_context(system_u:object_r:portage_exec_t,s0)`

But keep this for now then i'll handle the merge conflict when I pull these into the gentoo policy and then i'll upstream the differences after.

policy/modules/admin/portage.fc
```diff
@@ -24,6 +25,12 @@
 /usr/portage/distfiles/svn-src(/.*)?	gen_context(system_u:object_r:portage_srcrepo_t,s0)
 
 /var/db/pkg(/.*)?	gen_context(system_u:object_r:portage_db_t,s0)
+/var/db/repos(/.*)?	gen_context(system_u:object_r:portage_ebuild_t,s0)
+/var/cache/binpkgs(/.*)?  gen_context(system_u:object_r:portage_ebuild_t,s0)
+/var/cache/distfiles/cvs-src(/.*)?	gen_context(system_u:object_r:portage_srcrepo_t,s0)
+/var/cache/distfiles/egit-src(/.*)?	gen_context(system_u:object_r:portage_srcrepo_t,s0)
+/var/cache/distfiles/git-src(/.*)?	gen_context(system_u:object_r:portage_srcrepo_t,s0)
```
perfinion: This should be `git[0-9]-src` like in:

https://gitweb.gentoo.org/proj/hardened-refpolicy.git/tree/policy/modules/admin/portage.fc#n26

pebenito:
My preference (but not rule) is that changes like this and #186 should go through distro policies rather than direct upstream. I can go either way.

policy/modules/admin/portage.fc
Jarel1337: Haha alright but if you want this kept I'll keep it in. I'm still very new to policy development.

pebenito:
Please squash commits and add signed-off-by.

PR #188, opened 2020-02-19
add interface to start/stop syslog units, allow journald to remove /run/log/journal

bluca:


policy/modules/system/logging.if
```diff
@@ -427,6 +427,26 @@ interface(`logging_domtrans_syslog',`
...
+## </param>
+#
+interface(`logging_startstop_syslog',`
+	gen_require(`
+		type syslogd_unit_t;
+		class service start;
+		class service stop;
+	')
+	allow $1 syslogd_unit_t:service start;
+	allow $1 syslogd_unit_t:service stop;
```
pebenito: nit: the permissions should go on the same line, e.g. `{ start stop }`, both on the class line and in the rule.

bluca: Fixed

PR #191, opened 2020-02-21
Add usbguard

topimiettinen:
Usbguard enforces the USB device authorization policy for all USB
devices. Users can be authorized to manage rules and make device
authorization decisions using a command line tool.

Add rules for usbguard. Optionally, allow authorized users to control
the daemon, which requires usbguard-daemon to be able modify its rules
in /etc/usbguard.

### DIFF
--- a/policy/modules/admin/usbguard.fc
+++ b/policy/modules/admin/usbguard.fc
@@ -1,0 +1,7 @@
+/etc/usbguard					-d	gen_context(system_u:object_r:usbguard_conf_t,s0)
+/etc/usbguard/rules\.conf				gen_context(system_u:object_r:usbguard_rules_t,s0)
+/etc/usbguard/.+					gen_context(system_u:object_r:usbguard_conf_t,s0)
+
+/usr/sbin/usbguard-daemon			--	gen_context(system_u:object_r:usbguard_daemon_exec_t,s0)
+
+/var/log/usbguard(/.*)?					gen_context(system_u:object_r:usbguard_log_t,s0)
--- a/policy/modules/admin/usbguard.te
+++ b/policy/modules/admin/usbguard.te
@@ -1,0 +1,62 @@
+policy_module(usbguard, 1.0)
+
+########################################
+#
+# Declarations
+#
+
+## <desc>
+##	<p>
+##	Determine whether authorized users can control the daemon,
+##	which requires usbguard-daemon to be able modify its rules in
+##	/etc/usbguard.
+##	</p>
+## </desc>
+gen_tunable(usbguard_user_modify_rule_files, false)
+
+type usbguard_t;
+type usbguard_daemon_exec_t;
+init_daemon_domain(usbguard_t, usbguard_daemon_exec_t)
+
+type usbguard_conf_t;
+files_config_file(usbguard_conf_t)
+
+type usbguard_log_t;
+logging_log_file(usbguard_log_t)
+
+type usbguard_rules_t;
+files_config_file(usbguard_rules_t)
+
+# /dev/shm
+type usbguard_tmpfs_t;
+files_tmpfs_file(usbguard_tmpfs_t)
+
+########################################
+#
+# Usbguard local policy
+#
+
+allow usbguard_t self:capability { chown dac_read_search fowner };
+allow usbguard_t self:netlink_kobject_uevent_socket create_socket_perms;
+allow usbguard_t self:unix_stream_socket rw_stream_socket_perms;
+
+files_read_etc_files(usbguard_t)
+list_dirs_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+read_files_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+read_files_pattern(usbguard_t, usbguard_conf_t, usbguard_rules_t)
+
+manage_dirs_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+manage_files_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+mmap_read_files_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+fs_tmpfs_filetrans(usbguard_t, usbguard_tmpfs_t, { dir file })
+
+append_files_pattern(usbguard_t, usbguard_log_t, usbguard_log_t)
+create_files_pattern(usbguard_t, usbguard_log_t, usbguard_log_t)
+logging_log_filetrans(usbguard_t, usbguard_log_t, file)
+setattr_files_pattern(usbguard_t, usbguard_log_t, usbguard_log_t)
+
+dev_rw_sysfs(usbguard_t)
+
+tunable_policy(`usbguard_user_modify_rule_files',`
+	manage_files_pattern(usbguard_t, usbguard_conf_t, usbguard_rules_t)
+')
--- a/policy/modules/admin/usbguard.if
+++ b/policy/modules/admin/usbguard.if
@@ -1,0 +1,23 @@
+## <summary>
+## Usbguard enforces the USB device authorization policy for all USB
+## devices.
+## </summary>
+
+#####################################
+## <summary>
+##	Connect to usbguard with a unix domain stream socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`usbguard_stream_connect',`
+	gen_require(`
+		type usbguard_t, usbguard_tmpfs_t;
+	')
+
+	files_search_pids($1)
+	stream_connect_pattern($1, usbguard_tmpfs_t, usbguard_tmpfs_t, usbguard_t)
+')
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -1209,6 +1209,11 @@ template(`userdom_unpriv_user_template', `
 	optional_policy(`
 		systemd_dbus_chat_logind($1_t)
 	')
+
+	# Allow controlling usbguard
+	tunable_policy(`usbguard_user_modify_rule_files',`
+		usbguard_stream_connect($1_t)
+	')
 ')
 
 #######################################


topimiettinen:
The users authorized by `IPCAllowedUsers`, `IPCAllowedGroups` or `IPCAccessControlFiles` directives in `/etc/usbguard/usbguard-daemon.conf` can be allowed control the daemon, but this draft version only restricts whether the daemon can write its own files. For tighter control, the IPC interface should be controlled by policy (or maybe something simple like `allow user_t usbguard_t:unix_stream_socket connectto;`) but here especially I'd request comments on what should be the appropriate level.

policy/modules/admin/usbguard.fc
```diff
@@ -0,0 +1,9 @@
+/etc/usbguard						gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/rules\.conf				gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/usbguard-daemon\.conf			gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/IPCAccessControl\.d(/.*)?			gen_context(system_u:object_r:usbguard_etc_t,s0)
```
bauen1: These could probably be merged into one (regex) entry.

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,66 @@
...
+allow usbguard_t self:capability { chown dac_override dac_read_search fowner };
+
+# /sys
+allow usbguard_t sysfs_t:dir read;
+allow usbguard_t sysfs_t:file { open read write };
+allow usbguard_t sysfs_t:lnk_file { getattr read };
+
+# /dev/shm
+allow usbguard_t tmpfs_t:dir { add_name create remove_name rmdir search setattr write };
+allow usbguard_t tmpfs_t:file { create map open read setattr unlink write };
```
bauen1: Please refer to https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide#local-policy-rules

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,66 @@
...
+allow usbguard_t var_log_t:dir search;
+
+type usbguard_var_log_t;
+allow usbguard_t usbguard_var_log_t:file manage_file_perms;
+logging_log_file(usbguard_var_log_t)
+logging_log_filetrans(usbguard_t, usbguard_var_log_t, file)
+manage_files_pattern(usbguard_t, usbguard_var_log_t, usbguard_var_log_t)
+
+allow usbguard_t self:netlink_kobject_uevent_socket { bind create read setopt };
+allow usbguard_t self:unix_stream_socket { accept listen };
```
bauen1: If possible use object permissions set as defined in `policy/support/obj_perm_sets.spt`

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,66 @@
...
+gen_require(`
+	type var_log_t;
+	type var_t;
+        type sysfs_t;
+        class dir { add_name create getattr open read remove_name rmdir search setattr write };
+        class file { create getattr map open read setattr unlink write };
+        class filesystem getattr;
+        class lnk_file { getattr read };
+        class netlink_kobject_uevent_socket { bind create read setopt };
+        class unix_stream_socket { accept listen };
```
bauen1: These aren't necessary, since `policy_module` already requires them for you.

policy/modules/admin/usbguard.fc
topimiettinen: Fixed in 36b7876 (updated while you reviewed).

topimiettinen:
@bauen1 Pushed an updated version, hopefully should be in line with StyleGuide.

policy/modules/admin/usbguard.te
topimiettinen: fixed

policy/modules/admin/usbguard.te
topimiettinen: fixed

policy/modules/admin/usbguard.te
topimiettinen: Hopefully now in line

policy/modules/admin/usbguard.fc
```diff
@@ -0,0 +1,8 @@
+/etc/usbguard						gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/IPCAccessControl\.d(/.*)?			gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/rules\.conf				gen_context(system_u:object_r:usbguard_etc_t,s0)
+/etc/usbguard/usbguard-daemon\.conf			gen_context(system_u:object_r:usbguard_etc_t,s0)
```
doverride: just use:
```
/etc/usbguard -d gen_context(system_u:object_r:usbguard_etc_t,s0)
/etc/usbguard/.* gen_context(system_u:object_r:usbguard_etc_t,s0)
```
and get it over with

doverride: use conf instead of etc

policy/modules/admin/usbguard.fc
```diff
@@ -0,0 +1,8 @@
...
+
+/usr/sbin/usbguard-daemon			--	gen_context(system_u:object_r:usbguard_daemon_exec_t,s0)
+
+/var/log/usbguard(/.*)?					gen_context(system_u:object_r:usbguard_var_log_t,s0)
```
doverride: might as well use "usbguard_log_t"

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,58 @@
...
+# Declarations
+#
+
+## <desc>
+##	<p>
+##	Determine whether usbguard-daemon can modify its rules in
+##	/etc/usbguard.
+##	</p>
+## </desc>
+gen_tunable(usbguard_modify_rule_files, false)
```
doverride: not sure if this is worth a boolean

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,58 @@
...
+
+type usbguard_var_log_t;
+logging_log_file(usbguard_var_log_t)
+
+########################################
+#
+# Usbguard local policy
+#
+
+allow usbguard_t self:capability { chown dac_override dac_read_search fowner };
```
doverride: the dac_override is due to a bug in the ipc library. A fix for this is upstream so with that you might be able omit the dac_override thereby greatly enhance security

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,58 @@
...
+
+manage_dirs_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+manage_files_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+mmap_read_files_pattern(usbguard_t, usbguard_tmpfs_t, usbguard_tmpfs_t)
+fs_tmpfs_filetrans(usbguard_t, usbguard_tmpfs_t, { dir file })
+
+append_files_pattern(usbguard_t, usbguard_var_log_t, usbguard_var_log_t)
+create_files_pattern(usbguard_t, usbguard_var_log_t, usbguard_var_log_t)
+logging_log_filetrans(usbguard_t, usbguard_var_log_t, file)
+manage_files_pattern(usbguard_t, usbguard_var_log_t, usbguard_var_log_t)
```
doverride: You might be able to do without this. not sure.. however either you use the finer grained append_files_pattern + create_files_pattern + ??? or you use manage_files_pattern. no point in using both

policy/modules/admin/usbguard.te
topimiettinen: Allowing a daemon to write its own config files does not seem a good idea to me, but this is the default. I'd imagine that in more security conscious environments (where this tool would be also more useful than single user desktops) this would not be allowed. Could there be some other way than using a boolean?

policy/modules/admin/usbguard.te
topimiettinen: ok

policy/modules/admin/usbguard.fc
topimiettinen: ok, thanks

policy/modules/admin/usbguard.te
topimiettinen: Seems to work without.

policy/modules/admin/usbguard.fc
topimiettinen: done

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,57 @@
...
+
+dev_rw_sysfs(usbguard_t)
+
+files_read_etc_files(usbguard_t)
+list_dirs_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+read_files_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+tunable_policy(`usbguard_modify_rule_files',`
+	manage_dirs_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+	manage_files_pattern(usbguard_t, usbguard_conf_t, usbguard_conf_t)
+')
```
pebenito: This should be at the bottom to follow the style guide.

policy/modules/admin/usbguard.te
pebenito: What are the scenarios that it would be modifying it's own config files? Can you partition config files into RO and RW files (and corresponding types).

doverride:

>
> What are the scenarios that it would be modifying it's own config files? Can you partition config files into RO and RW files (and corresponding
> types).
>

It allows you to grant/deny access at runtime and to optionally make those
rules persistent. You might be able to limit it to "rule.conf" in /etc/usbguard.

-- 
Dominick Grift

topimiettinen:
The config files are:
- /etc/usbguard/usbguard-daemon.conf: main config file
- /etc/usbguard/IPCAccessControl.d/$UID: for each UID, a file specifies what the user is authorized to do (like list rules or allow/reject devices), for example:
```
Devices=modify list listen
Policy=list
Exceptions=listen
```
- /etc/usbguard/rules.conf: possibly writable rules for the USB devices, for example:
```
reject id 8087:8008
allow
```
removes one USB device with VID:DID 8087:8008.

Only /etc/usbguard/rules.conf need to be writable if authorized users are allowed to modify the rules, the other files can be RO. I claimed earlier that by default, the users can control the devices but at least in Debian, the administrator needs to enable this by either adding the users to `plugdev` group, or by adding files to IPCAccessControl.d.

So if the boolean is not overkill, it should control whether the daemon can write to /etc/usbguard/rules.conf (which should get new type usbguard_rules_t). Other files should keep the type usbguard_conf_t, which are RO for the daemon.

topimiettinen:
In this updated version, the RW/RO split seems to work, but there's something wrong with user access. I get the error below when trying to install the policy:

```
Failed to resolve booleanif statement at /var/lib/selinux/default/tmp/modules/100/auditadm/cil:853
semodule:  Failed!
```

pebenito:
The boolean name doesn't match what is used in the conditional statement: `usbguard_modify_rule_files` vs. `usbguard_user_modify_rule_files`.

topimiettinen:
Thanks, that was the problem.

policy/modules/admin/usbguard.fc
```diff
@@ -0,0 +1,7 @@
+/etc/usbguard					-d	gen_context(system_u:object_r:usbguard_conf_t,s0)
+/etc/usbguard/rules.conf				gen_context(system_u:object_r:usbguard_rules_t,s0)
+/etc/usbguard/.*					gen_context(system_u:object_r:usbguard_conf_t,s0)
```
pebenito: The lint fails on this regex. It should probably be `/etc/usbguard/.+`

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,62 @@
...
+
+type usbguard_rules_t;
+files_config_file(usbguard_rules_t)
+
+# /dev/shm
+type usbguard_tmpfs_t;
+files_tmpfs_file(usbguard_tmpfs_t)
+
+type usbguard_log_t;
+logging_log_file(usbguard_log_t)
```
pebenito: This should move up after the `usbguard_conf_t` decls.

policy/modules/admin/usbguard.te
```diff
@@ -0,0 +1,62 @@
...
+########################################
+#
+# Usbguard local policy
+#
+
+allow usbguard_t self:capability { chown dac_read_search fowner };
+allow usbguard_t self:netlink_kobject_uevent_socket create_socket_perms;
+allow usbguard_t self:unix_stream_socket rw_stream_socket_perms;
+
+dev_rw_sysfs(usbguard_t)
```
pebenito: This should move just above the tunable block.

policy/modules/admin/usbguard.fc
topimiettinen: OK

policy/modules/admin/usbguard.te
topimiettinen: fixed

topimiettinen:
Thanks for the review. I fixed the issues and removed WIP.

PR #194, opened 2020-03-10
Add dnl builtins to places in support macros where blocks ending in n…

dburgener:
…ewlines are being included.

This was done in many places previously, but not all.

Signed-off-by: Daniel Burgener <Daniel.Burgener@microsoft.com>

### DIFF
--- a/policy/support/loadable_module.spt
+++ b/policy/support/loadable_module.spt
@@ -81,7 +81,7 @@ define(`interface',` dnl
 	`define(`$1',` dnl
 	pushdef(`policy_call_depth',incr(policy_call_depth)) dnl
 	policy_m4_comment(policy_call_depth,begin `$1'(dollarsstar)) dnl
-	$2
+	$2 dnl
 	popdef(`policy_call_depth') dnl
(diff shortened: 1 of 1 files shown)


pebenito:
I'm not clear that this is needed. It seems like this would make it potentially more difficult to see where the blocks are in the generated files. Can you provide some before/after snippets of example output that this helps?

dburgener:
In looking to get specific examples, I noticed that there are some cases involving indentation where the original change proposed does make things less readable.  So I've changed the PR to only include the interface case, which I think is consistently an improvement.  The difference in the tmp files on this change amounts to this every time the end of an interface call is expanded (example taken from the expansion of abrt.te line 207):

With the change:
```
       allow abrt_t httpd_modules_t:dir { getattr search open read lock ioctl };
#line 206

#line 206
##### end apache_list_modules(abrt_t) depth: 0
```

Without the change:
```
        allow abrt_t httpd_modules_t:dir { getattr search open read lock ioctl };
#line 206

#line 206

#line 206
##### end apache_list_modules(abrt_t) depth: 0
```

This change matches the behavior of templates.

PR #196, opened 2020-03-17
Add further "watch" permissions

gtrentalancia:
Add further "watch" permissions.

Modified/updated modules:

- userdomain
- wm
- mozilla
- getty

### DIFF
--- a/policy/modules/system/getty.te
+++ b/policy/modules/system/getty.te
@@ -46,6 +46,7 @@ files_lock_filetrans(getty_t, getty_lock_t, file)
 allow getty_t getty_log_t:file { append_file_perms create_file_perms setattr_file_perms };
 logging_log_filetrans(getty_t, getty_log_t, file)
 
+allow getty_t getty_runtime_t:dir watch;
 manage_files_pattern(getty_t, getty_runtime_t, getty_runtime_t)
 files_pid_filetrans(getty_t, getty_runtime_t, file)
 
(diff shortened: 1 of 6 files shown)


policy/modules/system/getty.te
```diff
@@ -48,6 +48,7 @@ logging_log_filetrans(getty_t, getty_log_t, file)
 
 manage_files_pattern(getty_t, getty_runtime_t, getty_runtime_t)
 files_pid_filetrans(getty_t, getty_runtime_t, file)
+getty_watch_runtime_dir(getty_t)
```
pebenito: Since the interface is used here, it doesn't seem to warrant an interface yet.

policy/modules/system/userdomain.if
```diff
@@ -364,6 +368,8 @@ interface(`userdom_manage_home_role',`
 	# cjp: this should probably be removed:
 	allow $2 user_home_dir_t:dir { manage_dir_perms relabel_dir_perms };
 
+	userdom_watch_user_home_content_dirs($2)
```
pebenito: Same thing here, interface doesn't seem necessary yet.

policy/modules/system/getty.te
gtrentalancia: ACK

policy/modules/system/userdomain.if
gtrentalancia: ACK

gtrentalancia:
Updated with requested changes.

PR #152, opened 2020-01-03
policy for systemd-fstab-generator

bauen1:
Implementing a policy for systemd-fstab-generator instead of allowing `initrc_t` to write to `init_runtime_t`.

Tested on:
- [x] debian minimal install (enforcing=1)

Comments are appreciated.

PR #197, opened 2020-03-18
systemd generators

dsugar100:
The recent change to label systemd-fstab-generator (pull request #152) caused some problems for me on RHEL7.  Initially with the lvm2-activation-generator and then some in the systemd-sysv-generator.

This series of patches labels all generators with a generic type so now generators run in sysemd_generator_t domain (rather than initrc_t).

Then for the generators that have specific needs they are split into their own domains with added permissions.  I used the 'systemd_generator_type' as an attribute for all generators to grant initial base set of permissions.

Finally a new type is created for the files in /run/systemd/generator (systemd_generator_runtime_t) to better restrict write access to these set of files.

I can squash this all into a single commit if desired, but this shows my thought process through the series of changes.


pebenito:
Please squash.

I think I'm ok with this, leaving open for other comments, e.g. @doverride

doverride:
Creating a private type just for /run//systemd/generators is probably not optimal.
There are globally two kinds of units, ones that are written and ones that are read only, but it does not really matter much because the ones that are readonly can be manipulated by adding written drop ins.
I generally make a distinction between those two kinds of units.

The more important issue in my view is: "who has access to "write" these writable units? Arguably, generators and even systemd itself are relatively trusted compared to some regular process running either systemd-run (which essentially creates a writable unit and tells systemd to interpret it) or a process just dropping a generator (yes there is this concept of a runtime generator, ie, a process can just dump a script and systemd will run it.

systemd itself can not be prevented from "writting" writable units in the long run. for example the systemd set-property command essentially causes systemd to create units in iether /etc/systemd/system.control or /run/systemd/system.control (for runtime only) 

So yes theres little point in differentiating too much, making a distiction between readonly units (the ones in /usr/lib/ssytemd) and runtime units (the ones in /run and /etc) is probably to most reasonable compromise. the generators and systemd just need to be ble to create units period. The focus should probably be on who has access to systemctl set-property/systemd-run to create runtime units. or who has access to write the units directly to the filesystem?

doverride:
another consideration for making a distiction between runtime users and regular read only units is that generally runtime units all have the same label whether readonly units can have private labels.

even units in /etc/systemd should probably be considered to be runtime units that cannot reliable have private labels, simply because systemd allows you to create units there with systemctl and can create units there on your behalf. it does that often without setfscreatecon currently so any per unit level specs for /etc will not apply consistenty

So yes another distinction: which units can reliably be labeled individually? practically only the units in /usr/lib/system.

So that leaves us with:
1. differentiate between units in /usr/lib/systemd and units anywhere else
2. focus on untrusted processes access to create units in these non /usr/lib-locations either direct filesystem access or indirect access via systemctl/systemd-run

dsugar100:
@doverride thanks for the feedback, but I'm not sure I'm following all of it.  I think you are suggesting to not change the type of /run/systemd/generator and /run/systemd/generator.late but leave them as init_runtime_t.

But at the end of your second comment I think you are suggesting that it is best to let systemd do everything it needs to do and not run the generators in a different domain? 

So rather than making this change I should figure out what broke lvm2-activation-generator running at initrc_t and fix it that way?

doverride:
Dave Sugar <notifications@github.com> writes:

> @doverride thanks for the feedback, but I'm not sure I'm following all of it. I think you are suggesting to not change the type of
> /run/systemd/generator and /run/systemd/generator.late but leave them as init_runtime_t.

I suggest two main types of unit files: "readonly" (for /usr/lib) and
"runtime" (for every other location)

>
> But at the end of your second comment I think you are suggesting that it is best to let systemd do everything it needs to do and not run the
> generators in a different domain?

Well yes and no, I think you actually struck a pretty good balance
there.

LVM generator is actually a relatively compelling  reason to run the
confined as it wants read/write storage access.

Howver eventually systemd will need that access itself anyway to support
the RootImage= functionality.

So in the end my comment was not about running generators in the
systemd domain or not. Instead it was about how I believe it is probably
not optimal to create a private type for /run/sysemd/generator alone.

There are other such location that probably could use similar treatment
like /run/systemd/transient
>
> So rather than making this change I should figure out what broke lvm2-activation-generator running at initrc_t and fix it that way?

I am not opposed to this change and I think its a pretty good balance.
I am just suggesting reconsidering the unit labeling.

>
> —
>

-- 
Dominick Grift

doverride:
But , yes on this topic. pid1 probably eventually needs all the access that the generators need plus trusted entities like the package manager can just drop-in runtime generators and systemd will happily run them. So i suppose the question is then whether it makes sense to confine the short unning generators bundled with systemd. Its expensive....

In my new dssp3 policy i take a radical new approach: trust systemd --system, its generators, trust dbus--system, trust package managers, trust the default login shell (al of the above run with one and the same context associated with "the system". system services that are not targeted will also end up in that context. So its really an opt-in model. This makes the policy a lot less complex and less intrusive by default, but you can at the same time leverage the power of the model to great extent. (so its truly a targeted model in every sense of the word) With this model i am trying to not gt distracted by entities that eventually end up with broad access anyway. So, its not least privilege.

dsugar100:
Ok - I will propose the following types, feel free to make other suggestions:

systemd_generator_exec_t for the following (with other specific binaries labeled as needed):
/usr/lib/systemd/system-environment-generators/.*
/usr/lib/systemd/system-generators/.*
/usr/lib/systemd/user-environment-generators/.*
/usr/lib/systemd/user-generators/.*

systemd_generator_runtime_t:
/run/systemd/generator{.late,.early}*/.*

Then something else for here? maybe systemd_generator_etc_t or sysemd_generator_file_t (or use systemd_unit_file_t like other stuff in /etc/systemd/system)
I'm not seeing these directories being used on RHEL.  I can provide labels, but I'm not sure what to do for permissions.
/etc/systemd/system-generators/*
/etc/systemd/user-generators/*

Finally the systemd-generator docs list: maybe systemd_generator_runtime_t
/run/systemd/system-generators/*
/run/systemd/user-generators/*
but again I'm not seeing these directories used on RHEL so permissions may be a bit wrong for a system that is putting files here.

And I will stick with the attribute systemd_generator_type as a way to provide the base required access.

Does that sound like a reasonable next step?
Again, thanks for your feedback.

doverride:
Dave Sugar <notifications@github.com> writes:

> Ok - I will propose the following types, feel free to make other suggestions:
>
> systemd_generator_exec_t for the following (with other specific binaries labeled as needed):
> /usr/lib/systemd/system-environment-generators/.*
> /usr/lib/systemd/system-generators/.*
> /usr/lib/systemd/user-environment-generators/.*
> /usr/lib/systemd/user-generators/.*
>
> systemd_generator_runtime_t:
> /run/systemd/generator{.late,.early}/.
>
> Then something else for here? maybe systemd_generator_etc_t or sysemd_generator_file_t (or use systemd_unit_file_t like other stuff in
> /etc/systemd/system)
> I'm not seeing these directories being used on RHEL. I can provide labels, but I'm not sure what to do for permissions.
> /etc/systemd/system-generators/*
> /etc/systemd/user-generators/*
>
> Finally the systemd-generator docs list: maybe systemd_generator_runtime_t
> /run/systemd/system-generators/*
> /run/systemd/user-generators/*
> but again I'm not seeing these directories used on RHEL so permissions may be a bit wrong for a system that is putting files here.
>
> And I will stick with the attribute systemd_generator_type as a way to provide the base required access.
>
> Does that sound like a reasonable next step?
> Again, thanks for your feedback.

why do you want to label /run/systemd/generator type
systemd_generator_runtime_t?

it just a (runtime) unit location like for example /run/systemd/transient

the only difference between the two is that the generators create units
in the former and that systemd creates units in the later.

but eventually both locations share these common properties:

1. they both have units created at runtime. so the "service" checks will
apply to them.
2. they both have units with unpredictable names so you cannot
reasonably label units there with individual types (meaning that you
cannot enforce service access vectors on a per unit bases in these
locations and that for example if a process can start one unit there, it
can start all of them)

>
> —
>

-- 
Dominick Grift

dsugar100:
I am proposing systemd_generator_runtime_t as that fits the pattern of other labels in /run/systemd/ but I'm not particularly attached to the name.  The existing systemd_unit_t or init_runtime_t label could be used instead if you think that is more appropriate.  As you say, these are unit files.  

Or something new, maybe systemd_transient_unit_t may better fit the needs for these files.  It indicates it is a transient file.

Thoughts?   Other suggestions?

doverride:
Something not as specific as systemd_generator_runtime_t. transient or runtime unit sounds more appropriate but i would argue that if you introduce this then you probably should do it comprehensively and in a seperate patch.

So for the process perspective i think your patch is fine. but i dont think the systemd_generator_runtime_t file aspect has been addressed properly.

So if you leave out the systemd_generator_runtime_t aspect out of this patch then i have no comments at all.

policy/modules/system/systemd.fc
```diff
@@ -63,6 +70,8 @@
 
 /run/systemd/ask-password(/.*)?	gen_context(system_u:object_r:systemd_passwd_runtime_t,s0)
 /run/systemd/ask-password-block(/.*)?	gen_context(system_u:object_r:systemd_passwd_runtime_t,s0)
+/run/systemd/generator(/.*)?	gen_context(system_u:object_r:systemd_generator_runtime_t,s0)
+/run/systemd/generator\.late(/.*)?	gen_context(system_u:object_r:systemd_generator_runtime_t,s0)
```
pebenito: I agree with doverride, this should be revised as unit locations.

policy/modules/system/systemd.fc
```diff
@@ -18,8 +18,15 @@
 /usr/bin/systemd-notify			--	gen_context(system_u:object_r:systemd_notify_exec_t,s0)
 
 # Systemd generators
-/usr/lib/systemd/system-generators/systemd-fstab-generator	    --	    gen_context(system_u:object_r:systemd_fstab_generator_exec_t,s0)
-/usr/lib/systemd/system-generators/systemd-gpt-auto-generator	    --	    gen_context(system_u:object_r:systemd_gpt_generator_exec_t,s0)
+/usr/lib/systemd/system-environment-generators/.*				--	gen_context(system_u:object_r:systemd_generator_exec_t,s0)
+/usr/lib/systemd/system-generators/.*							--	gen_context(system_u:object_r:systemd_generator_exec_t,s0)
+/usr/lib/systemd/user-environment-generators/.*					--	gen_context(system_u:object_r:systemd_generator_exec_t,s0)
+/usr/lib/systemd/user-generators/.*								--	gen_context(system_u:object_r:systemd_generator_exec_t,s0)
```
pebenito: This should be revised to differentiate between system and user sessions.

policy/modules/system/systemd.fc
dsugar100: Ok, I will leave the changes for systemd_generator_runtime_t out and they can be considered in a future patch.  For now the type for /run/systemd/generator/.* will remain init_runtime_t.

I will have an updated squashed patch shortly.

doverride: i recorded some of my thoughts on systemd units: https://www.youtube.com/watch?v=FbZfe80usJI

dsugar100:
Ok, I have updated this commit to remove to squash the changes into a single commit, reverting the labeling of /run/systemd/generator back to init_runtime_t.

Let me know if there is anything else needed.

dsugar100:
Please wait to merge this just a bit longer.  I found another problem with the systemd-cryptsetup-generator this time that I'm looking into and hope to be able to update this patch shortly.  I'm basically seeing the following additional denials.

audit(1585500099.139:6): avc:  denied  { read } for  pid=1635 comm="systemd-cryptse"
    path="/run/systemd/generator/dev-mapper-luks\x2d6a613af0\x2d0a61\x2d462f\x2d8679\x2d1b0d964fbc88.device.d/.#90-device-timeout.confsOskdU" dev="tmpfs" ino=12243 scontext=system_u:system_r:systemd_generator_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=file permissive=1

audit(1585500099.139:7): avc:  denied  { setattr } for  pid=1635 comm="systemd-cryptse" name=".#90-device-timeout.confsOskdU" dev="tmpfs" ino=12243 scontext=system_u:system_r:systemd_generator_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=file permissive=1

audit(1585500099.139:8): avc:  denied  { rename } for  pid=1635 comm="systemd-cryptse" name=".#90-device-timeout.confsOskdU" dev="tmpfs" ino=12243 scontext=system_u:system_r:systemd_generator_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=file permissive=1

I'm planning on adding these permissions to the systemd_generator_type rather than make a new domain for this generator.

dsugar100:
Ok, I have updated the commit with the additional access required for systemd-cryptsetup-generator.

policy/modules/system/systemd.te
```diff
@@ -61,6 +62,14 @@ init_unit_file(systemd_binfmt_unit_t)
 type systemd_conf_t;
 files_config_file(systemd_conf_t)
 
+type systemd_generator_t;
+type systemd_generator_exec_t;
+init_system_domain(systemd_generator_t, systemd_generator_exec_t)
```
pebenito: After further thought, it seems like init_system_domain() should also go into the systemd_generator() interface, so then you have a block like this:
```
type systemd_generator_t;
￼type systemd_generator_exec_t;
systemd_unit_generator(systemd_generator_t, systemd_generator_exec_t)
```
I think we should also rename the interface to `systemd_unit_generator`.

dsugar100: Yes, that seems like a reasonable change.  And does make it a bit cleaner.  I have pushed the changes for this.  Let me know if there is anything else that is needed.

PR #198, opened 2020-03-18
Label more devices

dsugar100:
There is a STIG requirement (CCE-27326-8) that all files in /dev be labeled (something other than 'device_t').  On the systems I am working on there are a few files labeled device_t.  This series of simple patches resolves labeling and makes the STIG check happy.


policy/modules/kernel/devices.te
```diff
@@ -278,6 +284,13 @@ dev_node(tpm_device_t)
 type urandom_device_t;
 dev_node(urandom_device_t)
 
+#
+# uhid_device_t is the thpe of /dev/uhid - 
+# User-space I/O driver support for HID subsystem
+#
+type uhid_device_t;
+dev_node(uhid_device_t)
+
```
pebenito: Needs to move up to keep ordering.

pebenito: Also please squash and add sign-off.

dsugar100:
Changes applied and squashed to a single commit.  Let me know if anything else needs to be done.

PR #201, opened 2020-03-21
Makefile: add target rebuild-interface-db

cgzones:
Rebuild the policy interface database with 'sepolgen-ifgen'.
This database is required for reference style policy generation by
'audit2allow --reference'


Makefile
```diff
@@ -547,6 +550,15 @@ endif
...
 	$(verbose) $(INSTALL) -m 644 $(support)/Makefile.devel $(headerdir)/Makefile
 
+########################################
+#
+# Rebuild policy interface database
+#
+rebuild-interface-db:
+	$(verbose) rm -rf $(headerdir)
+	$(verbose) $(MAKE) install-headers
+	$(verbose) $(SEPOLGEN_IFGEN) $(VERBOSE_FLAG) -i $(headerdir)
```
pebenito: This should probably be reworked to have `build-interface-db` and `rebuild-interface-db` targets which would work like the `load` and `reload` targets in the `Rules.monolithic`.

Makefile
```diff
@@ -547,6 +550,17 @@ endif
...
+########################################
+#
+# Build policy interface database
+#
+build-interface-db: install-headers
+	$(verbose) $(SEPOLGEN_IFGEN) $(VERBOSE_FLAG) -i $(headerdir)
+
+rebuild-interface-db:
+	$(verbose) rm -rf $(headerdir)
+	$(verbose) $(MAKE) build-interface-db
```
pebenito: Wouldn't something like this work, without having to recursively call this makefile:
```
build-interface-db: install-headers $(tmpdir)/load

rebuild-interface-db $(tmpdir)/load:
	$(verbose) $(SEPOLGEN_IFGEN) $(VERBOSE_FLAG) -i $(headerdir)
	@touch $(tmpdir)/load
```
If something in headerdir needs to be deleted, my preference would be not to delete everything.

cgzones:
`sepolgen-ifgen` currently dislikes these policy constructs:

```
refpolicy/include/system/userdomain.if: Syntax error on line 185 ` [type=TICK]
```
https://github.com/SELinuxProject/refpolicy/blob/41963e7116d57d2ea1d63367bb34828125d739fa/policy/modules/system/userdomain.if#L185

```
refpolicy/include/kernel/storage.if: Syntax error on line 133 $2 [type=IDENTIFIER]
```
https://github.com/SELinuxProject/refpolicy/blob/41963e7116d57d2ea1d63367bb34828125d739fa/policy/modules/kernel/storage.if#L133

```
refpolicy/include/kernel/devices.if: Syntax error on line 2805 $2 [type=IDENTIFIER]
```
https://github.com/SELinuxProject/refpolicy/blob/41963e7116d57d2ea1d63367bb34828125d739fa/policy/modules/kernel/devices.if#L2805

```
refpolicy/include/kernel/devices.if: Syntax error on line 2883 $2 [type=IDENTIFIER]
```
https://github.com/SELinuxProject/refpolicy/blob/41963e7116d57d2ea1d63367bb34828125d739fa/policy/modules/kernel/devices.if#L2883

```
refpolicy/include/kernel/devices.if: Syntax error on line 2959 $2 [type=IDENTIFIER]
```
https://github.com/SELinuxProject/refpolicy/blob/41963e7116d57d2ea1d63367bb34828125d739fa/policy/modules/kernel/devices.if#L2959

I think it doesn't like a) quoted arguments to `gen_tunable` and b) non-quoted arguments to `tunable_policy`

cgzones:
With the patches from https://github.com/SELinuxProject/selinux/pull/247, the `interfac_info` map builds successfully

Makefile
pebenito: This does not seem to be addressed.

cgzones: The initial design for the targets was:
  - `build-*`: generate the interface - always run `sepolgen-ifgen`, so calling the target twice generates the database twice
  - `rebuild-*`: generate the interface but assure that no leftover interfaces are in the header directory (after a re-factorization, removal of an module ...)

But since Refpolicy takes backward-compatibility serious the reasons for the `rebuild-*` target probably do not apply.

Dropped the `rebuild-*` target.

PR #202, opened 2020-03-21
Rules: allow the usage of class sets in context_defaults

cgzones:
Allow class sets , e.g. defined in policy/support/obj_perm_sets.spt, to
be used in default_* statements in the file policy/context_defaults


cgzones:
> Needs signed-off-by.

Added

PR #203, opened 2020-03-21
RFC: introduce gen_system_context

cgzones:
Add a wrapper for gen_context, which hard-codes the user identity
`system_u`, the role `object_r` and defaults to the mls-sensitivity `s0`.

File contexts like
    `gen_context(system_u:object_r:foo_t,s0)`
can be simplified to
    `gen_system_context(foo_t)`


pebenito:
Thanks for looking at this mundane part of the policy. However, I'm not completely convinced it should be done. I see the reasons for this:
1. reduce duplication
2. reduce errors
3. make it easy to change common part of context

Since the macro name is relatively long by necessity, it does 1 but doesn't reduce it by much. It does 2, but these can be uncovered at build time by validation of the file_contexts. I believe 3 is too rare of a need to factor into the decision.

bauen1:
>     1. reduce duplication
> 

The reduction of duplication isn't a lot line-wise in terms of characters, but the "complexity" of every line is reduced and therefor becomes easier to read.
Checking that 1 macro with 3 duplicated "arguments" is what you expect it to be is harder than just checking if 1 macro is what you expect it to be.

> reduce errors

While typos can be catched by validating `file_contexts` this would, to some degree, ensure they don't happen in the first place as often.

pebenito:
Closing, however I'm not saying this, or something like this, can never be merged in the future.

PR #208, opened 2020-04-03
Setup domain for tpm2_* binaries

dsugar100:
The various /bin/tpm2_* binaries use dbus to communicate
with tpm2-abrmd and also can directly access /dev/tpmrm0.  This
seems like a way to help limit access to the TPM by running the
tpm_* binaries in their own domain.

I setup this domain because I have a process that needs to use
tpm2_hmac to encode something, but didn't want that domain to
have direct access to the TPM.  I did some basic testing to verify
that the other tpm2_* binaries have basically the same access needs.
But it wasn't through testing of all the tpm2_* binaries.

### DIFF
--- a/policy/modules/services/tpm2.fc
+++ b/policy/modules/services/tpm2.fc
@@ -1,4 +1,5 @@
 /usr/bin/tpm2-abrmd								--	gen_context(system_u:object_r:tpm2_abrmd_exec_t,s0)
+/usr/bin/tpm2_[^/]+									--	gen_context(system_u:object_r:tpm2_exec_t,s0)
 
 /usr/sbin/tpm2-abrmd								--	gen_context(system_u:object_r:tpm2_abrmd_exec_t,s0)
 
--- a/policy/modules/services/tpm2.if
+++ b/policy/modules/services/tpm2.if
@@ -2,6 +2,95 @@
 
 ########################################
 ## <summary>
+##	Execute tpm2_* processes
+##	in the caller domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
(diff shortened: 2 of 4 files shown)


policy/modules/services/tpm2.if
```diff
@@ -56,3 +145,23 @@ interface(`tpm2_status_abrmd',`
...
+## <summary>
+##	access tpm2-abrmd fifos
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access
+##	</summary>
+## </param>
+#
+interface(`tpm2_rw_pipes_abrmd',`
```
pebenito, line 159: `tpm2_rw_abrmd_pipes`

dsugar100: Updated - sorry for the delay - I was testing with some other changes and everything seems OK.

policy/modules/services/tpm2.fc
```diff
@@ -1,4 +1,5 @@
 /usr/bin/tpm2-abrmd								--	gen_context(system_u:object_r:tpm2_abrmd_exec_t,s0)
+/usr/bin/tpm2_.*									--	gen_context(system_u:object_r:tpm2_exec_t,s0)
```
pebenito, line 2: Need to replace `.*` with `.+`. This showed up in the CI build, though I'm not sure why the result isn't blocking the PR.

pebenito: I think I fixed the CI integration.

pebenito:
CI has a linking failure: https://travis-ci.org/github/SELinuxProject/refpolicy/jobs/675567067#L3419

Issue #190, opened 2020-02-20
Allow system_dbusd_t read/write privilege to tun_tap_device_t


dsommers:
I'm working on the [OpenVPN 3 Linux](https://github.com/OpenVPN/openvpn3-linux) project which is a brand new OpenVPN client which makes heavy use of D-Bus to solve a lot of challenges the current OpenVPN 2.x generation has on modern Linux systems.

OpenVPN 3 Linux depends heavily on D-Bus, where multiple daemons serve very specific task and the IPC happens over D-Bus.  One challenge we have on SELinux enabled systems (in particular Fedora and RHEL) is that SELinux does not allow the dbus-daemon to pass a FD from one D-Bus service to another one when the FD is tied to  `/dev/net/tun`.

Currently we ship our own SELinux policy to resolve this issue, which can be seen here in [openvpn3.te](https://github.com/OpenVPN/openvpn3-linux/blob/master/src/selinux/openvpn3.te).

The policy we wrote attempted to be a generic as possible (with the filename being the exception), as this doesn't look like an OpenVPN only related issue, but something which could hit anyone wanting to pass a FD to a tun device over D-Bus.

If this looks like a reasonable solution which could be applied to the the SELinux reference policy, I'm happy to submit a pull-request for it.

dsommers:
For reference, the AVC denial without this policy looks like this:

    type=AVC msg=audit(1582227542.557:3045): avc:  denied  { read write } for  pid=1741 comm="dbus-daemon" path="/dev/net/tun" dev="devtmpfs" ino=486 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:tun_tap_device_t:s0 tclass=chr_file permissive=0

This is from a RHEL-7.7 system.

pebenito:
Is this denial breaking the fd passing or is it simply generating a denial message? Since dbus is only passing the fd for the device, I wouldn't be surprised if this may be something that could be dontaudited.

dsommers:
The denial causes the D-Bus connection from the calling process to be disconnected (typical behavior when permission errors occurs).  Resulting in the calling process not to being able to continue working properly.  Allowing this event to happen, everything works as expected.

[...]

PR #209, opened 2020-04-03
dbus: Add tunable - dbus_can_pass_tuntap_fd

dsommers:
D-Bus services wanting to pass file descriptors for
tun/tap devices need to read/write privileges to /dev/tun.

Without this privilege the following denial will happen:

    type=AVC msg=audit(1582227542.557:3045): avc:  denied  { read write } for  pid=1741 comm="dbus-daemon" path="/dev/net/tun" dev="devtmpfs" ino=486 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:tun_tap_device_t:s0 tclass=chr_file permissive=0

This is needed by OpenVPN 3 Linux, where an unprivileged
process (openvpn3-service-client) requests a tun device
from a privileged service (openvpn3-service-netcfg) over
the D-Bus system bus.

GitHub-Issue: #190
Signed-off-by: David Sommerseth <davids@openvpn.net>

### DIFF
--- a/policy/modules/services/dbus.te
+++ b/policy/modules/services/dbus.te
@@ -9,6 +9,15 @@ gen_require(`
 # Declarations
 #
 
+## <desc>
+## <p>
+## Allow dbus-daemon system bus to access /dev/net/tun
+## which is needed to pass tun/tap device file descriptors
+## over D-Bus.  This is needed by openvpn3-linux.
+## </p>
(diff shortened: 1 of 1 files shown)


pebenito:
I ended up removing the "can" from the tunable name.

PR #210, opened 2020-04-04
systemd-user-runtime-dir: add missed required permissions

bauen1:
systemd-user-runtime-dir reads /proc/sys/kernel/osrelease and the
selinux config

Missed these in #183 or they where required by a recent upgrade of my server from debian stable to unstable.


bauen1:
I still have to test this.

bauen1:
Tested.

PR #211, opened 2020-04-06
Parallel segenxml

cgzones:
*  Makefile: Parallelize XML documentation generation
   This can decrease the time for 'make conf',
   e.g. with 4 cores from 8.8 to 4.9 seconds.

*  Makefile: generate temporary files for configuration in separate directory
   Use a separate directory so the clean target from Rules.* does not remove them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,7 +270,8 @@ generated_fc := $(basename $(foreach dir,$(all_layers),$(wildcard $(dir)/*.fc.in
 # when a generated file is already generated
 detected_mods := $(sort $(foreach dir,$(all_layers),$(wildcard $(dir)/*.te)) $(generated_te))
 
-modxml := $(addprefix $(tmpdir)/, $(detected_mods:.te=.xml))
+modxml := $(addprefix $(tmpdir)/,$(patsubst $(moddir)/%,%,$(filter $(moddir)/%,$(detected_mods:.te=.xml))))
+localmodxml := $(addprefix $(tmpdir)/,$(patsubst $(local_moddir)/%,%,$(filter $(local_moddir)/%,$(detected_mods:.te=.xml))))
 layerxml := $(sort $(addprefix $(tmpdir)/, $(notdir $(addsuffix .xml,$(all_layers)))))
 layer_names := $(sort $(notdir $(all_layers)))
 all_metaxml = $(call detect-metaxml, $(layer_names))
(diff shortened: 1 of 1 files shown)


pebenito:
I tried `make conf` with HEAD and then `make -j4 conf` with this PR, and it went up from 9.39s to 11.01s. With 8 jobs I got to 10.8s. Even with `QUIET=y` it didn't make a significant difference.

cgzones:
That's odd. I'll do some testing on different machines...

bauen1:
Tested on a 4 core macbook air:

```
$ # original
$ make bare
$ time make conf
real 0m25.433s

$ # this pr
$ make bare
$ time make conf
real 0m26.815s

$ # this pr 4 cores
$ make bare
$ time make conf -j4
real 0m14.515s
```

:+1:

cgzones:
Thanks for testing @bauen1.
There seems to be overhead for the Makefile dependency resolution (e.g. an increase for single job run).
Please also share the time it takes to configure the original branch; cause there is already multitasking by generating all the layers (top directories in policy/modules/) in parallel, but since the service layer is big and for each layer all the module documentation files are generated serially, that's a bottleneck.

bauen1:
> Please also share the time it takes to configure the original branch; cause there is already multitasking by generating all the layers (top directories in policy/modules/) in parallel, but since the service layer is big and for each layer all the module documentation files are generated serially, that's a bottleneck.
```
$ # branch master
$ make bare
$ time make conf -j4
real 0m17.265s
```

dburgener:
Results from my machine:

This pr make conf:
real	0m26.936s
user	0m21.642s
sys	0m4.550s

This pr make -j4 conf:
real	0m10.142s
user	0m21.487s
sys	0m4.385s

This pr make -j16 conf:
real	0m8.393s
user	0m23.393s
sys	0m4.514s

master make conf:
real	0m15.625s
user	0m13.275s
sys	0m2.746s

master make -j4 conf:
real	0m10.612s
user	0m14.138s
sys	0m2.834s

master make -j16 conf:
real	0m11.583s
user	0m16.372s
sys	0m3.076s

I'm seeing a massive hit for "make conf", similar performance on -j4 and a slight win on -j16.

cgzones:
Seems the Makefile dependency resolution gets heavier.
Seems not worth the trouble.

PR #212, opened 2020-04-07
Don't allow creating regular files in /dev

topimiettinen:
Init, init scripts and udisks don't need to be able to create regular
files in /dev.

Thanks to Jarkko Sakkinen for the idea.


topimiettinen:
@jsakkine-intel

PR #213, opened 2020-04-11
Makefile: generate temporary documentation files in separate directory 

cgzones:
Use a separate directory for files generated my 'make conf', so the clean targets from Rules.* do not remove them.


pebenito:
I'm a bit confused. The heading says config files, but this seems to be documentation/xml files. Please clarify.

cgzones:
It is a bit confusing because the files are documentation ones, but the are generated by `make conf`.
I improved the commit message.

PR #214, opened 2020-04-13
mozilla: allow firefox to use user namespaces for sandboxing

bauen1:
See https://wiki.mozilla.org/Security/Sandbox/Specifics#Linux for reference.


bauen1:
https://travis-ci.org/github/bauen1/refpolicy/jobs/674486804 failed because of #215

PR #220, opened 2020-04-14
Fix mismatches between object class and permission macro.

dburgener:
In many cases, this won't result in a change in the actual policy generated, but if the definitions of macros are changed going forward, the mismatches could cause issues.

### DIFF
--- a/policy/modules/admin/su.if
+++ b/policy/modules/admin/su.if
@@ -55,7 +55,7 @@ template(`su_restricted_domain_template', `
 	# By default, revert to the calling domain when a shell is executed.
 	corecmd_shell_domtrans($1_su_t,$2)
 	allow $2 $1_su_t:fd use;
-	allow $2 $1_su_t:fifo_file rw_file_perms;
+	allow $2 $1_su_t:fifo_file rw_fifo_file_perms;
 	allow $2 $1_su_t:process sigchld;
 
 	kernel_read_system_state($1_su_t)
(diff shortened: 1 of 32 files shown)


policy/modules/admin/su.if
```diff
@@ -55,7 +55,7 @@ template(`su_restricted_domain_template', `
 	# By default, revert to the calling domain when a shell is executed.
 	corecmd_shell_domtrans($1_su_t,$2)
 	allow $2 $1_su_t:fd use;
-	allow $2 $1_su_t:fifo_file rw_file_perms;
+	allow $2 $1_su_t:fifo_file rw_fifo_file_perms;
```
pebenito, line 58: There are several domtrans interfaces that should be fixed to use `rw_inherited_fifo_file_perms`, to be consistent with what is in `domtrans_pattern()`

dburgener: I got a few of those in the first round.  I think I got them all with the latest update.

policy/modules/services/ssh.te
```diff
@@ -183,7 +183,7 @@ tunable_policy(`allow_ssh_keysign',`
 	domain_auto_transition_pattern(ssh_t, ssh_keysign_exec_t, ssh_keysign_t)
 	allow ssh_keysign_t ssh_t:fd use;
 	allow ssh_keysign_t ssh_t:process sigchld;
-	allow ssh_keysign_t ssh_t:fifo_file rw_file_perms;
+	allow ssh_keysign_t ssh_t:fifo_file rw_inherited_fifo_file_perms;
```
pebenito, line 186: This looks like it could be simplified into `domtrans_pattern()`.

PR #225, opened 2020-04-16
Add divert to generated_definitions creation.

dburgener:
During normal m4 parsing, m4 outputs a blank line for each define() call.  This results in the first roughly 500 lines of the .tmp files for each module being largely blank lines.  Adding divert() calls to the m4 generation for generated_definitions redirects this output, so the beginning of the actual policy appears near the top of the .tmp files.

Signed-off-by: Daniel Burgener <Daniel.Burgener@microsoft.com>

### DIFF
--- a/Rules.modular
+++ b/Rules.modular
@@ -128,9 +128,11 @@ $(tmpdir)/pre_te_files.conf: $(base_pre_te_files)
 $(tmpdir)/generated_definitions.conf:
 	@test -d $(tmpdir) || mkdir -p $(tmpdir)
 # define all available object classes
-	$(verbose) $(genperm) $(avs) $(secclass) > $@
+	@cat $(m4divert) > $@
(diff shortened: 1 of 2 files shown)


dburgener:
There seem to be two issue with using the divert files as you suggest.

1. m4support already includes the divert files.  To also wrap generated_definitions.conf, we'd need to include them a second time in the dependencies for $(tmpdir)/%.mod, but make automatically removes extra dependencies, so this line:

`$(tmpdir)/%.mod: $(m4support) $(m4divert) $(tmpdir)/generated_definitions.conf $(m4undivert) $(tmpdir)/all_interfaces.conf %.te $(m4terminate)`

Resolves to the following set of files during a build

pebenito:
Why not cat the files into $@ instead of the echo?

Issue #221, opened 2020-04-15
Access to udev "database"


bigon:
Hello,

The udev module still references the `udev_tbl_t` as being stored in `/dev`, but these days, it's located in `/run`. That prevents some applications (like pcscd) to work properly.

Red Hat went the way of removing the `udev_tbl_t` type completely, see https://github.com/fedora-selinux/selinux-policy/commit/382acd84f3

Would that be the road to go as well?

pebenito:
Sounds like it is the right choice.

PR #230, opened 2020-04-17
Drop support for /dev/.udev and like

bigon:
This location is gone for quite some times and the udevdb has been moved
to /run/udev.

Drop the udev_tbl_t and deprecate the udev_read_db() function

This inspired from changes in the Red Hat policy

Signed-off-by: Laurent Bigonville <bigon@bigon.be>

Fixes: #221

### DIFF
--- a/policy/modules/apps/pulseaudio.te
+++ b/policy/modules/apps/pulseaudio.te
@@ -235,7 +235,6 @@ optional_policy(`
 optional_policy(`
 	udev_read_pid_files(pulseaudio_t)
 	udev_read_state(pulseaudio_t)
-	udev_read_db(pulseaudio_t)
 ')
 
 optional_policy(`
--- a/policy/modules/services/colord.te
+++ b/policy/modules/services/colord.te
@@ -133,7 +133,6 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(colord_t)
 	udev_read_pid_files(colord_t)
 ')
 
--- a/policy/modules/services/modemmanager.te
+++ b/policy/modules/services/modemmanager.te
@@ -55,6 +55,5 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(modemmanager_t)
 	udev_manage_pid_files(modemmanager_t)
 ')
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -238,7 +238,7 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(auditd_t)
+	udev_read_pid_files(auditd_t)
 ')
 
 ########################################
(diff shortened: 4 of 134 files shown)


policy/modules/apps/pulseaudio.te
```diff
@@ -235,7 +235,7 @@ optional_policy(`
 optional_policy(`
 	udev_read_pid_files(pulseaudio_t)
 	udev_read_state(pulseaudio_t)
-	udev_read_db(pulseaudio_t)
+	udev_read_pid_files(pulseaudio_t)
```
pebenito, line 238: Already has the permission.

This prompts a broader question: If the domain doesn't already have the /run access, maybe there no longer is a need? Perhaps we should be dropping the permission instead.

policy/modules/services/colord.te
```diff
@@ -133,7 +133,6 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(colord_t)
```
pebenito, line 136: Why was this removed, but the others switched over to the pid one?

Same question for consolekit below.

policy/modules/services/modemmanager.te
```diff
@@ -55,6 +55,6 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(modemmanager_t)
+	udev_read_pid_files(modemmanager_t)
 	udev_manage_pid_files(modemmanager_t)
```
pebenito, line 59: Already has more permissions.

policy/modules/system/logging.te
```diff
@@ -607,7 +607,7 @@ optional_policy(`
 ')
 
 optional_policy(`
-	udev_read_db(syslogd_t)
+	udev_read_pid_files(syslogd_t)
 	# for systemd-journal to read seat data from /run/udev/data
 	udev_read_pid_files(syslogd_t)
```
pebenito, line 612: Already has the perms

policy/modules/services/colord.te
bigon: The call to `udev_read_pid_files(colord_t)` is just bellow :)

policy/modules/system/udev.if
```diff
@@ -214,10 +214,10 @@ interface(`udev_manage_rules_files',`
 #
 interface(`udev_dontaudit_search_db',`
 	gen_require(`
-		type udev_tbl_t;
+		type udev_runtime_t;
 	')
 
-	dontaudit $1 udev_tbl_t:dir search_dir_perms;
+	dontaudit $1 udev_runtime_t:dir search_dir_perms;
```
pebenito, line 220: This interface should be deprecated too.

pebenito:

bigon:
I think that the rework should still be done and drop support for the ancient `/dev/.udev` location but I'm not sure I'll have time to continue on this

bigon:
Thx

PR #234, opened 2020-04-19
Allow systemd-networkd to handle ICMP and DHCP packets

topimiettinen:
Allow systemd-networkd to send and receive ICMPv6 Router Solicitation
and Router Advertisement packets (in reality all unlabeled packets)
and DHCP client packets.

### DIFF
--- a/policy/modules/kernel/corenetwork.if.m4
+++ b/policy/modules/kernel/corenetwork.if.m4
@@ -872,6 +872,13 @@ create_packet_interfaces($1_client)
 create_packet_interfaces($1_server)
 ')
 
+#
+# network_packet_simple(packet_name)
+#
+define(`network_packet_simple',`
+create_packet_interfaces($1)
+')
+
 # create_ibpkey_*_interfaces(name, subnet_prefix, pkeynum,mls_sensitivity)
 # (these wrap create_port_interfaces to handle attributes and types)
 define(`create_ibpkey_type_interfaces',`create_ibpkey_interfaces($1,ibpkey_t,type,determine_reserved_capability(shift($*)))')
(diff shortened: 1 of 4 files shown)


dsugar100:
I feel like you shouldn't need to use corenet_sendrecv_unlabeled_packets(systemd_networkd_t).  You should be able to label the packets.

topimiettinen:
It's a big hole, though similar rules exist for dhcpd and NetworkManager.

There's indeed `network_packet()`, so perhaps it's enough to add `network_packet(icmp)` somewhere?

dsugar100:
It looks like dhcp and networkmanager both use corenet_all_recvfrom_unlabeled which I'm pretty sure is different than corenet_sendrecv_unlabeled_packets (though I'm sure I can't explain the differences between unlabeled packets and unlabeled associations).

topimiettinen:
I think associations and peers are related to labeled networking with IPSec or CIPSO, though there's awfully little info for SELinux networking beyond basic port types.

In the updated version there's ICMP packet type, replacing need to access unlabeled packets. Perhaps it should be used also elsewhere.

topimiettinen:
`network_packet(icmp)` generates distinct `icmp_client_packet_t` and`icmp_server_packet_t`. But I'm not sure how all ICMP/ICMPv6 packets fit to this model. For some packets, like Echo Request / Echo Reply, or Router Solicitation / Router Advertisement the relation is clear. But what about for example error codes like Destination Unreachable, are they server or client packets?

What if there was only one packet type, `icmp_packet_t` without server/client distinction, or this type could be in addition to `icmp_client_packet_t` and `icmp_server_packet_t`? Then systemd-networkd should need access to `icmp_client_packet_t` and `icmp_packet_t`.

pebenito:
I think having only ICMP packet type makes sense.

topimiettinen:
Updated with only `icmp_packet_t`.

Issue #235, opened 2020-04-20
tmp/all_interfaces.conf keeps content of previous builds


cgzones:
Is there a reason why `tmp/all_interfaces.conf` keeps the content of old builds.

https://github.com/SELinuxProject/refpolicy/blob/dd04789465b602e7891b4c9511c1c1c1235a117c/Rules.modular#L139-L145

In particular line 142 appends to the temporary file.

cgzones:
Any comments?

PR #236, opened 2020-04-20
Override old all_interfaces.conf.tmp file

cgzones:
Do not keep interfaces from previous builds.

Fixes: #235


pebenito:
Can this proceed?

PR #238, opened 2020-04-22
Build and install Netfilter rules

topimiettinen:
Build SECMARK rules for iptables and NFT, install them as
/etc/selinux/$NAME/contexts/netfilter_contexts{,.nft}.

### DIFF
--- a/Rules.modular
+++ b/Rules.modular
@@ -222,7 +222,7 @@ clean:
 	$(verbose) rm -f $(base_conf)
 	$(verbose) rm -f $(base_fc)
 	$(verbose) rm -f $(builddir)*.pp
-	$(verbose) rm -f $(net_contexts)
+	$(verbose) rm -f $(net_contexts) $(net_contexts_nft)
 	$(verbose) rm -fR $(tmpdir)
 
 .PHONY: default all policy base modules install load clean validate
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /file_contexts
 /homedir_template
 /net_contexts
+/net_contexts.nft
 /policy.conf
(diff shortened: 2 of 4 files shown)


Rules.modular
```diff
@@ -45,7 +45,7 @@ base: $(base_pkg)
 
 modules: $(mod_pkgs)
 
-install: $(instpkg) $(appfiles)
+install: $(instpkg) $(appfiles) $(ncpath) $(ncpath_nft)
```
pebenito, line 48: I think this (and the equivalent change in `Rules.monlithic` is the only thing I disagree with. Since SECMARK is rarely used, I don't think it should be installed by default. I'd be fine with a install-netcontexts or similar new target.

pebenito: Actually, on further thought, the reason this does not exist current for modular policy is that the original intent was that modules could provide labeling fragments, like the .fc files. I think having the modular policy install net contexts is broken.

topimiettinen: I suppose SECMARK is not used because it's not so easy to use without good documentation and examples. The files are not read by SELinux itself but external tools like Netfilter, so perhaps they should be installed in `/usr/share/doc/selinux/examples` or something like that instead?

pebenito: That would work too. I'd be fine with an install target for docs.

topimiettinen: Labeling fragments for modules also make sense. Then if a module is obsoleted (or just disabled), also related labeling and port rules are removed. It would be simplest just to move all `network_port()` items to new per module labeling files. This probably would need changes to semodule, in addition to m4 machinery and gennetfilter.

PR #239, opened 2020-04-22
Fix bird2

bauen1:
On debian the bird socket has moved to `/run/bird/bird.ctl` and config files are now under `/etc/bird/`

### DIFF
--- a/policy/modules/services/bird.fc
+++ b/policy/modules/services/bird.fc
@@ -11,3 +11,5 @@
 /var/log/bird\.log.*	--	gen_context(system_u:object_r:bird_log_t,s0)
 
 /run/bird\.ctl	-s	gen_context(system_u:object_r:bird_runtime_t,s0)
+
+/run/bird(/.*)?		gen_context(system_u:object_r:bird_runtime_t,s0)
--- a/policy/modules/services/bird.te
+++ b/policy/modules/services/bird.te
@@ -26,17 +26,23 @@ files_pid_file(bird_runtime_t)
 # Local policy
 #
 
-allow bird_t self:capability net_admin;
+allow bird_t self:capability { net_admin net_raw };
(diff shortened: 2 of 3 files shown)


policy/modules/services/bird.fc
```diff
@@ -11,3 +11,5 @@
 /var/log/bird\.log.*	--	gen_context(system_u:object_r:bird_log_t,s0)
 
 /run/bird\.ctl	-s	gen_context(system_u:object_r:bird_runtime_t,s0)
+
+/run/bird	-d	gen_context(system_u:object_r:bird_runtime_t,s0)
```
pebenito, line 15: Why not have the directory's contents too?

policy/modules/services/bird.te
```diff
@@ -26,17 +26,23 @@ files_pid_file(bird_runtime_t)
 # Local policy
 #
 
-allow bird_t self:capability net_admin;
+allow bird_t self:capability { net_admin net_raw };
 allow bird_t self:netlink_route_socket create_netlink_socket_perms;
 allow bird_t self:tcp_socket create_stream_socket_perms;
+allow bird_t self:unix_stream_socket { listen accept };
```
pebenito, line 32: Increase perms to `create_stream_socket_perms`. The other perms from that set are hiddin in the `logging_send_syslog_msg()` call.

policy/modules/services/bird.fc
bauen1: Fixed.

policy/modules/services/bird.te
bauen1: Fixed.

pebenito:
Please squash the commits and then I'll merge.

bauen1:
Sorry, I accidentally deleted the branch.

PR #240, opened 2020-04-22
Allow users to use systemd run

bauen1:
This can be used to e.g. create temporary units under `systemd --user`


policy/modules/system/userdomain.if
```diff
@@ -864,6 +864,7 @@ template(`userdom_common_user_template',`
 
 	optional_policy(`
 		systemd_role_template($1, $1_r, $1_t)
+		systemd_exec_systemd_run($1_t)
```
pebenito, line 867: It seems like this call should be part of the`systemd_role_template()` instead.

PR #241, opened 2020-04-22
ssh: fix for debian wrapper script

bauen1:
debian ships a wrapper script that moves the ssh-agent socket to
/run/user/$UID/openssh_agent

### DIFF
--- a/policy/modules/services/ssh.if
+++ b/policy/modules/services/ssh.if
@@ -408,6 +408,10 @@ template(`ssh_role_template',`
 	allow $3 $1_ssh_agent_t:fifo_file rw_inherited_fifo_file_perms;
 	allow $3 $1_ssh_agent_t:process sigchld;
 
+	ifdef(`distro_debian', `
+		userdom_user_runtime_filetrans($1_ssh_agent_t, ssh_agent_tmp_t, sock_file)
+	')
+
 	tunable_policy(`use_nfs_home_dirs',`
 		fs_manage_nfs_files($1_ssh_agent_t)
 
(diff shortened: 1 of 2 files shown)


policy/modules/services/ssh.if
```diff
@@ -361,6 +361,10 @@ template(`ssh_role_template',`
 	manage_sock_files_pattern($1_ssh_agent_t, ssh_agent_tmp_t, ssh_agent_tmp_t)
 	files_tmp_filetrans($1_ssh_agent_t, ssh_agent_tmp_t, { dir sock_file })
 
+	ifdef(`distro_debian', `
+		userdom_user_runtime_filetrans($1_ssh_agent_t, ssh_agent_tmp_t, sock_file, "openssh_agent")
+	')
+
```
pebenito, line 367: This should move down immediately above the `use_nfs_home_dirs` tunable block.

policy/modules/services/ssh.if
```diff
@@ -408,6 +408,10 @@ template(`ssh_role_template',`
 	allow $3 $1_ssh_agent_t:fifo_file rw_inherited_fifo_file_perms;
 	allow $3 $1_ssh_agent_t:process sigchld;
 
+	ifdef(`distro_debian', `
+		userdom_user_runtime_filetrans($1_ssh_agent_t, ssh_agent_tmp_t, sock_file, "openssh_agent")
```
pebenito, line 412: I missed this the first time, but the name (`"openssh_agent"`) doesn't seem necessary here.

PR #249, opened 2020-05-04
netutils: allow ping to send and receive ICMP packets

topimiettinen:
Let ping send and receive ICMP packets when Netfilter SECMARK packet
labeling is active.


topimiettinen:
This rule could replace the more permissive rules which allow raw and unlabeled packets, if SECMARK (maybe also Netlabel) labeling was always active, or perhaps there could be a new boolean to deny access to raw and unlabeled packets.

PR #253, opened 2020-05-08
Misc cleanup

cgzones:
* samba: fix wrong interface context smbd_runtime_t
* chromium: drop dead conditional block
* example: use module name matching file name
* consolesetup: drop unused requires
* unconfined: clarify unconfined_t stub usage in unconfined_domain_noaudit()
* portage: drop bizarre conditional TODO blocks
* init/systemd: move systemd_manage_all_units to init_manage_all_units
* tpm2: small fixes
* xserver: declare all callers of template xserver_common_x_domain_template as template
* files/logging: move var_run_t filecontext to defining module
* files/miscfiles: move usr_t filecontext to defining module
* chromium/libraries: move lib_t filecontext to defining module

### DIFF
--- a/doc/example.te
+++ b/doc/example.te
@@ -1,5 +1,5 @@
 
-policy_module(myapp,1.0.0)
+policy_module(example,1.0.0) # a non-base module name must match the file name
 
 ########################################
 #
--- a/policy/modules/admin/portage.if
+++ b/policy/modules/admin/portage.if
@@ -195,14 +195,6 @@ interface(`portage_compile_domain',`
 		fs_manage_nfs_files($1)
 		fs_manage_nfs_symlinks($1)
 	')
-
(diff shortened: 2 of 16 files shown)


policy/modules/services/xserver.if
```diff
@@ -335,6 +335,29 @@ interface(`xserver_non_drawing_client',`
...
+##	is the prefix for user_t).
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Client domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`xserver_common_x_domain',`
```
pebenito, line 360: While this is not declaring any types, it is still a template because it uses `$1_input_xevent_t`; it is referencing a templated type.

cgzones:
Dropped the `xserver` changes.

PR #255, opened 2020-05-11
application: applications can be executed from ssh without pty

bauen1:
For example ansible uses `ssh localhost sudo id` to become root.
This doesn't appear to be necessary in redhat due to https://src.fedoraproject.org/rpms/openssh/blob/master/f/openssh-6.6p1-privsep-selinux.patch

### DIFF
--- a/policy/modules/system/application.te
+++ b/policy/modules/system/application.te
@@ -13,6 +13,10 @@ optional_policy(`
 optional_policy(`
 	ssh_sigchld(application_domain_type)
 	ssh_rw_stream_sockets(application_domain_type)
+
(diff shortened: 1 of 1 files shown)


policy/modules/system/application.te
```diff
@@ -13,6 +13,9 @@ optional_policy(`
 optional_policy(`
 	ssh_sigchld(application_domain_type)
 	ssh_rw_stream_sockets(application_domain_type)
+	ifndef(`distro_redhat', `
```
pebenito, line 16: Nit: please add a blank line after line 15.

PR #260, opened 2020-05-15
can_exec(): move from misc_macros to misc_patterns

cgzones:
The file misc_macros.spt is due heavy usage of the m4 language
hard to parse for third party tools.
Move the macro can_exec() to misc_patterns.spt, which contains
only interface like define blocks.

### DIFF
--- a/policy/support/misc_macros.spt
+++ b/policy/support/misc_macros.spt
@@ -64,12 +64,6 @@ define(`gen_context',`$1`'ifdef(`enable_mls',`:$2')`'ifdef(`enable_mcs',`:s0`'if
 
 ########################################
 #
-# can_exec(domain,executable)
-#
-define(`can_exec',`allow $1 $2:file { mmap_exec_file_perms ioctl lock execute_no_trans };')
-
-########################################
-#
 # gen_bool(name,default_value)
 #
 define(`gen_bool',`
--- a/policy/support/misc_patterns.spt
+++ b/policy/support/misc_patterns.spt
@@ -111,3 +111,12 @@ define(`admin_process_pattern',`
 
 	allow $1 $2:process { ptrace signal_perms };
 ')
+
+#
+# File execution pattern
+#
+# Parameters:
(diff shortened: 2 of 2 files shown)


doverride:
cgzones <notifications@github.com> writes:

> The file misc_macros.spt is due heavy usage of the m4 language
> hard to parse for third party tools.
> Move the macro can_exec() to misc_patterns.spt, which contains
> only interface like define blocks.

the existing exec_files_pattern() is can_exec()'s pattern equivalent

>
> --------------------------------------------------------------------------------------------------------------------------------------
>
>
>   https://github.com/SELinuxProject/refpolicy/pull/260
>
> Commit Summary
>
> * can_exec(): move from misc_macros to misc_patterns
>
> File Changes
>
> * M policy/support/misc_macros.spt (6) 
> * M policy/support/misc_patterns.spt (9) 
>
> Patch Links:
>
> * https://github.com/SELinuxProject/refpolicy/pull/260.patch
> * https://github.com/SELinuxProject/refpolicy/pull/260.diff
>
> —
>

-- 
Dominick Grift

cgzones:
`exec_files_pattern()` is not exactly an equivalent, cause it requires the type of the containing directory, which in many cases is `bin_t` and therefore not in scope.

dburgener:
Every single pattern in the support directory consists of multiple lines, and ends with the suffix "_pattern".  It seems to me as though can_exec() fits more at home in the macros (although it is a bit of a corner case in there as well since it I believe is the only one that generates an allow rule).

I'd say that third party tools can probably handle this situation fairly easily by just treating can_exec() as a special case.

doverride:
cgzones <notifications@github.com> writes:

> exec_files_pattern() is not exactly an equivalent, cause it requires the type of the containing directory, which in many cases is
> bin_t and therefore not in scope.

True, but that is what makes it a pattern and why can_exec() does not
belong there

>
> —
>

-- 
Dominick Grift

PR #261, opened 2020-05-17
Confined debian fixes

bauen1:
A number of changes necessary to allow debian (and probably other systemd based installations) to boot without the unconfined module.

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -79,10 +79,8 @@ mls_trusted_object(initctl_t)
 
 type initrc_t, init_script_domain_type, init_run_all_scripts_domain;
 type initrc_exec_t, init_script_file_type;
-domain_type(initrc_t)
-domain_entry_file(initrc_t, initrc_exec_t)
+init_domain(initrc_t, initrc_exec_t)
 init_named_socket_activation(initrc_t, init_runtime_t)
-role system_r types initrc_t;
 # should be part of the true block
 # of the below init_upstart tunable
 # but this has a typeattribute in it
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -433,6 +433,7 @@ ifdef(`init_systemd',`
 	logging_send_audit_msgs(init_t)
 	logging_relabelto_devlog_sock_files(init_t)
 	logging_relabel_generic_log_dirs(init_t)
+	logging_audit_socket_activation(init_t)
 
 	# lvm2-activation-generator checks file labels
 	seutil_read_file_contexts(init_t)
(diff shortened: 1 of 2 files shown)


policy/modules/system/init.te
```diff
@@ -78,10 +82,8 @@ mls_trusted_object(initctl_t)
 
 type initrc_t, init_script_domain_type, init_run_all_scripts_domain;
 type initrc_exec_t, init_script_file_type;
-domain_type(initrc_t)
-domain_entry_file(initrc_t, initrc_exec_t)
+init_daemon_domain(initrc_t, initrc_exec_t)
```
pebenito, line 85: `init_daemon_domain()` is for long-running processes. `initrc_t` is not long-running. It is also for transitions _out of_ `initrc_t`. `init_domain()` may be acceptable.

policy/modules/system/init.te
```diff
@@ -61,6 +61,10 @@ role system_r types init_t;
 type init_runtime_t alias init_var_run_t;
 files_pid_file(init_runtime_t)
 
+ifdef(`init_systemd',`
+	init_mountpoint(init_runtime_t)
```
pebenito, line 65: This does not need to be in a `init_systemd` block.

policy/modules/system/init.te
```diff
@@ -78,10 +79,8 @@ mls_trusted_object(initctl_t)
 
 type initrc_t, init_script_domain_type, init_run_all_scripts_domain;
 type initrc_exec_t, init_script_file_type;
-domain_type(initrc_t)
-domain_entry_file(initrc_t, initrc_exec_t)
+init_domain(initrc_t, initrc_exec_t)
```
pebenito, line 82: There is a `init_domtrans_script(init_t)` later in the file that can be dropped.

bauen1: What about this part of `init_domtrans_script` ?
```
ifdef(`enable_mcs',`
    range_transition $1 initrc_exec_t:process s0;
')

ifdef(`enable_mls',`
    range_transition $1 initrc_exec_t:process s0 - mls_systemhigh;
')
```

pebenito: Replaced with calls to `init_ranged_domain()` in similar ifdefs.

pebenito: the `init_ranged_domain()` calls would be in init.te

bauen1: Do you want me to add a call to `init_ranged_domain` instead of `init_domtrans_script` since I don't see how the `range_transition` statements are applied otherwise.

bauen1: Like this ?

pebenito:
Missing a sign-off.

policy/modules/system/init.te
pebenito: Yes.

PR #262, opened 2020-05-17
Misc fixes 1

bauen1:
Things that don't warrant their own pull-request


policy/modules/kernel/corecommands.fc
```diff
@@ -178,7 +178,9 @@ ifdef(`distro_gentoo',`
 /usr/lib/fence(/.*)?			gen_context(system_u:object_r:bin_t,s0)
 /usr/lib/pgsql/test/regress/.*\.sh --	gen_context(system_u:object_r:bin_t,s0)
 /usr/lib/qt.*/bin(/.*)?			gen_context(system_u:object_r:bin_t,s0)
-/usr/lib/wicd/monitor\.py 	-- 	gen_context(system_u:object_r:bin_t, s0)
+/usr/lib/wicd/monitor\.py 	-- 	gen_context(system_u:object_r:bin_t,s0)
+/usr/lib/apt/apt-helper		--	gen_context(system_u:object_r:bin_t,s0)
+/usr/lib/apt/apt.systemd.daily	--	gen_context(system_u:object_r:bin_t,s0)
```
pebenito, line 183: This is a CI error:
```
/home/travis/build/SELinuxProject/refpolicy/policy/modules/kernel/corecommands.fc:183: unescaped dot still present /usr/lib/apt/apt.systemd.daily after being reduced to /usr/lib/apt/apt.systemd.daily (suggestion: use \. to match a dot, or a charset like [^/])
```

policy/modules/kernel/files.if
```diff
@@ -2909,6 +2909,24 @@ interface(`files_mounton_etc_dirs',`
...
+## </param>
+#
+interface(`files_watch_etc_symlinks', `
+	gen_require(`
+		type etc_t;
+	')
+
+	allow $1 etc_t:lnk_file watch;
+')
+
```
pebenito, line 2929: This should go down with the other `etc_t` symlink interfaces, around line 3220.

bauen1:
I've added more fixes for various annoyances.

policy/modules/kernel/filesystem.fc
```diff
@@ -20,6 +20,8 @@
 /sys/fs/pstore	-d	gen_context(system_u:object_r:pstore_t,s0)
 /sys/fs/pstore/.*	<<none>>
 
+/sys/kernel/tracing(/.*)?	<<none>>
```
cgzones, line 23: Maybe also add the entry `/sys/kernel/tracing                -d      gen_context(system_u:object_r:tracefs_t,s0)` ?

doverride:
cgzones <notifications@github.com> writes:

> @cgzones commented on this pull request.
>
> --------------------------------------------------------------------------------------------------------------------------------------
>
> In policy/modules/kernel/filesystem.fc:
>
>> @@ -20,6 +20,8 @@
>  /sys/fs/pstore	-d	gen_context(system_u:object_r:pstore_t,s0)
>  /sys/fs/pstore/.*	<<none>>
>  
> +/sys/kernel/tracing(/.*)?	<<none>>
>
> Maybe also add the entry /sys/kernel/tracing -d gen_context(system_u:object_r:tracefs_t,s0) ?

Then tracefs_t would have to be allowed to associate with sysfs_t
filesystems (if it is not allowed that already)

This solution is simpler because you rely on genfscon() which to me
makes sense as you cannot create files on tracefs manually anyway so
that then implies that you should never have to relabel files on there.

>
> —
>

-- 
Dominick Grift

doverride:
Dominick Grift <dominick.grift@defensec.nl> writes:

> cgzones <notifications@github.com> writes:
>
>> @cgzones commented on this pull request.
>>
>> --------------------------------------------------------------------------------------------------------------------------------------
>>
>> In policy/modules/kernel/filesystem.fc:
>>
>>> @@ -20,6 +20,8 @@
>>  /sys/fs/pstore	-d	gen_context(system_u:object_r:pstore_t,s0)
>>  /sys/fs/pstore/.*	<<none>>
>>  
>> +/sys/kernel/tracing(/.*)?	<<none>>
>>
>> Maybe also add the entry /sys/kernel/tracing -d gen_context(system_u:object_r:tracefs_t,s0) ?
>
> Then tracefs_t would have to be allowed to associate with sysfs_t
> filesystems (if it is not allowed that already)
>
> This solution is simpler because you rely on genfscon() which to me
> makes sense as you cannot create files on tracefs manually anyway so
> that then implies that you should never have to relabel files on there.
>

Not to mention that this would then make tracefs_t a mountpoint type for
tracefs_t filesystems. I don't see a reason to overcomplicate this.

>>
>> —
>> You are receiving this because you are subscribed to this thread.
>> Reply to this email directly, view it on GitHub, or unsubscribe.
>>

-- 
Dominick Grift

bauen1:
> Please respond to @doverride 's comments, as they seem to apply.

I think I already did by not implementing @cgzones suggestions ?

doverride:
bauen1 <notifications@github.com> writes:

>  Please respond to @doverride 's comments, as they seem to apply.
>
> I think I already did by not implementing @cgzones suggestions ?

Yes, I think this is the way it should be.

There can be subtile differences between the various filesystems that
support seclabel that may, or may not, warrant different treatment.

For example the pstore entries above (but also cgroup). These
filesystems allow you to actually create files on them. In that
scenario you may, or may not, want file context specs to back up any
genfscon() type labeling.

But if the filesystem does not allow you to create files on it in the
first place then I do not see any reason to add file context specs for
it.

tracefs_t type should never have to associate with any filesystem type
other than itself. (same as with the bpf entry above)

>
> —
>

-- 
Dominick Grift

pebenito:
My bad, I got confused by the mix of responses.

cgzones:
thanks @doverride, makes sense

PR #266, opened 2020-05-29
Fix autorelabel

bauen1:
The systemd autolabel generator was broken (likely by #197). This adds the required access to allow it to function correctly.
Alternativeley the generator could be targeted seperately.

This PR also adds a policy for the autorelabel service.
Since it runs `genhomedircon` it needs to transition to `semanage_t` however this access probably shouldn't be given to `initrc_t`.
It also ensures that `restorecon` is run with `SystemHigh` so even if selinux is in enforcing mode with MCS or MLS enabled everything will be relabeled properly.

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -300,6 +300,8 @@ fs_register_binary_executable_type(systemd_binfmt_t)
 # generic generator local policy
 #
 
+allow systemd_generator_type self:process setfscreate;
+
 corecmd_search_bin(systemd_generator_type)
 
 dev_read_sysfs(systemd_generator_type)
@@ -321,6 +323,8 @@ init_write_pid_files(systemd_generator_type)
 kernel_use_fds(systemd_generator_type)
 kernel_read_system_state(systemd_generator_type)
 kernel_read_kernel_sysctls(systemd_generator_type)
+
+seutil_libselinux_linked(systemd_generator_type)
 
 #######################################
 #
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -388,6 +388,9 @@ optional_policy(`
 corecmd_exec_bin(systemd_selinux_autorelabel_generator_t)
 corecmd_exec_shell(systemd_selinux_autorelabel_generator_t)
 
+# for /.autorelabel
+files_read_etc_runtime_files(systemd_selinux_autorelabel_generator_t)
+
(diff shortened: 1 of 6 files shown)


policy/modules/system/systemd.te
```diff
@@ -300,13 +312,18 @@ fs_register_binary_executable_type(systemd_binfmt_t)
 # generic generator local policy
 #
 
-corecmd_search_bin(systemd_generator_type)
+allow systemd_generator_type self:process setfscreate;
+
+corecmd_exec_bin(systemd_generator_type)
+corecmd_exec_shell(systemd_generator_type)
```
pebenito, line 318: I don't think this should be applied to all generators.

policy/modules/system/systemd.te
```diff
@@ -1137,6 +1158,42 @@ optional_policy(`
...
+init_read_state(systemd_selinux_autorelabel_t)
+
+miscfiles_read_localization(systemd_selinux_autorelabel_t)
+
+seutil_read_config(systemd_selinux_autorelabel_t)
+seutil_domtrans_setfiles(systemd_selinux_autorelabel_t)
+# for genhomedircon
+seutil_domtrans_semanage(systemd_selinux_autorelabel_t)
+
+quota_domtrans(systemd_selinux_autorelabel_t)
```
pebenito, line 1195: This should be optional unless it absolutely can't work without this.

bauen1: I'll have to test if there's a way to work around this, the script invokes `[ -x "/sbin/quotaoff" ] && /sbin/quotaoff -aug`.
Same with genhomdircon but it might be possible to replace it with read/write access to /etc/selinux.

pebenito: If it's only running quotaoff/on only if the tool is executable, I'd think this would be fine as optional. If someone has a policy with no quota module but quotas enable somehow, there will be more problems than this tool.

bauen1: Oh I misunderstood you.

policy/modules/system/systemd.te
bauen1: After the discussion on the mailing list, should I revert this to its original form ?

pebenito:
Thanks for this. Since the generator domains will be collapsed following discussions on the list, this PR will need to be rebased afterwards. Please open a new PR for this at that time.

Issue #246, opened 2020-05-01
Add support for userdb


bigon:
Hello,

With the recent versions of systemd, there is a new `userdb` component added.

`libnss-systemd` is now trying to connect to a socket located in `/run/systemd/userdb/` that meas that any domain (including a user one) that should resolve user/group id might try to connect to it.

There is also an optional daemon running

Fedora policy already has support for this that adds custom types.

bauen1:
> There is also an optional daemon running

Which daemon ?

doverride:
Laurent Bigonville <notifications@github.com> writes:

> Hello,
>
> With the recent versions of systemd, there is a new userdb component added.
>
> libnss-systemd is now trying to connect to a socket located in /run/systemd/userdb/ that meas that any domain (including a user one)
> that should resolve user/group id might try to connect to it.
>
> There is also an optional daemon running
>
> Fedora policy already has support for this that adds custom types.

There are currently 3 sockets and 1 symlink in /run/user/userdb if you have
systemd-homed enabled:

io..systemd.DynamicUser - This is the new way to resolve dynamic users
previously this was done by dbus chatting with pid1 (init_t)

not sure if suppose for that was ever implemented in refpolicy. NSS
passwd/group clients need to be able to list /run/systemd/userdb and write
this socket to unix stream connect to init_t to be able to resolve
dynamic uids/gids (along with needing to get attributes of
/etc/systemd/dont-synthesize-nobody)

So since in refpolicy there is no distinction between glibc nss
passwd/group and glibc nss hosts (and all the others like services),
callers of auth_use_nsswitch() should be allowed to resolve dynamic
users this way (dbus is no longer used for this which is nice since now
you dont need to be dbus client anymore just for glibc passwd/group (you
still need it for glibc nss hosts though (for mymachines and myhostname) 

Then there is io.systemd.Home, this is new for systemd-homed and
pam-systemd-home, nss passwd group clients (auth_use_nswitch() callers
)should also be able to list /run/systemd/userdb and write this
io.systemd.Home socket to connect to systemd-homed (if its enabled and
running) to resolve users managed with homed (i guess)

lastly theres io.systemd.Multiplexer (and a symlink
io.systemd.NameServiceSwitch pointing to it). As far as i know with the
exception of userdbctl (userdbctl uses it to connect to
systemd-userdbd), only "pam clients" need to be able list
/run/systemd/userdb AFAIK, read the io.systemd.NameServiceSwitch symlink and
[...]

PR #267, opened 2020-05-30
systemd-sysusers: add policy

bauen1:
Running a system without the unconfined module, this is another service that needs additional privileges (modifying /etc/shadow and friends).
This is blocked by #246 but open for comments.

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -1126,3 +1126,48 @@ interface(`systemd_write_all_user_keys',`
 
 	allow $1 systemd_user_session_type:key write;
 ')
+
+########################################
+## <summary>
+##  Execute systemd-sysusers in the
+##  systemd sysusers domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_domtrans_sysusers', `
+	gen_require(`
+		type systemd_sysusers_t, systemd_sysusers_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, systemd_sysusers_exec_t, systemd_sysusers_t)
+')
+
+########################################
+## <summary>
+##  Run systemd-sysusers with a domain transition.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="role">
+##  <summary>
+##  Role allowed access.
+##  </summary>
+## </param>
+## <rolecap/>
+#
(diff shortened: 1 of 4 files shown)


policy/modules/system/systemd.te
```diff
@@ -1158,6 +1165,36 @@ seutil_read_file_contexts(systemd_sessions_t)
...
+#########################################
+#
+# Sysusers local policy
+#
+
+allow systemd_sysusers_t self:capability { chown fsetid };
+allow systemd_sysusers_t self:process setfscreate;
+allow systemd_sysusers_t self:unix_dgram_socket { create_socket_perms sendto };
+
+dev_write_kmsg(systemd_sysusers_t)
```
pebenito, line 1178: Please see `systemd_log_parse_environment()` which will handle several of the perms here.

policy/modules/system/systemd.if
```diff
@@ -1126,3 +1126,46 @@ interface(`systemd_write_all_user_keys',`
...
+## </param>
+interface(`systemd_domtrans_systemd_sysusers', `
+	gen_require(`
+		type systemd_sysusers_t, systemd_sysusers_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, systemd_sysusers_exec_t, systemd_sysusers_t)
+')
+########################################
```
pebenito, line 1148: Nit: missing whitespace

policy/modules/system/systemd.if
```diff
@@ -1126,3 +1126,48 @@ interface(`systemd_write_all_user_keys',`
...
+##  Execute systemd-sysusers in the
+##  systemd sysusers domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_domtrans_systemd_sysusers', `
```
pebenito, line 1141: I think we should drop the extra "systemd" in the interface names here and below.

policy/modules/system/systemd.te
```diff
@@ -1158,6 +1165,34 @@ seutil_read_file_contexts(systemd_sessions_t)
...
+#########################################
+#
+# Sysusers local policy
+#
+
+allow systemd_sysusers_t self:capability { chown fsetid };
+allow systemd_sysusers_t self:process setfscreate;
+allow systemd_sysusers_t self:unix_dgram_socket sendto;
+
+dev_write_kmsg(systemd_sysusers_t)
```
pebenito, line 1178: This should be handled by `systemd_log_parse_environment()`

bauen1: I may have messed up git, I'll fix this.

PR #268, opened 2020-06-01
add rule for the management socket file

McSim85:
OpenVPN allows using unix socket files for the management interface instead of TCP ports. This is important in servers where non-admin users are also allowed to SSH in, because limits their access to the management interface.

Example directive:
management /run/openvpn/server.sock unix
management-client-user root
management-client-group root

However, here are no SELinux rule in the current version of code, that allows the creation of such socket file.
So, it was denied during start\stop of the service and logged as messages:
----------------
type=AVC msg=audit(1591045213.430:2109): avc:  denied  { unlink } for  pid=8880 comm="openvpn" name="server.sock" dev="tmpfs" ino=13559 scontext=system_u:system_r:openvpn_t:s0 tcontext=system_u:object_r:openvpn_var_run_t:s0 tclass=sock_file permissive=1
type=AVC msg=audit(1591045213.434:2110): avc:  denied  { create } for  pid=8880 comm="openvpn" name="server.sock" scontext=system_u:system_r:openvpn_t:s0 tcontext=system_u:object_r:openvpn_var_run_t:s0 tclass=sock_file permissive=1
----------------

So I had to create one manually every time when I deploy my OpenVPN. Like that:

```
$ cat ovpn_sock.te

module ovpn_sock 1.0;

require {
        type openvpn_var_run_t;
        type openvpn_t;
        class sock_file { create unlink };
}

#============= openvpn_t ==============
allow openvpn_t openvpn_var_run_t:sock_file { create unlink };
```

I wonder if it would be possible to integrate this into the code.


policy/modules/services/openvpn.te
```diff
@@ -1,4 +1,4 @@
-policy_module(openvpn, 1.18.0)
+policy_module(openvpn, 1.18.1)
```
bauen1, line 1: Don't increment the module version, this will be done when the PR is merged. (If multiple people create a pr for the same module it could end up causing merge conflicts)

policy/modules/services/openvpn.te
```diff
@@ -67,6 +67,8 @@ allow openvpn_t openvpn_etc_t:dir list_dir_perms;
 allow openvpn_t openvpn_etc_t:file read_file_perms;
 allow openvpn_t openvpn_etc_t:lnk_file read_lnk_file_perms;
 
+allow openvpn_t openvpn_runtime_t:sock_file { create unlink };
```
bauen1, line 70: The `create` and `unlink` permissions should probably be replaced by `manage_sock_file_perms`.
This could also be moved down to the other `manage_..._pattern` calls and be replaced by `manage_sock_files_pattern(openvpn_t, openvpn_runtime_t, openvpn_runtime_t)`
Since this states the intent more clearly: Managing a socket file that is labeled `openvpn_runtime_t` and is inside a directory labeled `openvpn_runtime_t`

McSim85:
Hi @bauen1 
Thank you for your comments.
I've squashed my commits.
As a result commit, I made `manage_sock_files_pattern(openvpn_t, openvpn_runtime_t, openvpn_runtime_t)` as you suggested.
Also, I've analyzed other git like [Fedora ](https://github.com/fedora-selinux/selinux-policy-contrib/blob/b35f071eace9e06117f78cdda3dd6692388dff6f/openvpn.te#L103)and suggested additional string `files_pid_filetrans(openvpn_t, openvpn_runtime_t, { sock_file file dir })`.

PR #274, opened 2020-06-14
Remove the ada module, it is unecessary and not touched since ~2008

bauen1:
It is only used to allow the compiler execmem / execstack but we have
unconfined_execmem_t for that.

Reduces the resulting policy binary by 2990 bytes.
Removing a module might require some cooperation with the package maintainers of the various distributions.


policy/modules/apps/ada.te
```diff
@@ -1,27 +0,0 @@
...
-########################################
-#
-# Declarations
-#
-
-attribute_role ada_roles;
-roleattribute system_r ada_roles;
-
-type ada_t;
-type ada_exec_t;
```
pebenito, line 12: There should be aliases added for these types.

PR #276, opened 2020-06-15
systemd: Merge generator domains.

pebenito:
If these processes are compromised they can write units to do malicious
actions, so trying to tightly protect the resources for each generator
is not effective.

Made the fstools_exec() optional, although it is unlikely that a system
would not have the module.

Only aliases for removed types in previous releases are added.  The
systemd_unit_generator() interface and systemd_generator_type attribute
were not released and are dropped without deprecation.

Signed-off-by: Chris PeBenito <pebenito@ieee.org>


dsugar100:
Is there a way we can leave the systemd_generator_type attribute?
I have written a generator that needs permissions not granted by the base.  This change would break generators that are not part of the base OS but added by others for special use cases.

pebenito:
Can you give an example? My initial reaction would be no, since a generator is just a `init_system_domain()`.

dsugar100:
Ok, looking back at the policy I have, I agree with you.  There is no need to keep the attribute.  I have updated my policy in preparation for this change.

PR #277, opened 2020-06-17
Resolve neverallow failure introduced in #273

dsugar100:


pebenito:
I think this is probably too broad since the main thing is we can't get into a state where kcore is not readable/writeable, so I feel like relaxing the neverallow may be the better choice, e.g. `~{ getattr mounton }`.

dsugar100:
Yes, that is a much simpler solution.

PR #279, opened 2020-07-01
Systemd generator

bigon:
Several fixes for the systemd generator

I expanded `init_getattr_all_script_files` interface to allow getting the attributes of the systemd unit files in addition of the initscripts, please tell me if you want a new one

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -380,6 +380,7 @@ files_list_usr(systemd_generator_t)
 fs_list_efivars(systemd_generator_t)
 fs_getattr_xattr_fs(systemd_generator_t)
 
+init_getattr_all_script_files(systemd_generator_t)
 init_create_pid_files(systemd_generator_t)
 init_manage_pid_dirs(systemd_generator_t)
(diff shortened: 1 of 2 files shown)


bigon:
Or maybe we should create a different label per generator?

There are other generator like `systemd-cryptsetup-generator` that seems to require `setfscreate`

policy/modules/system/systemd.te
```diff
@@ -379,6 +380,7 @@ files_list_usr(systemd_generator_t)
 fs_list_efivars(systemd_generator_t)
 fs_getattr_xattr_fs(systemd_generator_t)
 
+init_getattr_all_script_files(systemd_generator_t)
```
bauen1, line 383: `systemd-sysv-generator` will also _read_ all script files to determine the "init options" (https://github.com/SELinuxProject/refpolicy/commit/e63ad81ced3f8be92bb18b9d5f84388a04185819)

pebenito:

bigon:
Not sure, this should maybe be abandoned in favor of @bauen1 work

PR #284, opened 2020-07-07
openvpn: more versatile file context regex for ipp.txt

alexminder:


policy/modules/services/openvpn.fc
```diff
@@ -1,5 +1,5 @@
 /etc/openvpn(/.*)?	gen_context(system_u:object_r:openvpn_etc_t,s0)
-/etc/openvpn/ipp\.txt	--	gen_context(system_u:object_r:openvpn_etc_rw_t,s0)
+/etc/openvpn/([^/]*/)?ipp\.txt	--	gen_context(system_u:object_r:openvpn_etc_rw_t,s0)
```
pebenito, line 2: Since you have the `?` I think it should be `/etc/openvpn/([^/]+/)?ipp\.txt`.

PR #286, opened 2020-07-22
Also label polkit-agent-helper-1 when installed directly in /usr/libexec

bigon:
Debian now installs that executable directly in /usr/libexec for the
version 0.105

Signed-off-by: Laurent Bigonville <bigon@bigon.be>

### DIFF
--- a/policy/modules/services/policykit.fc
+++ b/policy/modules/services/policykit.fc
@@ -11,6 +11,7 @@
 # Systemd unit file
 /usr/lib/systemd/system/[^/]*polkit.*	--	gen_context(system_u:object_r:policykit_unit_t,s0)
 
+/usr/libexec/polkit-agent-helper-1	--	gen_context(system_u:object_r:policykit_auth_exec_t,s0)
 /usr/libexec/polkit-read-auth-helper	--	gen_context(system_u:object_r:policykit_auth_exec_t,s0)
 /usr/libexec/polkit-grant-helper.*	--	gen_context(system_u:object_r:policykit_grant_exec_t,s0)
 /usr/libexec/polkit-resolve-exe-helper.*	--	gen_context(system_u:object_r:policykit_resolve_exec_t,s0)


policy/modules/services/policykit.fc
```diff
@@ -11,6 +11,7 @@
 # Systemd unit file
 /usr/lib/systemd/system/[^/]*polkit.*	--	gen_context(system_u:object_r:policykit_unit_t,s0)
 
+/usr/libexec/polkit-agent-helper-1	--	gen_context(system_u:object_r:policykit_auth_exec_t,s0)
```
pebenito, line 14: Perhaps the version should be handled with `.*` instead, like the other lines? I'm unsure since there seems to be a bunch of variance between distros.

bigon: The fc here look a bit a mess.

On my debian unstable machine, the only file installed are `/usr/libexec/polkit-agent-helper-1` and `/usr/libexec/polkitd` (for version 0.105). The version in debian experimental (0.116) is still installing the same files in `/usr/lib/polkit-1`

I'm not sure if the other file contexts are still used anywhere by any distro

I've opened a [bug upstream](https://gitlab.freedesktop.org/polkit/polkit/-/issues/125) to officially move them to `/usr/libexec/polkit-1` so maybe we could wait until that bug is fixed upstream and carry the patch in debian only

pebenito: > I've opened a [bug upstream](https://gitlab.freedesktop.org/polkit/polkit/-/issues/125) to officially move them to `/usr/libexec/polkit-1` so maybe we could wait until that bug is fixed upstream and carry the patch in debian only

Sounds like a good plan.

pebenito:

bigon:
This is blocked by https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/63 I guess

pebenito:
Added the question label since the github action for marking stale issues/PRs has a bug handling labels with spaces in it. The external dependency label was supposed to block the stale action.

pebenito:
@bigon the upstream PR hasn't made any progress for 6 months, so I'm closing this PR. If it is finally resolved, please do resubmit a new PR.

PR #288, opened 2020-07-28
init: Revise init_startstop_service() build option blocks.

pebenito:
Revise to use ifelse to have a clear set of criteria for enabling the
various options.  Additionally, if no options are enabled, run_init
permissions are provided as a default.


pebenito:
@perfinion @bigon please review

PR #289, opened 2020-07-28
kernel: Drop unlabeled_t as a files_mountpoint().

pebenito:
This made unlabeled_t a file and provided much more access than an
unlabeled file should have.  Access to unlabeled objects should be
explicit.


dsugar100:
I'm now seeing a denial where aide_t can no longer read unlabeled_t.  I think it is due to this commit.
I'm wondering if this is intentional?
Not much has write access to unlabeled_t, but maybe allowing AIDE to read unlableled_t would be reasonable.

Thoughts?

pebenito:
I don't see why you would want AIDE to hash an unlabeled file.

dsugar100:
Only in the case the unlabeled_t file was changed.  That is why I'm mixed.  Based on the few domains that can read/write to unlabeled_t I feel the possibility of some malicious process being able to write to unlabeled_t is low, at the same time having AIDE be able to read/hash/validate to verify that some unlabeled_t file hasn't changed might have some small benefit if only in the added assurance there is no unexpected change maybe while the system was off/rescued/other odd case.

PR #291, opened 2020-08-10
Remove duplicated rules

yizhao1:


policy/modules/system/logging.te
```diff
@@ -605,8 +605,6 @@ optional_policy(`
 
 optional_policy(`
 	udev_read_db(syslogd_t)
-	# for systemd-journal to read seat data from /run/udev/data
-	udev_read_runtime_files(syslogd_t)
```
pebenito, line 609: This isn't duplicate see discussion in #230

policy/modules/services/xserver.te
```diff
@@ -883,7 +883,7 @@ optional_policy(`
 ')
 
 optional_policy(`
-	resmgr_stream_connect(xdm_t)
+	resmgr_stream_connect(xserver_t)
```
pebenito, line 886: I'm ok with this being in the PR, but it should be a separate commit since it is a correction.

policy/modules/system/logging.te
yizhao1: If this is only for systemd-journal, there is alaredy such line (line 557) in init_systemd conditon block.

policy/modules/services/xserver.te
pebenito: I think this is still relevant.

PR #295, opened 2020-08-13
travis-ci: add interface db build

cgzones:
Check that sepolgen-ifgen (creates the database for audit2allow -R) works on the policy.

Note: This needs a manual travis-ci cache clean-up, as sepolgen-ifgen from the SELinux sub-directory python is required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@
 language: python
 python: 3.5
 
-matrix:
-  fast_finish: true
-
 env:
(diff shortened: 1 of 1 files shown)


pebenito:
I deleted all of the caches on Travis but it still fails.

cgzones:
I am not sure any more this is worth the effort:
For the sepolgen-ifgen python script the SELinux python wrappers are required and for them to build libpython3.5-dev needs to be installed (cause we running in a python3.5 environment). But libpython3.5-dev is not available in bionic.

pebenito:
I am open to this change if the CI situation improves.

PR #299, opened 2020-08-23
acpi.te: Allow acpid_t to shutdown the system

### LINKED ISSUE #298: acpid_t blocks VM shutdown
I have a Debian 10 libvirt/KVM host with a Debian 10 VM guest, if I run:

```
sudo virsh shutdown guest
```

...the guest does not shutdown. If I disable `dontaudit`'s, I see this within the guest's logs:

```
type=AVC msg=audit(1598187082.086:163): avc:  denied  { getattr } for  pid=583 comm="powerbtn-acpi-s" path="/usr/share/acpi-support/policy-funcs" dev="sda1" ino=2888143 scontext=system_u:system_r:acpid_t:s0 tcontext=system_u:object_r:usr_t:s0 tclass=file permissive=0
```

Policy version on Debian/stable is:

```
ii  selinux-policy-default        2:2.20190201-2               all          Strict and Targeted variants of the SELinux policy
```

jpds:
acpi.te: Allow acpid_t to shutdown the system - this is required to handle shutdown calls from libvirt. Fixes #298.

### DIFF
--- a/policy/modules/services/acpi.te
+++ b/policy/modules/services/acpi.te
@@ -137,7 +137,6 @@ auth_use_nsswitch(acpid_t)
 
 init_domtrans_script(acpid_t)
 init_telinit(acpid_t)
-init_write_initctl(acpid_t)
 
 libs_exec_ld_so(acpid_t)
 libs_exec_lib_files(acpid_t)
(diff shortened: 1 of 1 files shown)


policy/modules/services/acpi.te
```diff
@@ -135,6 +136,8 @@ domain_dontaudit_list_all_domains_state(acpid_t)
 auth_use_nsswitch(acpid_t)
 
 init_domtrans_script(acpid_t)
+init_telinit(acpid_t)
+init_write_initctl(acpid_t)
```
pebenito, line 140: `init_write_initctl()` shouldn't be necessary if you have `init_telinit()`.

jpds: Removed.

PR #301, opened 2020-08-26
selint: fix S-010

bauen1:


cgzones:
The second commit conflicts with https://github.com/SELinuxProject/refpolicy/commit/f8f87a80850e39378eccdcbca54788cb47aec7e9

policy/modules/kernel/devices.te
```diff
@@ -385,6 +385,6 @@ files_associate_tmp(device_node)
 #
 
 allow devices_unconfined_type self:capability sys_rawio;
-allow devices_unconfined_type device_node:blk_file { manage_blk_file_perms relabelfrom relabelto map execute quotaon mounton audit_access execmod watch };
-allow devices_unconfined_type device_node:chr_file { manage_chr_file_perms relabelfrom relabelto map execute quotaon mounton execmod audit_access watch };
```
cgzones, line 389: Seems you dropped the permission `map`, its quite uncommon, but e.g. `dev_read_sound()` grants it

policy/modules/services/afs.if
```diff
@@ -54,7 +54,7 @@ interface(`afs_rw_cache',`
 	')
 
 	files_search_var($1)
-	allow $1 afs_cache_t:file { read write };
```
cgzones, line 57: Here `open` gets added, but the interface name does not mention inherited access

cgzones (commented):
Out of curiosity: did you encounter any incorrect/incomplete/missing/odd tags by SELint, anything to improve?

policy/modules/services/afs.if
pebenito: I suspect in this case adding `open` is correct. My guess is that afs isn't used much so no one has actually hit this.

policy/modules/kernel/devices.te
pebenito: Agreed.

policy/modules/services/afs.if
bauen1: The `open` access is expected, otherwise the interface should be named `afs_rw_inherited_cache_files` or similiar

policy/modules/kernel/devices.te
bauen1: Very good catch, fixed.

bauen1:
> Out of curiosity: did you encounter any incorrect/incomplete/missing/odd tags by SELint, anything to improve?

Not yet, but I also haven't used it much.

> The second commit conflicts with f8f87a8

I've dropped the conflicting commit.

cgzones (approved):

PR #294, opened 2020-08-13
Add SELint to CI

cgzones:

PR #302, opened 2020-08-27
Looks like this got dropped in pull request #294

dsugar100:
Adding map permission back, seeing the following denial:
localhost kernel: type=1400 audit(1598497795.109:57): avc:  denied  { map } for  pid=1054 comm="modprobe" path="/usr/lib/modules/3.10.0-1127.19.1.el7.x86_64/modules.dep.bin" dev="dm-0" ino=23711 scontext=system_u:system_r:kmod_t:s0 tcontext=system_u:object_r:modules_dep_t:s0 tclass=file permissive=0

### DIFF
--- a/policy/modules/system/modutils.te
+++ b/policy/modules/system/modutils.te
@@ -44,6 +44,7 @@ allow kmod_t self:rawip_socket create_socket_perms;
 # Read module config and dependency information
 list_dirs_pattern(kmod_t, modules_conf_t, modules_conf_t)
 read_files_pattern(kmod_t, modules_conf_t, modules_conf_t)
+allow kmod_t modules_dep_t:file map;
 list_dirs_pattern(kmod_t, modules_dep_t, modules_dep_t)
 manage_files_pattern(kmod_t, modules_dep_t, modules_dep_t)
 files_kernel_modules_filetrans(kmod_t, modules_dep_t, file)


cgzones (approved):
Indeed

policy/modules/system/modutils.te
```diff
@@ -44,6 +44,7 @@ allow kmod_t self:rawip_socket create_socket_perms;
 # Read module config and dependency information
 list_dirs_pattern(kmod_t, modules_conf_t, modules_conf_t)
 read_files_pattern(kmod_t, modules_conf_t, modules_conf_t)
+allow kmod_t modules_dep_t:file map;
```
pebenito, line 47: Please change line 46 to `mmap_read_files_pattern` instead.

cgzones: line 46 is about `modules_conf_t`.
BTW: is there any reason why the permission `map` is not included in `manage_file_perms`?

pebenito: My misake. Map is specifically controlled because it is not revocable.

bauen1: > Map is specifically controlled because it is not revocable.

Somewhat unrelated to this pull-request, as far as I know `map` isn't a special permission at all.
Are you referring to changes to the label of the file while it is opened, where `read` are checked again against the new label, but access to mapped areas is not ?

pebenito: > > Map is specifically controlled because it is not revocable.
> 
> Somewhat unrelated to this pull-request, as far as I know `map` isn't a special permission at all.
> Are you referring to changes to the label of the file while it is opened, where `read` are checked again against the new label, but access to mapped areas is not ?

Yes.

PR #304, opened 2020-08-30
init.te: Made init_run_all_scripts_domain unconditional

jpds:
...so it works on non-systemd systems like Gentoo.

Signed-off-by: Jonathan Davies <jpds@protonmail.com>

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -228,10 +228,10 @@ seutil_read_default_contexts(init_t)
 
 miscfiles_read_localization(init_t)
 
-ifdef(`init_systemd',`
-	# handle instances where an old labeled init script is encountered.
-	typeattribute init_t init_run_all_scripts_domain;
+# handle instances where an old labeled init script is encountered.
+typeattribute init_t init_run_all_scripts_domain;
 
+ifdef(`init_systemd',`
 	allow init_t self:unix_dgram_socket { create_socket_perms sendto };
 	allow init_t self:process { setsockcreate setfscreate setrlimit };
 	allow init_t self:process { getcap setcap getsched setsched };


doverride:
Jonathan Davies <notifications@github.com> writes:

> ...so it works on non-systemd systems like Gentoo.

Even though I think this is wrong, I do not know how this supposed to
work.

With this rule init_t will transition to initrc_t on all
init_script_file_type by default. So one will lose the option to transition to private
script domains.

The problem here is that refpolicy has a private script type for audit:
auditd_initrc_exec_t, and that init_t currently runs that script without a
domain transition, leaving auditd running as init_t

Is one supposed to explicitly add transitions from init_t on labeled initrc
scripts to initrc_t?

Why does init_t run init scripts with private init script types (line
auditd_initrc_exec_t) without a transition by default?

(i guess because it cannot know whether one wants to run it with a
transition to initrc_t or whether one wants to run it with a transition
to some private script domain like for example auditd_initrc_t....

>
> Signed-off-by: Jonathan Davies jpds@protonmail.com
>
> -----------------------------------------------------------------------------------------------------------------------------------------------
>
>
>   https://github.com/SELinuxProject/refpolicy/pull/304
>
> Commit Summary
>
> * init.te: Made init_run_all_scripts_domain unconditional so it works on non-systemd systems like Gentoo.
>
> File Changes
>
> * M policy/modules/system/init.te (6) 
>
> Patch Links:
>
> * https://github.com/SELinuxProject/refpolicy/pull/304.patch
> * https://github.com/SELinuxProject/refpolicy/pull/304.diff
>
> —
>

-- 
Dominick Grift

pebenito:
Since you had another Gentoo-specific change can you do this through Gentoo too? At this point I think Gentoo is the only non-systemd distro that we're supporting upstream.

PR #305, opened 2020-08-30
corecommands.fc: Handle Gentoo python-exec binary directories.

jpds:
Gentoo stores Python scripts in separate directories: https://wiki.gentoo.org/wiki/Project:Python/python-exec

Signed-off-by: Jonathan Davies <jpds@protonmail.com>


pebenito:
My preference would be for this to go through the Gentoo tree and upstreamed by them.

PR #269, opened 2020-06-04
Systemd userdb

bauen1:
This only addresses the scenario where `userdb` is disabeled, e.g. debian.
Partially Fixes #246 

`init_unix_stream_socket_connectto` is used because `init_stream_connect` was already taken and grants additional (unwanted) access that is not necessary.

PR #307, opened 2020-09-04
Set of fixes made while using a minimal Buildroot generated system

atenart:
Hello,

While working on adding SELinux support in [Buildroot](https://buildroot.org) (a tool widely used to generate small or embedded Linux systems), I came across multiple denied actions when using the refpolicy. I made tests using a small image, with systemd as the init and was not able to boot properly and to go up to the login prompt. As a result, I patched the refpolicy and (most of) those changes ended up in this PR.

The reason I made this PR is I'd like more systems to be supported by the refpolicy, and this PR will at least help to start the discussion on how to do that properly. Many of the changes that were required for the Buildroot generated image seem to also be part of the [changes](https://git.yoctoproject.org/cgit/cgit.cgi/meta-selinux/tree/recipes-security/refpolicy/refpolicy) made in [Yocto/OE](https://www.yoctoproject.org/) for their images. I believe there's a real benefit to support those systems in the official refpolicy (note that Yocto is carrying 80 patches on top of the refpolicy!).

I tried to use available functions whenever possible, but might have missed some. I also chose to make all of the changes unconditional so far, but that might be something to explore as I'm not sure wether or not all of those changes should be made for everyone. That's a point where I'd like some input, as I assume the refpolicy is at least working already for some.

Note that systems generated by Buildroot or Yocto/OE aren't distributions per-say, and can't be considered as ones as those systems could end up to be quite different. Using the DISTRO mechanism do not seem to fit here. We could think of other conditional mechanisms if needed.

Thanks!

### DIFF
--- a/policy/modules/services/dbus.if
+++ b/policy/modules/services/dbus.if
@@ -143,6 +143,8 @@ interface(`dbus_system_bus_client',`
 	stream_connect_pattern($1, system_dbusd_runtime_t, system_dbusd_runtime_t, system_dbusd_t)
 
 	dbus_read_config($1)
+	dbus_list_system_bus_runtime($1)
+	dbus_read_system_bus_runtime_named_sockets($1)
 ')
 
 #######################################
--- a/policy/modules/services/dbus.if
+++ b/policy/modules/services/dbus.if
@@ -596,6 +596,24 @@ interface(`dbus_watch_system_bus_runtime_dirs',`
 
 ########################################
 ## <summary>
+##	List system bus runtime directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dbus_list_system_bus_runtime',`
+	gen_require(`
+		type system_dbusd_runtime_t;
+	')
+
+	allow $1 system_dbusd_runtime_t:dir list_dir_perms;
+')
+
+########################################
+## <summary>
 ##	Watch system bus runtime named sockets.
 ## </summary>
 ## <param name="domain">
@@ -610,6 +628,24 @@ interface(`dbus_watch_system_bus_runtime_named_sockets',`
 	')
 
 	allow $1 system_dbusd_runtime_t:sock_file watch;
+')
+
+########################################
+## <summary>
+##	Read system bus runtime named sockets.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dbus_read_system_bus_runtime_named_sockets',`
+	gen_require(`
+		type system_dbusd_runtime_t;
+	')
+
+	allow $1 system_dbusd_runtime_t:sock_file read;
 ')
 
 ########################################
--- a/policy/modules/system/locallogin.te
+++ b/policy/modules/system/locallogin.te
@@ -59,6 +59,7 @@ kernel_read_system_state(local_login_t)
 kernel_read_kernel_sysctls(local_login_t)
 kernel_search_key(local_login_t)
 kernel_link_key(local_login_t)
+kernel_getattr_proc(local_login_t)
 
 corecmd_list_bin(local_login_t)
 # cjp: these are probably not needed:
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -524,7 +524,7 @@ ifdef(`init_systemd',`
 	allow syslogd_t self:netlink_audit_socket connected_socket_perms;
 	allow syslogd_t self:capability2 audit_read;
 	allow syslogd_t self:capability { chown setgid setuid sys_ptrace };
-	allow syslogd_t self:netlink_audit_socket { getattr getopt read setopt write };
+	allow syslogd_t self:netlink_audit_socket { getattr getopt read setopt write nlmsg_write };
 
 	# remove /run/log/journal when switching to permanent storage
 	allow syslogd_t var_log_t:dir rmdir;
--- a/policy/modules/system/sysnetwork.if
+++ b/policy/modules/system/sysnetwork.if
@@ -346,6 +346,8 @@ interface(`sysnet_read_config',`
 	')
 
 	files_search_etc($1)
+	files_search_runtime($1)
+	allow $1 net_conf_t:dir list_dir_perms;
 	allow $1 net_conf_t:file read_file_perms;
 
 	ifdef(`distro_debian',`
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -461,6 +461,7 @@ selinux_get_fs_mount(systemd_hw_t)
 selinux_use_status_page(systemd_hw_t)
 
 init_read_state(systemd_hw_t)
+init_search_runtime(systemd_hw_t)
 
 seutil_read_config(systemd_hw_t)
 seutil_read_file_contexts(systemd_hw_t)
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -785,6 +785,7 @@ dev_write_kmsg(systemd_networkd_t)
 files_read_etc_files(systemd_networkd_t)
 files_watch_runtime_dirs(systemd_networkd_t)
 files_watch_root_dirs(systemd_networkd_t)
+fs_getattr_xattr_fs(systemd_networkd_t)
 
 auth_use_nsswitch(systemd_networkd_t)
 
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1095,6 +1095,7 @@ auth_use_nsswitch(systemd_resolved_t)
 
 files_watch_root_dirs(systemd_resolved_t)
 files_watch_runtime_dirs(systemd_resolved_t)
+files_list_runtime(systemd_resolved_t)
 
 init_dgram_send(systemd_resolved_t)
 
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -362,6 +362,8 @@ seutil_search_default_contexts(systemd_coredump_t)
 #
 
 allow systemd_generator_t self:fifo_file rw_fifo_file_perms;
+allow systemd_generator_t self:capability dac_override;
+allow systemd_generator_t self:process setfscreate;
 
 corecmd_getattr_bin_files(systemd_generator_t)
 
--- a/policy/modules/system/udev.te
+++ b/policy/modules/system/udev.te
@@ -421,3 +421,5 @@ kernel_read_kernel_sysctls(udevadm_t)
 kernel_read_system_state(udevadm_t)
 
 seutil_read_file_contexts(udevadm_t)
+
+fs_getattr_xattr_fs(udevadm_t)
--- a/policy/modules/kernel/corecommands.fc
+++ b/policy/modules/kernel/corecommands.fc
@@ -155,6 +155,7 @@ ifdef(`distro_gentoo',`
 /usr/bin/sesh			--	gen_context(system_u:object_r:shell_exec_t,s0)
 /usr/bin/scponly		--	gen_context(system_u:object_r:shell_exec_t,s0)
 /usr/bin/scponlyc		--	gen_context(system_u:object_r:shell_exec_t,s0)
+/usr/bin/sh			--	gen_context(system_u:object_r:shell_exec_t,s0)
 /usr/bin/smrsh			--	gen_context(system_u:object_r:shell_exec_t,s0)
 /usr/bin/tcsh			--	gen_context(system_u:object_r:shell_exec_t,s0)
 /usr/bin/yash			--	gen_context(system_u:object_r:shell_exec_t,s0)


policy/modules/system/udev.te
```diff
@@ -170,6 +170,7 @@ init_start_all_units(udev_t)
 init_stop_all_units(udev_t)
 # for hdparm init script run by udev
 init_get_script_status(udev_t)
+init_list_runtime(udev_t)
```
doverride, line 173: This should probably be part of the existing auth_use_nsswitch() call below. There is a PR that addresses that:
https://github.com/SELinuxProject/refpolicy/pull/269

policy/modules/services/dbus.te
```diff
@@ -172,6 +172,8 @@ ifdef(`init_systemd', `
 
 	# Recent versions of dbus are started as Type=notify
 	init_write_runtime_socket(system_dbusd_t)
+
+	allow system_dbusd_t init_t:unix_stream_socket connectto;
```
doverride, line 176: Idem ditto: should (probably) be addressed as part of auth_use_nsswitch()
https://github.com/SELinuxProject/refpolicy/pull/269

atenart: Good to know, that can be dropped from the PR then. Thanks!

policy/modules/system/locallogin.te
```diff
@@ -123,6 +123,7 @@ auth_manage_pam_runtime_dirs(local_login_t)
 auth_manage_pam_runtime_files(local_login_t)
 auth_manage_pam_console_data(local_login_t)
 auth_domtrans_pam_console(local_login_t)
+auth_use_pam_motd_dynamic(local_login_t)
```
doverride, line 127: If this is a Debian-based image then you should probably build with ifdef distro_debian, which would allow this access
But even if you want to allow this unconditionally, then you should remove the distro_debian conditional as that then is redundant

atenart: This is not a Debian based image, but an image generated by using Buildroot (see the PR description). My aim is to discuss how to handle such changes properly if they do not apply to all systems supported by the refpolicy.

I have one question though: do you know why so far Debian-based distributions alone need such a rule, and others do not?

policy/modules/system/locallogin.te
```diff
@@ -138,6 +138,9 @@ userdom_use_unpriv_users_fds(local_login_t)
 userdom_sigchld_all_users(local_login_t)
 userdom_create_all_users_keys(local_login_t)
 
+allow local_login_t var_log_t:file { rw_file_perms create };
+allow local_login_t var_run_t:file { rw_file_perms create };
+
```
doverride, line 143: You should instead add file transitions there so that these files don't end up mislabeled.

I would consider creating an interface like: `auth_log_filetrans_lastlog` with something like this in there:
`logging_log_filetrans($1, lastlog_t, file, lastlog)`. and then, i guess, `create an auth_create_lastlog_files() and call that.

for utmp i suppose call:
`init_runtime_filetrans_utmp` ... an interface to allow creating the file (something like `init_create_initrc_runtime_files()` seems to be missing so i guess that would have to be created as well

policy/modules/system/systemd.te
```diff
@@ -402,6 +402,8 @@ storage_raw_read_fixed_disk(systemd_generator_t)
 
 systemd_log_parse_environment(systemd_generator_t)
 
+term_dontaudit_use_unallocated_ttys(systemd_generator_t)
```
doverride, line 405: Youre not allowing it here as you suggest in the topic. Youre silently block it instead. I would probably *allow* this access.

policy/modules/system/systemd.te
```diff
@@ -1075,6 +1075,8 @@ allow systemd_resolved_t systemd_networkd_runtime_t:dir watch;
 allow systemd_resolved_t system_dbusd_runtime_t:dir read;
 allow systemd_resolved_t system_dbusd_runtime_t:sock_file read;
 
+allow systemd_resolved_t var_run_t:dir read;
```
doverride, line 1079: Youre not allowed to reference var_run_t here directly. instead use something like "files_list_runtime()"
probably also ends up wanting to "watch" (that is if your kernel has support for the watch permission)

policy/modules/system/systemd.te
```diff
@@ -1194,6 +1194,7 @@ files_delete_usr_files(systemd_tmpfiles_t)
 files_list_home(systemd_tmpfiles_t)
 files_list_locks(systemd_tmpfiles_t)
 files_manage_generic_tmp_dirs(systemd_tmpfiles_t)
+files_manage_non_auth_files(systemd_tmpfiles_t)
```
doverride, line 1200: set the state of the `systemd_tmpfiles_manage_all` boolean to true instead

policy/modules/system/locallogin.te
atenart: OK, that seems better indeed. I'll look into it, thanks!

policy/modules/system/systemd.te
atenart: I forgot to remove this one from the PR, sorry about that. And indeed we shoudl allow this instead :)

policy/modules/system/getty.te
```diff
@@ -47,8 +47,11 @@ allow getty_t getty_log_t:file { append_file_perms create_file_perms setattr_fil
 logging_log_filetrans(getty_t, getty_log_t, file)
 
 allow getty_t getty_runtime_t:dir watch;
+allow getty_t getty_runtime_t:file watch;
 manage_files_pattern(getty_t, getty_runtime_t, getty_runtime_t)
 files_runtime_filetrans(getty_t, getty_runtime_t, file)
+allow getty_t init_runtime_t:dir read;
+allow getty_t init_runtime_t:sock_file write;
```
doverride, line 54: Youre not allowed to reference init_runtime_t directly here. This looks like this might be "io.systemd.DynamicUsers", if so then auth_use_nsswitch(getty_t) should address it if and when https://github.com/SELinuxProject/refpolicy/pull/269 is merged

policy/modules/system/systemd.te
atenart: Will do!

policy/modules/system/systemd.te
atenart: OK, I'll look into that. Thanks!

policy/modules/system/getty.te
```diff
@@ -92,6 +96,12 @@ logging_send_syslog_msg(getty_t)
 
 miscfiles_read_localization(getty_t)
 
+allow getty_t sysctl_t:dir search;
+allow getty_t sysctl_kernel_t:dir search;
+allow getty_t sysctl_kernel_t:file { read open };
+
+allow getty_t init_t:unix_stream_socket connectto;
```
doverride, line 103: Not allowed to reference init_t here. probably io.systemd.dynamicusers. which should be addressed with auth_use_nsswitch() when https://github.com/SELinuxProject/refpolicy/pull/269 is merged

policy/modules/system/getty.te
```diff
@@ -92,6 +96,12 @@ logging_send_syslog_msg(getty_t)
 
 miscfiles_read_localization(getty_t)
 
+allow getty_t sysctl_t:dir search;
+allow getty_t sysctl_kernel_t:dir search;
+allow getty_t sysctl_kernel_t:file { read open };
```
doverride, line 101: not allowed to reference sysctl_t and sysctl_kernel_t here. Use the appropriate existing interfaces

policy/modules/system/locallogin.te
doverride: Well you called a pretty specific interface that to me suggested that your device was using the PAM motd-dynamic module (which is PAM module used by Debian)

Why exactly is that execute permission needed?

doverride: Actually now i see it "sh" is mislabeled ....
find / -inum 416

it should be labeled shell_exec_t

once you fixed  the label you should be able to omit this commit

atenart: Thanks! This fixed this issue.

policy/modules/system/systemd.te
```diff
@@ -1188,6 +1201,7 @@ files_read_etc_runtime_files(systemd_tmpfiles_t)
 files_relabel_all_lock_dirs(systemd_tmpfiles_t)
 files_relabel_all_runtime_dirs(systemd_tmpfiles_t)
 files_relabel_all_tmp_dirs(systemd_tmpfiles_t)
+files_relabel_non_auth_files(systemd_tmpfiles_t)
```
pebenito, line 1204: Not necessary, that's why there is the `systemd_tmpfiles_manage_all` boolean.

policy/modules/system/logging.te
```diff
@@ -524,7 +524,9 @@ ifdef(`init_systemd',`
 	allow syslogd_t self:netlink_audit_socket connected_socket_perms;
 	allow syslogd_t self:capability2 audit_read;
 	allow syslogd_t self:capability { chown setgid setuid sys_ptrace };
-	allow syslogd_t self:netlink_audit_socket { getattr getopt read setopt write };
+	allow syslogd_t self:netlink_audit_socket { getattr getopt read setopt write nlmsg_write };
+
+	logging_set_loginuid(syslogd_t)
```
pebenito, line 529: This doesn't seem to make sense. It isn't a login program and it's not a user.

policy/modules/system/sysnetwork.if
```diff
@@ -346,8 +346,12 @@ interface(`sysnet_read_config',`
 	')
 
 	files_search_etc($1)
+	files_search_runtime($1)
+	allow $1 net_conf_t:dir list_dir_perms;
 	allow $1 net_conf_t:file read_file_perms;
 
+	read_files_pattern($1, net_conf_t, net_conf_t)
```
pebenito, line 353: This is a subset of the above two lines.

policy/modules/system/systemd.te
```diff
@@ -751,6 +754,11 @@ allow systemd_networkd_t self:tun_socket { create_socket_perms relabelfrom relab
 allow systemd_networkd_t self:udp_socket create_socket_perms;
 allow systemd_networkd_t self:unix_dgram_socket create_socket_perms;
 
+allow systemd_networkd_t system_dbusd_runtime_t:dir read;
+allow systemd_networkd_t system_dbusd_runtime_t:sock_file read;
+allow systemd_networkd_t var_run_t:dir read;
```
pebenito, line 759: this plus init_search_runtime = init_list_runtime

policy/modules/system/systemd.te
```diff
@@ -751,6 +754,11 @@ allow systemd_networkd_t self:tun_socket { create_socket_perms relabelfrom relab
 allow systemd_networkd_t self:udp_socket create_socket_perms;
 allow systemd_networkd_t self:unix_dgram_socket create_socket_perms;
 
+allow systemd_networkd_t system_dbusd_runtime_t:dir read;
+allow systemd_networkd_t system_dbusd_runtime_t:sock_file read;
```
pebenito, line 758: These would need new interfaces.

policy/modules/system/systemd.te
```diff
@@ -1063,6 +1071,11 @@ allow systemd_resolved_t self:tcp_socket { accept listen };
 
 allow systemd_resolved_t systemd_networkd_runtime_t:dir watch;
 
+allow systemd_resolved_t system_dbusd_runtime_t:dir read;
+allow systemd_resolved_t system_dbusd_runtime_t:sock_file read;
```
pebenito, line 1075: this would call the new interfaces from above.

policy/modules/system/systemd.te
```diff
@@ -751,6 +754,11 @@ allow systemd_networkd_t self:tun_socket { create_socket_perms relabelfrom relab
 allow systemd_networkd_t self:udp_socket create_socket_perms;
 allow systemd_networkd_t self:unix_dgram_socket create_socket_perms;
 
+allow systemd_networkd_t system_dbusd_runtime_t:dir read;
+allow systemd_networkd_t system_dbusd_runtime_t:sock_file read;
+allow systemd_networkd_t var_run_t:dir read;
+allow systemd_networkd_t fs_t:filesystem getattr;
```
pebenito, line 760: `fs_getattr_xattr_fs(systemd_networkd_t)`

pebenito (changes requested):

policy/modules/system/logging.te
doverride: I suspect that `audit2allow -R` was being used here and in other places before. It probably returned a less than optimal interface in absence of a better alternative. Scary, I recently actually suggested that we remove that functionality for audit2allow as it is proven counter productive.

bauen1: Instead of completely removing it, maybe have it suggest appropriate macros for permission sets.
E.g. `allow a_t b_t:file read_file_perms` instead of `allow a_t b_t:file { open read }`.
This would mirror some of the functionality of selint.

policy/modules/system/systemd.te
atenart: Thanks! I removed it.

policy/modules/system/logging.te
atenart: I'm not using audit2allow. I removed the logging_set_loginuid() call for now; and I'll investigate it.

policy/modules/system/systemd.te
atenart: I added a new interface and replaced the two above and the two below lines.

policy/modules/system/systemd.te
atenart: I'm not sure what you meant here.

policy/modules/system/systemd.te
atenart: I'll use this interface, thanks!

policy/modules/services/dbus.if
```diff
@@ -594,6 +594,24 @@ interface(`dbus_watch_system_bus_runtime_dirs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dbus_read_system_bus_runtime_dirs',`
+	gen_require(`
+		type system_dbusd_runtime_t;
+	')
+
+	allow $1 system_dbusd_runtime_t:dir read;
```
pebenito, line 612: We use the verb `list` instead of `read` for read on directories. Please change the interface name to `dbus_list_system_bus_runtime` and use `list_dir_perms` instead of `read`.

policy/modules/system/systemd.te
```diff
@@ -756,6 +759,8 @@ allow systemd_networkd_t self:tun_socket { create_socket_perms relabelfrom relab
 allow systemd_networkd_t self:udp_socket create_socket_perms;
 allow systemd_networkd_t self:unix_dgram_socket create_socket_perms;
 
+fs_getattr_xattr_fs(systemd_networkd_t)
```
pebenito, line 762: Please move after the files_* calls below.

policy/modules/system/systemd.te
```diff
@@ -1068,6 +1075,8 @@ allow systemd_resolved_t self:tcp_socket { accept listen };
 
 allow systemd_resolved_t systemd_networkd_runtime_t:dir watch;
 
+files_list_runtime(systemd_resolved_t)
```
pebenito, line 1078: This should go with the other files_* below.

policy/modules/system/systemd.te
```diff
@@ -801,6 +806,8 @@ optional_policy(`
 	dbus_connect_system_bus(systemd_networkd_t)
 	dbus_watch_system_bus_runtime_dirs(systemd_networkd_t)
 	dbus_watch_system_bus_runtime_named_sockets(systemd_networkd_t)
+	dbus_read_system_bus_runtime_dirs(systemd_networkd_t)
+	dbus_read_system_bus_runtime_named_sockets(systemd_networkd_t)
```
pebenito, line 810: Thinking about these further it may make more sense to add these to `dbus_system_bus_client` instead.

pebenito (changes requested):

policy/modules/services/dbus.if
atenart: I renamed dbus_read_system_bus_runtime_dirs to dbus_list_system_bus_runtime and used list_dir_perms.
Thanks!

policy/modules/system/systemd.te
atenart: I added those two calls directly in dbus_system_bus_client.

pebenito (approved):

atenart:
Thanks!

PR #308, opened 2020-09-05
init: allow systemd to map the SELinux status page

cgzones:
systemd v247 will access the SELinux status page.

see https://github.com/systemd/systemd/pull/16821

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -424,8 +424,7 @@ dev_read_sysfs(systemd_hostnamed_t)
 
 files_read_etc_files(systemd_hostnamed_t)
 
-selinux_get_enforce_mode(systemd_hostnamed_t)
-selinux_map_security_files(systemd_hostnamed_t)
+selinux_use_status_page(systemd_hostnamed_t)
 
 seutil_read_file_contexts(systemd_hostnamed_t)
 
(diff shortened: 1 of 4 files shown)


policy/modules/system/systemd.te
```diff
@@ -424,6 +424,8 @@ dev_read_sysfs(systemd_hostnamed_t)
 
 files_read_etc_files(systemd_hostnamed_t)
 
+selinux_map_security_files(systemd_hostnamed_t)
```
topimiettinen, line 427: It probably does not hurt to add these where they strictly would not be needed, but since `hostnamed` (and some others below) didn't need any access to SELinux state before (there are no rules like `selinux_get_enforce_mode()` for `systemd_hostnamed_t`) perhaps it does not need it now either. So far I've only noticed the need for `systemd-tmpfiles` and `user-runtime-dir` (which seems to be missing from this PR).

cgzones: The mapping of the SELinux status page is in the same C-function as `selabel_open(3)`, so it should affect all domains having `selinux_map_security_files()` access.

pebenito:
I think we should take this opportunity to make a new interface that has all of the perms needed for using the status page. We should probably do the same for userspace AVC users, selabel_lookup, etc. but I'm not asking for that here.

pebenito (changes requested):
See above

pebenito (approved):

PR #311, opened 2020-10-02
Corosync & pacemaker changes to work in enforcing

dsugar100:
Sorry for so many related patches.  But I think it will be easier to review and comment on and to see why some of things are being proposed.

I'm happy to squash into a single commit once everything is ready to be merged.

I think the controversial pieces are going to be the proposed interface init_filetrans_unit which is to dynamically create directories/files in systemd_unit_t directories with a filetrans to a new type (intended for /run/systemd/system/ ) so the domain creating unit files (pacemaker in this case) don't need to have full read/write access to all systemd_unit_t files.

 The other one is the new interface corosync_mmap_rw_tmpfs which is to map/read/write corosync tmpfs file.  I thought it was better than changing the existing corosync_rw_tmpfs interface.  Though thinking about it now, it might be more standard to just add a new interface to grant map permission only.  

### DIFF
--- a/policy/modules/services/corosync.fc
+++ b/policy/modules/services/corosync.fc
@@ -6,7 +6,14 @@
 /usr/sbin/corosync	--	gen_context(system_u:object_r:corosync_exec_t,s0)
 /usr/sbin/corosync-notifyd	--	gen_context(system_u:object_r:corosync_exec_t,s0)
 
+ifdef(`distro_redhat',`
+/usr/share/corosync/corosync			--	gen_context(system_u:object_r:corosync_exec_t,s0)
+/usr/share/corosync/corosync-notifyd	--	gen_context(system_u:object_r:corosync_exec_t,s0)
+')
+
 /var/lib/corosync(/.*)?	gen_context(system_u:object_r:corosync_var_lib_t,s0)
+
+/var/lock/subsys/corosync	--	gen_context(system_u:object_r:corosync_lock_t,s0)
 
 /var/log/cluster/corosync\.log.*	--	gen_context(system_u:object_r:corosync_var_log_t,s0)
 
--- a/policy/modules/services/corosync.te
+++ b/policy/modules/services/corosync.te
@@ -13,6 +13,9 @@ domain_obj_id_change_exemption(corosync_t)
 type corosync_initrc_exec_t;
 init_script_file(corosync_initrc_exec_t)
 
+type corosync_lock_t;
+files_lock_file(corosync_lock_t)
+
 type corosync_runtime_t alias corosync_var_run_t;
 files_runtime_file(corosync_runtime_t)
 
@@ -43,6 +46,9 @@ allow corosync_t self:shm create_shm_perms;
 allow corosync_t self:unix_dgram_socket sendto;
 allow corosync_t self:unix_stream_socket { accept connectto listen };
 
+manage_files_pattern(corosync_t, corosync_lock_t, corosync_lock_t)
+files_lock_filetrans(corosync_t, corosync_lock_t, file)
+
 manage_dirs_pattern(corosync_t, corosync_tmp_t, corosync_tmp_t)
 manage_files_pattern(corosync_t, corosync_tmp_t, corosync_tmp_t)
 relabel_files_pattern(corosync_t, corosync_tmp_t, corosync_tmp_t)
@@ -50,6 +56,7 @@ files_tmp_filetrans(corosync_t, corosync_tmp_t, { dir file })
 
 manage_dirs_pattern(corosync_t, corosync_tmpfs_t, corosync_tmpfs_t)
 manage_files_pattern(corosync_t, corosync_tmpfs_t, corosync_tmpfs_t)
+mmap_read_files_pattern(corosync_t, corosync_tmpfs_t, corosync_tmpfs_t)
 fs_tmpfs_filetrans(corosync_t, corosync_tmpfs_t, { dir file })
 
 manage_files_pattern(corosync_t, corosync_var_lib_t, corosync_var_lib_t)
--- a/policy/modules/services/pacemaker.te
+++ b/policy/modules/services/pacemaker.te
@@ -121,6 +121,7 @@ tunable_policy(`pacemaker_startstop_all_services',`
 
 optional_policy(`
 	corosync_read_log(pacemaker_t)
+	corosync_mmap_rw_tmpfs(pacemaker_t)
 	corosync_stream_connect(pacemaker_t)
 ')
 
--- a/policy/modules/services/pacemaker.te
+++ b/policy/modules/services/pacemaker.te
@@ -5,6 +5,13 @@ policy_module(pacemaker, 1.6.0)
 # Declarations
 #
 
+## <desc>
+## <p>
+## Allow pacemaker to start/stop services
+## </p>
+## </desc>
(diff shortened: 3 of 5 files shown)


policy/modules/services/corosync.fc
```diff
@@ -6,8 +6,13 @@
 /usr/sbin/corosync	--	gen_context(system_u:object_r:corosync_exec_t,s0)
 /usr/sbin/corosync-notifyd	--	gen_context(system_u:object_r:corosync_exec_t,s0)
 
+/usr/share/corosync/corosync			--	gen_context(system_u:object_r:corosync_exec_t,s0)
+/usr/share/corosync/corosync-notifyd	--	gen_context(system_u:object_r:corosync_exec_t,s0)
```
pebenito, line 10: I looked at Fedora 32 and Debian 11 packages and neither have these paths, so this seems to be RHEL7 specific.

policy/modules/services/corosync.te
```diff
@@ -22,6 +22,9 @@ files_tmp_file(corosync_tmp_t)
 type corosync_tmpfs_t;
 files_tmpfs_file(corosync_tmpfs_t)
 
+type corosync_lock_t;
+files_lock_file(corosync_lock_t)
```
pebenito, line 26: This should go up after the initrc decl.

policy/modules/services/corosync.te
```diff
@@ -58,6 +62,9 @@ manage_sock_files_pattern(corosync_t, corosync_var_lib_t, corosync_var_lib_t)
 manage_fifo_files_pattern(corosync_t, corosync_var_lib_t,corosync_var_lib_t)
 files_var_lib_filetrans(corosync_t,corosync_var_lib_t, { dir fifo_file file sock_file })
 
+manage_files_pattern(corosync_t, corosync_lock_t, corosync_lock_t)
+files_lock_filetrans(corosync_t, corosync_lock_t, file)
```
pebenito, line 66: This should go up after the `self` group.

policy/modules/services/pacemaker.if
```diff
@@ -1,5 +1,25 @@
...
+## </param>
+#
+interface(`pacemaker_unit_file',`
+	gen_require(`
+		attribute pacemakerunit;
+	')
+
+	init_unit_file($1)
+	typeattribute $1 pacemakerunit;
+')
```
pebenito, line 21: With no callers of this in the PR, I'm not convinced this is the right way to go.

policy/modules/services/pacemaker.te
```diff
@@ -24,20 +33,28 @@ files_tmpfs_file(pacemaker_tmpfs_t)
 type pacemaker_var_lib_t;
 files_type(pacemaker_var_lib_t)
 
+type pacemaker_var_log_t;
```
pebenito, line 36: Please remove the `var` from the type name. I don't want more type names that encode a path like this.

policy/modules/system/init.if
```diff
@@ -3552,6 +3552,43 @@ interface(`init_manage_all_unit_files',`
...
+## </param>
+#
+interface(`init_filetrans_unit',`
+	gen_require(`
+		type systemd_unit_t;
+	')
+	allow $1 systemd_unit_t:dir search_dir_perms;
+	allow $1 systemd_unit_t:lnk_file read_lnk_file_perms;
+	filetrans_pattern($1, systemd_unit_t, $2, $3, $4)
+')
```
pebenito, line 3590: I don't think I can go with this since it technically affects any of the system unit dir locations, not just the runtime unit dir, as you mention in the summary. Perhaps we should create a new runtime unit type.

What do you think @doverride

pebenito (changes requested):
I'm ok with the `corosync_mmap_rw_tmpfs`.

policy/modules/services/corosync.fc
dsugar100: This could be RHEL7 specific.  Should be be wrapped in a "ifdef distro_redhat" block?

policy/modules/system/init.if
dsugar100: Yes, I think it does make sense to setup a 'systemd_runtime_unit_t' or similar type for these files.

policy/modules/services/pacemaker.if
dsugar100: Ok, how else should this be done?  I used it in a module within a policy module specific to the project being worked on.  Pacemaker starts/stops a service as the particular node in the cluster becomes primary.  

Rather than an attribute, this could be an interface like 'pacemaker_startstop_unit' (or similar) and the rule in the interface could be 'allow pacemaker_t $1:service { start stop };'    

I don't see any existing interfaces that have this model.  The existing interfaces that are similar are all the reverse where some type is granted permission to start/stop a known service (i.e. clamav_startstop_clamd, ntp_startstop, ntp_startstop, etc..)

It seems too heavy handed to use init_start_all_units(pacemaker_t) and init_stop_all_units(pacemaker_t), but that would also work.

Thoughts?

policy/modules/services/pacemaker.te
dsugar100: While I'm making changes to these modules, do you want a patch renaming corosync_var_log_t to corosync_log_t?

pebenito: No. I'd rather do renames like that separately, and across the entire policy.

policy/modules/services/corosync.fc
pebenito: Yes.

policy/modules/services/pacemaker.if
pebenito: With no example, I can't tell if an alternate concept for this would work. I don't want to add something while potentially having to deprecate it as soon as a real example comes along.

dsugar100: Ok, I will change this portion of the patch to model it after how the monit module does something similar.  In the monit case there is a tunable (defaulted to false) that when enabled allows monit to start/stop any service.  I will update and we can then decide if that is OK or if a different approach is preferable.

pebenito: I think we should probably have and keep this tunable even if we add something like the `packemaker_unit_file`.

dsugar100: I have updated this with the tunable.  I agree, it would stay even if we do get to an interface for this at some point.

policy/modules/system/init.if
dsugar100: This has been removed.  I realized there was an entry in the file_contexts.subs_dist making /run/systemd/system labeled list /lib/systemd/system.  Once I removed that I was able to use existing interface init_runtime_filetrans, hopefully this is more reasonable now.

doverride: Sorry for late reply, yes probably best to create a special type for runtime unit files.

pebenito (approved):
Looks good, though I'm going to rename the tunable to `pacemaker_startstop_all_services` after merging.

pebenito (changes requested):
Actually, can you reduce the number of commits. It doesn't have to be one, but something not more than about 4 would be nice.

dsugar100:
Yes, I will squash into a reasonable number of commits.  While I'm doing the squash I will rename the tunable as you have suggested.

pebenito (approved):

PR #319, opened 2020-11-20
userdomain, xserver: move xdg rules to unpriv_user_template

0xC0ncord:
xdg rules are normally set in xserver. But, if a modular policy is being
used and the xserver module is not present, the required rules for users
to be able to access xdg content are never created and thus these files
and directories cannot be interacted with by users. This change will
create these rules even in a scenario where the xserver module is not
present, i.e. on a headless server where the X server is not necessary.


0xC0ncord:
Signed-off-by: 0xC0ncord <me@concord.sh>

PR #320, opened 2020-11-23
userdomain, xserver: move xdg rules to unpriv_user_template

0xC0ncord:
xdg rules are normally set in xserver. But, if a modular policy is being
used and the xserver module is not present, the required rules for users
to be able to access xdg content are never created and thus these files
and directories cannot be interacted with by users. This change will
create these rules even in a scenario where the xserver module is not
present, i.e. on a headless server where the X server is not necessary.


policy/modules/services/xserver.if
```diff
@@ -185,42 +185,6 @@ interface(`xserver_role',`
...
-		xdg_relabel_music($2)
-		xdg_watch_music_dirs($2)
-		xdg_manage_pictures($2)
-		xdg_relabel_pictures($2)
-		xdg_watch_pictures_dirs($2)
-		xdg_manage_videos($2)
-		xdg_relabel_videos($2)
-		xdg_watch_videos_dirs($2)
-
 		xdg_cache_filetrans($2, mesa_shader_cache_t, dir, "mesa_shader_cache")
```
pebenito, line 188: Why not move this too?

pebenito (changes requested):
I'm ok with this move, but instead of directly putting them in `userdom_unpriv_user_template()`, they should go in a new template called `userdom_xdg_user_template()` so that if a policy writer wants to add the XDG usage onto a new user they're creating, it is 1 line rather than 34.

pebenito:
Also need to resolve the DCO sign off issue.

policy/modules/services/xserver.if
0xC0ncord: > I'm ok with this move, but instead of directly putting them in `userdom_unpriv_user_template()`, they should go in a new template called `userdom_xdg_user_template()` so that if a policy writer wants to add the XDG usage onto a new user they're creating, it is 1 line rather than 34.

My rationale for this move is that I ran into an issue with this on an SELinux-enabled headless server. This server does not have a display server (X, wayland) installed at all and the policy is built modularly. As a result the xdg.pp module is never built into the running policy. Because of this, the types for `xdg_config_t` and so on exist, but their `xdg_config_type` attributes are never set. The noticeable impact is that as a user, I do not have permission to access any objects with these types, to include `.config` and `.local` directories in my own home.

pebenito: I don't understand how my suggestion undermines your usage? `userdom_unpriv_user_template()` would call `userdom_xdg_user_template()`.

0xC0ncord: > `userdom_unpriv_user_template()` would call `userdom_xdg_user_template()`.

I'm sorry I think I get it now. To resolve the issue with the `mesa_shader_cache_t` type, do you think that can just be wrapped in an `optional_policy` in that template?

pebenito: I think the mesa line should remain where it is.

pebenito:
A good point made in the IRC channel: it may make more sense to move all of the xdg stuff into userdomain since non-gui stuff now use these directories.

policy/modules/services/xserver.if
0xC0ncord: I have moved the xdg rules to a new `userdom_xdg_user_template()`

0xC0ncord (commented):
I have moved the xdg rules to a new `userdom_xdg_user_template()`

0xC0ncord:
Updated commit message.

pebenito (approved):

pebenito:
This didn't come up in the PR tests as we're in the process of changing from travis-ci to github, so there is a build error:

```
policy/modules/roles/auditadm.te:9:ERROR 'unknown type auditadm_t_t' at token ';' on line 195019:
#line 9
	allow auditadm_t_t xdg_cache_type:dir { open read getattr lock search ioctl add_name remove_name write };
checkpolicy:  error(s) encountered while parsing configuration
make: *** [Rules.monolithic:71: policy.32] Error 1
```

I fixed this prior to merging.

PR #321, opened 2020-12-01
Remove systemd-analyze domain and executable type

dburgener:
The systemd-analyze domain has no unique permissions, and general usage seems to be to simply run in the caller domain.  Running in the caller domain makes sense because the access required is generally the access to read the information that systemd-analyze reads.  If the calling domain should be able to read the information it makes sense to grant it directly.

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -5,7 +5,6 @@
 
 /run/log/journal(/.*)?				gen_context(system_u:object_r:systemd_journal_t,s0)
 
-/usr/bin/systemd-analyze		--	gen_context(system_u:object_r:systemd_analyze_exec_t,s0)
 /usr/bin/systemd-cgtop			--	gen_context(system_u:object_r:systemd_cgtop_exec_t,s0)
 /usr/bin/systemd-coredump		--	gen_context(system_u:object_r:systemd_coredump_exec_t,s0)
 /usr/bin/systemd-detect-virt		--	gen_context(system_u:object_r:systemd_detect_virt_exec_t,s0)
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -24,7 +24,7 @@ template(`systemd_role_template',`
 	gen_require(`
 		attribute systemd_user_session_type, systemd_log_parse_env_type;
 		type systemd_user_runtime_t, systemd_user_runtime_notify_t;
-		type systemd_run_exec_t, systemd_analyze_exec_t;
+		type systemd_run_exec_t;
 	')
 
 	#################################
@@ -63,7 +63,7 @@ template(`systemd_role_template',`
 	# systemctl --user
 	stream_connect_pattern($3, systemd_user_runtime_t, systemd_user_runtime_t, $1_systemd_t)
 
-	can_exec($3, { systemd_run_exec_t systemd_analyze_exec_t })
+	can_exec($3, systemd_run_exec_t)
 
 	dbus_system_bus_client($1_systemd_t)
 ')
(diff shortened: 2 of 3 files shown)


doverride:
Daniel Burgener <notifications@github.com> writes:

> The systemd-analyze domain has no unique permissions, and general usage seems to be to simply run in the caller domain. Running in the caller
> domain makes sense because the access required is generally the access to read the information that systemd-analyze reads. If the calling
> domain should be able to read the information it makes sense to grant it directly.

Not sure that i agree, but it is certainly not an init_daemon_domain()

For example it needs "service and system status" of pretty much
everything to blame, it watches /run/mount , it runs bpf progs, talks to
systemd using its private socket, needs "access" to nspawn containers
for systemd-analyze -M, dbus chats with various, reads and writes
systemd shm/tmpfs (something that one otherwise never needs)

heres some of the rules i accumulated over time to give an idea:

https://git.defensec.nl/?p=dssp3.git;a=blob_plain;f=policy/systemd/systemd_analyze.cil;h=d9bee8eec4cbd9c6e96fdd0a7f7c87597aff0a78;hb=HEAD

>
> Signed-off-by: Daniel Burgener dburgener@linux.microsoft.com
>
> -----------------------------------------------------------------------------------------------------------------------------------------------
>
>
>   https://github.com/SELinuxProject/refpolicy/pull/321
>
> Commit Summary
>
> * Remove systemd-analyze domain and executable type
>
> File Changes
>
> * M policy/modules/system/systemd.fc (1) 
> * M policy/modules/system/systemd.if (4) 
> * M policy/modules/system/systemd.te (4) 
>
> Patch Links:
>
> * https://github.com/SELinuxProject/refpolicy/pull/321.patch
> * https://github.com/SELinuxProject/refpolicy/pull/321.diff
>
> —
>

-- 
Dominick Grift

dburgener:
Thanks for the link.  That's really helpful context.  Maybe that all does justify fleshing out the domain to actually have these sorts of accesses (as is in refpolicy it's totally unusable).  My main thought is that most of the read-like accesses it has seem like things that would be better controlled via the calling domain (for example not allowing the caller to know about certain unit files).  The systemd IPC is definitely a case in favor of a separate domain though, although it seems to me like there is some tradeoff in terms of being able to specify the sorts of information the parent can read rather than just letting them get the full power of systemd-analyze because they are able to run it for some things.

doverride:
Daniel Burgener <notifications@github.com> writes:

> Thanks for the link. That's really helpful context. Maybe that all does justify fleshing out the domain to actually have these sorts of
> accesses (as is in refpolicy it's totally unusable). My main thought is that most of the read-like accesses it has seem like things that would
> be better controlled via the calling domain (for example not allowing the caller to know about certain unit files). The systemd IPC is
> definitely a case in favor of a separate domain though, although it seems to me like there is some tradeoff in terms of being able to specify
> the sorts of information the parent can read rather than just letting them get the full power of systemd-analyze because they are able to run
> it for some things.

Yes but you dont have to transition if you want to limit functionality
then just run it without a transition. But sysadm_t for example should
be able to run it with the full functionality and targeting it enhances
integrity so that it can only do what it needs to do.

>
> —
>

-- 
Dominick Grift

pebenito:
> My main thought is that most of the read-like accesses it has seem like things that would be better controlled via the calling domain (for example not allowing the caller to know about certain unit files).

Can systemd-analyze work without the full permissions? If not then the domain seems to make sense.

dburgener:
> Can systemd-analyze work without the full permissions? If not then the domain seems to make sense.

My impression from comparing what I'm seeing in our environment to @doverride's policy is that we're not seeing the full set here.  That sounds like a reasonable conclusion to me.  I'll go ahead and close this and work with our org to get some systemd-analyze policy upstreamed.

PR #322, opened 2020-12-01
Switch to GitHub actions for CI actions.

pebenito:
Signed-off-by: Chris PeBenito <pebenito@ieee.org>


pebenito:
Tested here: https://github.com/pebenito/refpolicy/actions/runs/394764907

pebenito:
@fishilico can you take a quick look since you did this for the userspace repo

.github/workflows/tests.yml
```diff
@@ -0,0 +1,156 @@
+name: Build tests
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
```
cgzones, line 7: I think limiting the branches here disables CI runs for future pull requests.

pebenito: My understanding is that it limits to PRs to the default branch and pushes to the default branch:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

It is also used in the example workflows:

https://github.com/actions/starter-workflows/blob/main/ci/python-package.yml
https://github.com/actions/starter-workflows/blob/main/ci/c-cpp.yml

.github/workflows/tests.yml
```diff
@@ -0,0 +1,156 @@
...
+          - {type: mls, distro: debian, monolithic: y, systemd: y, apps-off: unconfined}
+          - {type: mls, distro: gentoo, monolithic: y, systemd: n, apps-off: unconfined}
+  
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.5
```
fishilico, line 92: Why is this specific version used? If there is a rationale of using the lowest supported version of Python, a comment could be added to help deciding in the future whether this version should be upgraded (and how)

fishilico (approved):
Hello, I read the GitHub Actions configuration and it seems to be fine :) Thanks for working on this.

I have a minor comment: feel free to add a comment as I suggest or not.

.github/workflows/tests.yml
pebenito: It's what we have in the travis test. IIRC it was the oldest Python they support. It's meant to be the minimum Python version needed to build (or at least as old as CI will support). I'll add a comment.

PR #326, opened 2020-12-15
Use self keyword when an AV rule source type matches destination

dburgener:
This is reported in a new SELint check in soon to be released selint version 1.2.0

### DIFF
--- a/policy/modules/kernel/kernel.te
+++ b/policy/modules/kernel/kernel.te
@@ -263,7 +263,7 @@ kernel_mounton_proc_dirs(kernel_t)
 kernel_request_load_module(kernel_t)
 
 # Allow unlabeled network traffic
-allow unlabeled_t unlabeled_t:packet { forward_in forward_out };
+allow unlabeled_t self:packet { forward_in forward_out };
 corenet_in_generic_if(unlabeled_t)
 corenet_in_generic_node(unlabeled_t)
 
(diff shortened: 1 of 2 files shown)


pebenito:
I don't know that I agree with this SELint check. I don't see what value it provides.

dburgener:
I would argue that it makes policy more readable for the following reasons:

- Consistency of style.  For anyone who's spent a decent amount of time in policy "allow some_long_domain_t some_long_domain_t:fifo_file rw_fifo_file_perms;" takes quite a bit longer to understand and parse than "allow some_long_domain_t self:fifo_file rw_fifo_file_perms;"
- Helps avoid errors around similarly named domains.  The above example requires extra attention if a type named "long_name_chronyd_t" and another named "long_name_chronyc_t" both exist.  It would be helpful to know that these types differ and require some extra attention when reviewing policy.
- Ability to grep policy for self rules

The primary motivating example on my end is a policy file defining a number of types that all have the same prefix and differ by suffix and where the self keyword is consistently not used.  It would be easier and clearer to understand the policy if self rules were clearly marked as such rather than needing to spend energy on careful attention to whether the two domains are identical.

Fundamentally though, I think consistency of style is the clearest argument for it.

If you don't like it though, that's fine.  I can abandon this and when 1.2.0 comes out we can disable the check in the refpolicy CI.

dburgener:
BTW, I am inclined to say that the specific examples in this PR don't seem overly problematic to me.  Particularly the unlabeled_t one may actually be clearer in the original version. The primary benefit in my view is the ability to enable the check broadly for potentially more confusing situations as highlighted above.  If you're interested in checking for those broadly but disagree with these specific cases, we could disable on these when we enable the check.

PR #330, opened 2021-01-06
Mc/nsswitch usage

minimaxwell:
Hello,

This PR adds the use of auth_use_nsswitch() for getty and systemd-tmpfiles, as discussed in PR #307 .

Now that PR #269 is merged, we can use auth_use_nsswitch to solve issues where getty and systemd-tmpfiles would need to access the userdb.

Any review is welcome !

Thanks to @atenart for starting this work.

Regards,

Maxime

### DIFF
--- a/policy/modules/system/getty.te
+++ b/policy/modules/system/getty.te
@@ -84,6 +84,7 @@ term_setattr_unallocated_ttys(getty_t)
 term_setattr_console(getty_t)
 
 auth_rw_login_records(getty_t)
+auth_use_nsswitch(getty_t)
 
 init_rw_utmp(getty_t)
 
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1322,6 +1322,7 @@ auth_manage_var_auth(systemd_tmpfiles_t)
 auth_relabel_lastlog(systemd_tmpfiles_t)
 auth_relabel_login_records(systemd_tmpfiles_t)
 auth_setattr_login_records(systemd_tmpfiles_t)
+auth_use_nsswitch(systemd_tmpfiles_t)
 
 init_manage_utmp(systemd_tmpfiles_t)
 init_manage_var_lib_files(systemd_tmpfiles_t)


policy/modules/system/getty.te
```diff
@@ -84,6 +84,7 @@ term_setattr_unallocated_ttys(getty_t)
 term_setattr_console(getty_t)
 
 auth_rw_login_records(getty_t)
+auth_use_nsswitch(getty_t)
```
pebenito, line 87: This seems strange, do you know why this is happening in the first place?

minimaxwell: Sorry I should have added more context.

This is actually due to interactions with io.systemd.DynamicUsers and userdb.

Adding auth_use_nsswitch(getty_t) fixes the following avc errors on a Buildroot-based system :

audit: type=1400 audit(1610032673.915:3): avc:  denied  { read } for  pid=75 comm="agetty" name="userdb" dev="tmpfs" ino=812 scontext=system_u:system_r:getty_t tcontext=system_u:object_r:systemd_userdb_runtime_t tclass=dir permissive=1

audit: type=1400 audit(1610032673.915:3): avc:  denied  { open } for  pid=75 comm="agetty" path="/run/systemd/userdb" dev="tmpfs" ino=812 scontext=system_u:system_r:getty_t tcontext=system_u:object_r:systemd_userdb_runtime_t tclass=dir permissive=1

audit: type=1400 audit(1610032673.959:4): avc:  denied  { getattr } for  pid=75 comm="agetty" path="/run/systemd/userdb" dev="tmpfs" ino=812 scontext=system_u:system_r:getty_t tcontext=system_u:object_r:systemd_userdb_runtime_t tclass=dir permissive=1

audit: type=1400 audit(1610032673.979:5): avc:  denied  { search } for  pid=75 comm="agetty" name="userdb" dev="tmpfs" ino=812 scontext=system_u:system_r:getty_t tcontext=system_u:object_r:systemd_userdb_runtime_t tclass=dir permissive=1

audit: type=1400 audit(1610032673.979:5): avc:  denied  { write } for  pid=75 comm="agetty" name="io.systemd.DynamicUser" dev="tmpfs" ino=814 scontext=system_u:system_r:getty_t tcontext=system_u:object_r:systemd_userdb_runtime_t tclass=sock_file permissive=1

pebenito: It seems like this shouldn't be necessary since AFAICS agetty only does host/domainname lookups. Does agetty break over those denials? If not, maybe we should split up `auth_use_nsswitch()` into different nsswitch services.

doverride:

> @pebenito commented on this pull request.
>
> ------------------------------------------------------------------------------------------------------------
>
> In policy/modules/system/getty.te:
>
>> @@ -84,6 +84,7 @@ term_setattr_unallocated_ttys(getty_t)
>  term_setattr_console(getty_t)
>  
>  auth_rw_login_records(getty_t)
> +auth_use_nsswitch(getty_t)
>
> It seems like this shouldn't be necessary since AFAICS agetty only does host/domainname lookups. Does agetty
> break over those denials? If not, maybe we should split up auth_use_nsswitch() into different nsswitch
> services.

I do not think it does host/domain name lookups (why would it?), instead it uses nss
password/group, not sure why though but it does chowns/fsetids the
terminals and it does list /etc, so those could explain why it may want to
resolve names from passwd/group

The DynamicUser also indicates that its about resolving user/group names

>
> —
>

-- 
Dominick Grift

pebenito:
I did some digging and I found agetty uses `gethostname()` here: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/term-utils/agetty.c#n1481 and `getdomainname()` here: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/term-utils/agetty.c#n1499

getty will read in the username but that is passed to the login program which does the lookups, authentication and tty chmod/relabel; agetty doesn't do any processing with the username. It does do a `getgrnam("tty")` here https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/term-utils/agetty.c#n1044 which I guess would trigger all the userdb denials above if `tty` isn't a group in the `files` nsswitch service or systemd is the first nsswitch service for the group database.

pebenito:
These rules ended up merging in #359. Sorry for letting this hang. Thanks for your contribution!

PR #333, opened 2021-01-13
Add tunable to allow evdev passthrough with qemu

0xC0ncord:
A feature of qemu allows users to pass through input devices directly into the virtual guest. When this is enabled, a user can press both CTRL keys at the same time, at which point the input devices specified on the qemu commandline have their input events sent directly to the guest. This is done via the IOCTL permission on the specified devices. This merge adds a boolean to allow this feature to be enabled.

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -2189,6 +2189,24 @@ interface(`dev_manage_input_dev',`
 
 ########################################
 ## <summary>
+##	IOCTL the input event devices (/dev/input).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dev_ioctl_input_dev',`
+	gen_require(`
+		type event_device_t;
+	')
+
+	allow $1 event_device_t:chr_file ioctl;
+')
+
+########################################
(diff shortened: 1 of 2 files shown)


pebenito:
Do you know the exact ioctl(s) that are used? I don't know if we will use extended permissions in the future, but it would helpful to know if so.

0xC0ncord:
> Do you know the exact ioctl(s) that are used? I don't know if we will use extended permissions in the future, but it would helpful to know if so.

I don't, but let me see if I can find out.

0xC0ncord:
It [looks like](https://github.com/qemu/qemu/blob/master/hw/input/virtio-input-host.c#L119-L138) the relevant code is here, but that's just a preemptive guess from glancing at it. Looks like there are several ioctls being used, all of which are defined [here](https://github.com/qemu/qemu/blob/master/include/standard-headers/linux/input.h), namely `0x90`, `0x91` (`EVIOCGRAB` and `EVIOCREVOKE`).

pebenito:
Thanks! Please add a comment about these in the tunable block (line 460) and I'll merge it.

0xC0ncord:
> Thanks! Please add a comment about these in the tunable block (line 460) and I'll merge it.

Done!

PR #335, opened 2021-01-14
Remove modules for programs that are deprecated or no longer supported.

pebenito:


pebenito:
### Please comment on the related refpolicy mail list thread instead of here.

PR #336, opened 2021-01-26
spamassassin: add rspamd support and tunable

0xC0ncord:
Additional rules are required to enable rspamd support. This commit adds
file contexts for rspamd's files and adds a tunable that enables the
additional rules needed for rspamd to function.

### DIFF
--- a/policy/modules/services/spamassassin.te
+++ b/policy/modules/services/spamassassin.te
@@ -21,6 +21,14 @@ gen_tunable(spamassassin_can_network, false)
 ## </desc>
 gen_tunable(spamd_enable_home_dirs, false)
 
+## <desc>
+##	<p>
+##	Determine whether extra rules should
+##	be enabled to support rspamd.
+##	</p>
+## </desc>
+gen_tunable(rspamd_spamd, false)
+
 type spamd_update_t;
 typealias spamd_update_t alias spamd_gpg_t;
 type spamd_update_exec_t;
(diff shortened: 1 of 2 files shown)


policy/modules/services/spamassassin.te
```diff
@@ -384,6 +395,30 @@ tunable_policy(`spamd_enable_home_dirs',`
...
+
+	fs_search_tmpfs(spamd_t)
+	manage_dirs_pattern(spamd_t, spamd_tmpfs_t, spamd_tmpfs_t)
+	manage_files_pattern(spamd_t, spamd_tmpfs_t, spamd_tmpfs_t)
+	mmap_read_files_pattern(spamd_t, spamd_tmpfs_t, spamd_tmpfs_t)
+	fs_tmpfs_filetrans(spamd_t, spamd_tmpfs_t, { dir file })
+
+	corenet_tcp_connect_http_port(spamd_t)
+	corenet_tcp_connect_redis_port(spamd_t)
+')
```
pebenito, line 420: I'm not completely convinced we need a tunable, or at least not everything needs to be in the tunable. Most of the perms  are innocuous read perms and the tmpfs ones aren't that concerning either. If the tunable is needed, then it would only need to be the network connections and maybe setrlimits IMO.

0xC0ncord: I think it is better not to grant perms when they are not needed, hence the tunable for these rules. I don't really have a problem granting those read perms for the entire spamd domain though.

pebenito (approved):
We can keep it as-is, and then if it becomes necessary, pull stuff out of the tunable.

PR #339, opened 2021-01-26
sudo: add tunable for HTTP connections

0xC0ncord:
It is possible to edit the PAM configuration for sudo in order to require users to authenticate against different mechanisms to use sudo. One such mechanism that may be useful is yubico-pam, which verifies a user's Yubikey against a validation server over HTTP(s). This commit adds a global and per-user domain tunable to allow sudo domains to make HTTP connections to facilitate this functionality.


policy/modules/admin/sudo.if
```diff
@@ -35,6 +35,17 @@ template(`sudo_role_template',`
...
+	## <desc>
+	##	<p>
+	##	Determine whether the sudo domain
+	##	can connect to TCP HTTP ports. This
+	##	is needed if an additional authentication
+	##	mechanism via an HTTP server is
+	##	required for users to use sudo.
+	##	</p>
+	## </desc>
+	gen_tunable(sudo_$1_tcp_connect_http_port, false)
```
pebenito, line 47: Why does this need to be per-role, rather than all or nothing?

0xC0ncord: It really doesn't need to be. I figured that the flexibility of making it per-role could be valuable. If not, I don't object to removing this.

pebenito: Lets start with a single tunable and if that's too limiting, we can always make it more granular.

pebenito (changes requested):

pebenito (approved):

PR #343, opened 2021-01-28
init: label systemd units in /etc

0xC0ncord:
Units can be created and configured in `/etc/systemd/system`. This commit labels these files appropriately as `systemd_unit_t`.

### DIFF
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -15,6 +15,7 @@
 /libx32 /usr/lib
 /sbin /usr/sbin
 /etc/init.d /etc/rc.d/init.d
+/etc/systemd/system /usr/lib/systemd/system
 /lib/systemd /usr/lib/systemd
 /run/lock /var/lock
 /usr/lib32 /usr/lib


doverride:
I would add a equivalency rule instead

/etc/systemd/system /usr/lib/systemd/system

https://git.defensec.nl/?p=dssp3.git;a=blob;f=config-mcs/file_contexts.subs_dist;h=6cf4bde54ca4badfd3bfd8d3f7cc18d96a2e2e7d;hb=HEAD#l3

pebenito:
I would like to go in a different way with this. I think we need a new type for units under /etc/systemd/system* and also a new one for under /run/systemd/system. Then systemd_unit_t would generally be ro, admins can write under /etc/systemd/system* and then have finer control for whatever can write in /run/systemd.

cgzones:
Fwiw: I am using a file context substitution of `/etc/systemd/system /usr/lib/systemd/system`, so edited unit files inherit the label of the standard file context definitions `/usr/lib/systemd/system/foo\.service`.

doverride:
> I would like to go in a different way with this. I think we need a new type for units under /etc/systemd/system* and also a new one for under /run/systemd/system. Then systemd_unit_t would generally be ro, admins can write under /etc/systemd/system* and then have finer control for whatever can write in /run/systemd.

That would get really messy. I agree that runtime users are different but /etc/systemd/system and /usr/lib/systemd/system (and /etc/systemd/user plus /usr/lib/systemd/user for that matter) should not be

0xC0ncord:
I forgot about runtime units actually. I'm going to update this with those included but I'm expecting more discussion to come of how exactly this should be done, given the differing viewpoints.

doverride:
> > I would like to go in a different way with this. I think we need a new type for units under /etc/systemd/system* and also a new one for under /run/systemd/system. Then systemd_unit_t would generally be ro, admins can write under /etc/systemd/system* and then have finer control for whatever can write in /run/systemd.
> 
> That would get really messy. I agree that runtime units are different but /etc/systemd/system and /usr/lib/systemd/system (and /etc/systemd/user plus /usr/lib/systemd/user for that matter) should not be

One has to consider that the unit in /etc takes precedence over the unit in /usr/lib. So if admin decides to override a given unit in /usr/lib by creating a unit with the same name in /etc then the access checks apply

so if theres a server that needs to be able to for example start /usr/lib/systemd/system/foo.service, then that service also potentially need to be able to start /etc/systemd/system/foo.service. By keeping the two the same you ensure consistency and that improves robustness

0xC0ncord:
> You added the rules in the wrong location. But that aside, the units in /run/system should be treated differently. They are special.

Ah I see now. My bad. Perhaps we can create the equivalency for `/etc/systemd/system -> /lib/systemd/system` and create a new type for runtime units in a different PR?

pebenito:
> One has to consider that the unit in /etc takes precedence over the unit in /usr/lib.

This is the exact reason that I'm suggesting it be a different label.

doverride:
> > One has to consider that the unit in /etc takes precedence over the unit in /usr/lib.
> 
> This is the exact reason that I'm suggesting it be a different label.

Okay I just wanted everyone to be aware of the potential pitfalls lurking

pebenito:
> > > One has to consider that the unit in /etc takes precedence over the unit in /usr/lib.
> > 
> > This is the exact reason that I'm suggesting it be a different label.
> 
> Okay I just wanted everyone to be aware of the potential pitfalls lurking

Agreed there are difficulties; I'm not blocking this PR over this idea. There's issues beyond what you may be thinking, e.g., there's also /etc/systemd/system.attached for [protable services](https://systemd.io/PORTABLE_SERVICES/).

doverride:
> > > > One has to consider that the unit in /etc takes precedence over the unit in /usr/lib.
> > > 
> > > 
> > > This is the exact reason that I'm suggesting it be a different label.
> > 
> > 
> > Okay I just wanted everyone to be aware of the potential pitfalls lurking
> 
> Agreed there are difficulties; I'm not blocking this PR over this idea. There's issues beyond what you may be thinking, e.g., there's also /etc/systemd/system.attached for [portable services](https://systemd.io/PORTABLE_SERVICES/).

For sure. I actually use portable services. I have two: my bully eggdrop bot is a portable service and my opendkim service is a portable service.

```
[root@agnus kcinimod]# portablectl
NAME       TYPE      RO CRTIME                       MTIME                        USAGE  STATE
eggdrop    raw       no Tue 2020-05-12 11:44:56 CEST Tue 2020-05-12 12:31:59 CEST 494.1M running
lost+found directory no n/a                          n/a                          n/a    detached
opendkim   raw       no Tue 2020-04-14 11:43:18 CEST Thu 2021-01-28 12:33:59 CET  567.3M running
```

There is also system.control.

However in my vision it should work (but we could use a patch to systemctl to make it a bit more robust)

doverride:
But to touch a bit on that how that systemd-portable scenario fits into all of this:

In theory any service can be portable and ideally that should just work out of the box without any changes required to the policy.

This is why i also have a "/etc/systemd/system.attached /usr/lib/systemd/system" equivalence rule. In theory this works however in practice systemd-portabled is missing some setfscreate calls (so that it creates the units in /etc/systemd/system.attached with the proper unit labels.

Same with systemctl edit [--full] by the way. Ideally that command should use setfscreate so that it creates the unit in /etc/systemd/system with the appropriate label.

pebenito (approved):

PR #346, opened 2021-01-29
Added support for tmux xdg config path

jpds:
tmux attempts to read `~/.config/tmux/tmux.conf` and displays a permission denied error without this addition.

### DIFF
--- a/policy/modules/apps/screen.te
+++ b/policy/modules/apps/screen.te
@@ -110,3 +110,7 @@ tunable_policy(`use_nfs_home_dirs',`
 	fs_manage_nfs_named_pipes(screen_domain)
 	fs_read_nfs_symlinks(screen_domain)
 ')
+
+optional_policy(`
+	xdg_search_config_dirs(screen_domain)
+')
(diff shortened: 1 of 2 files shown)


policy/modules/apps/screen.te
```diff
@@ -97,6 +97,8 @@ userdom_create_user_pty(screen_domain)
 userdom_setattr_user_ptys(screen_domain)
 userdom_setattr_user_ttys(screen_domain)
 
+xdg_read_config_files(screen_domain)
```
pebenito, line 100: Is tmux reading other files under $HOME/.config or is this just for the directory access?

dsugar100: Also, please wrap this in a optional block so xdg module can be disabled and screen still enabled.

jpds: @pebenito Indeed for directory access, without it I hit an error.

@dsugar100 Done.

jpds: @pebenito Just tried it with search_dirs() and that works fine.

pebenito (approved):

PR #347, opened 2021-01-29
certbot: add support for acme.sh

0xC0ncord:
acme.sh is an alternative to cerbot written in pure Unix shell.

### DIFF
--- a/policy/modules/services/certbot.fc
+++ b/policy/modules/services/certbot.fc
@@ -1,4 +1,6 @@
 /usr/bin/certbot	--	gen_context(system_u:object_r:certbot_exec_t,s0)
 /usr/bin/letsencrypt	--	gen_context(system_u:object_r:certbot_exec_t,s0)
+/usr/share/acme\.sh/acme\.sh	--	gen_context(system_u:object_r:certbot_exec_t,s0)
 /var/lib/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_lib_t,s0)
 /var/log/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_log_t,s0)
+/var/lib/acme\.sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
--- a/policy/modules/services/certbot.te
+++ b/policy/modules/services/certbot.te
@@ -1,5 +1,13 @@
 policy_module(certbot, 1.0.0)
 
+## <desc>
+##	<p>
+##	Determine whether additional rules
+##	should be enabled to support acme.sh
+##	</p>
+## </desc>
+gen_tunable(certbot_acmesh, false)
+
 ########################################
 #
 # Declarations
@@ -92,6 +100,11 @@ sysnet_read_config(certbot_t)
 
 userdom_dontaudit_search_user_home_dirs(certbot_t)
 userdom_use_user_ptys(certbot_t)
+
+tunable_policy(`certbot_acmesh',`
+	corecmd_exec_bin(certbot_t)
+	corecmd_exec_shell(certbot_t)
+')
 
 optional_policy(`
 	# for writing to webroot


policy/modules/services/certbot.fc
```diff
@@ -2,3 +2,9 @@
 /usr/bin/letsencrypt	--	gen_context(system_u:object_r:certbot_exec_t,s0)
 /var/lib/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_lib_t,s0)
 /var/log/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_log_t,s0)
+
+/etc/acme-sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
+/usr/share/acme\.sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
```
doverride, line 7: I strongly suspect that this entry should not be needed.

policy/modules/services/certbot.fc
```diff
@@ -2,3 +2,9 @@
...
+/usr/share/acme\.sh/acme\.sh	--	gen_context(system_u:object_r:certbot_exec_t,s0)
+/root/\.acme\.sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
+/root/\.acme\.sh/acme\.sh	--	gen_context(system_u:object_r:certbot_exec_t,s0)
```
doverride, line 10: I would probably not support this "/root" scenario

doverride (commented):
some obervations

policy/modules/services/certbot.fc
0xC0ncord: Ok. I might need to adapt this to more of a general use case scenario. This is more or less translated directly from how I have it set up on my boxes. Perhaps just remove the context for `/usr/share/acme.sh` and move acme.sh's lib dir to `/var/lib/acme-sh` for consistency?

policy/modules/services/certbot.fc
0xC0ncord: Ok.

policy/modules/services/certbot.fc
0xC0ncord: I took another look at this. On my system acme.sh stores its additional scripts for deployment and such here in `/usr/share/acme-sh/{deploy,dnsapi,notify}`, so these almost certainly should be `certbot_exec_t`.

doverride:
0xC0ncord <notifications@github.com> writes:

>
>
> In policy/modules/services/certbot.fc:
>
>> @@ -2,3 +2,9 @@
>  /usr/bin/letsencrypt	--	gen_context(system_u:object_r:certbot_exec_t,s0)
>  /var/lib/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_lib_t,s0)
>  /var/log/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_log_t,s0)
> +
> +/etc/acme-sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
> +/usr/share/acme\.sh(/.*)?		gen_context(system_u:object_r:certbot_lib_t,s0)
>
> I took another look at this. On my system acme.sh stores its additional scripts for deployment and
> such here in /usr/share/acme-sh/{deploy,dnsapi,notify}, so these almost certainly should be
> certbot_exec_t.

are the scripts mode +x?
It could just functions that are sourced.
Probably best to not assume and instead just add what you can vouch for
and leave the remainder. Eventually it will surface and addressed 

>
> —
>

-- 
Dominick Grift

0xC0ncord:
> are the scripts mode +x? It could just functions that are sourced. Probably best to not assume and instead just add what you can vouch for and leave the remainder. Eventually it will surface and addressed

Good call. They are not, so my initial thought seems to be correct. As it stands `certbot_t` does not have `read` on generic `var_lib_t` which is why I think `certbot_lib_t` is appropriate here. My issue though is that `certbot_lib_t` seems to be used for the volatile files (certificates and such) as the policy is written, when I think there really should be `certbot_cert_t` and `certbot_tls_privkey_t` types for these.

doverride:
> > are the scripts mode +x? It could just functions that are sourced. Probably best to not assume and instead just add what you can vouch for and leave the remainder. Eventually it will surface and addressed
> 
> Good call. They are not, so my initial thought seems to be correct. As it stands `certbot_t` does not have `read` on generic `var_lib_t` which is why I think `certbot_lib_t` is appropriate here. My issue though is that `certbot_lib_t` seems to be used for the volatile files (certificates and such) as the policy is written, when I think there really should be `certbot_cert_t` and `certbot_tls_privkey_t` types for these. Alternatively, `certbot` could just manage generic certs but I think that is too broad.

No,the /usr/share/acme.sh(/.*)? directory can just stay usr_t. certbot_t has access to read usr_t type context and that is all it needs to source the various functions for the files in there

0xC0ncord:
> No,the /usr/share/acme.sh(/.*)? directory can just stay usr_t. certbot_t has access to read usr_t type context and that is all it needs to source the various functions for the files in there

Ah ha, now I understand. Just overthinking it is all.

policy/modules/services/certbot.fc
```diff
@@ -2,3 +2,6 @@
 /usr/bin/letsencrypt	--	gen_context(system_u:object_r:certbot_exec_t,s0)
 /var/lib/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_lib_t,s0)
 /var/log/letsencrypt(/.*)?	gen_context(system_u:object_r:certbot_log_t,s0)
+
+/usr/share/acme\.sh/acme\.sh	--	gen_context(system_u:object_r:certbot_exec_t,s0)
```
pebenito, line 6: Needs to be up with the other `/usr` lines.

policy/modules/services/certbot.te
```diff
@@ -1,5 +1,13 @@
 policy_module(certbot, 1.0.0)
 
+## <desc>
+##	<p>
+##	Determine whether additional rules
+##	should be enabled to support acme.sh
+##	</p>
+## </desc>
+gen_tunable(acmesh_certbot, false)
```
pebenito, line 9: Please reverse the tunable name so `certbot_` is first.

policy/modules/services/certbot.te
```diff
@@ -99,3 +107,8 @@ optional_policy(`
 
 	apache_search_config(certbot_t)
 ')
+
+tunable_policy(`acmesh_certbot',`
```
pebenito, line 111: Tunables go before optionals.

pebenito (changes requested):

pebenito (approved):

PR #351, opened 2021-02-02
dovecot, postfix: add missing accesses

0xC0ncord:
postfix_pipe_t requires reading dovecot configuration and connecting to
dovecot stream sockets if configured to use dovecot for local mail
delivery.

### DIFF
--- a/policy/modules/services/dovecot.if
+++ b/policy/modules/services/dovecot.if
@@ -63,6 +63,28 @@ interface(`dovecot_domtrans_deliver',`
 
 ########################################
 ## <summary>
+##	Read dovecot configuration content.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`dovecot_read_config',`
+	gen_require(`
+		type dovecot_etc_t;
+	')
+
+	files_search_etc($1)
+	allow $1 dovecot_etc_t:dir list_dir_perms;
+	allow $1 dovecot_etc_t:file read_file_perms;
+	allow $1 dovecot_etc_t:lnk_file read_lnk_file_perms;
+')
+
+########################################
+## <summary>
 ##	Create, read, write, and delete
 ##	dovecot spool files.
 ## </summary>
--- a/policy/modules/services/postfix.te
+++ b/policy/modules/services/postfix.te
@@ -596,6 +596,8 @@ corecmd_exec_bin(postfix_pipe_t)
 
 optional_policy(`
 	dovecot_domtrans_deliver(postfix_pipe_t)
+	dovecot_read_config(postfix_pipe_t)
+	dovecot_stream_connect(postfix_pipe_t)
 ')
 
 optional_policy(`


policy/modules/services/dovecot.if
```diff
@@ -61,6 +61,85 @@ interface(`dovecot_domtrans_deliver',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dovecot_rw_stream_sockets',`
+	gen_require(`
+		type dovecot_t;
+	')
+
+	allow $1 dovecot_t:unix_stream_socket { connectto rw_stream_socket_perms };
```
doverride, line 80: I suspect that the rw_stream_socket_perms is a bit too much. Would just `rw_socket_perms` work instead?

policy/modules/services/dovecot.if
```diff
@@ -61,6 +61,85 @@ interface(`dovecot_domtrans_deliver',`
...
+##	</summary>
+## </param>
+#
+interface(`dovecot_rw_runtime_sock_files',`
+	gen_require(`
+		type dovecot_runtime_t;
+	')
+
+    rw_sock_files_pattern($1, dovecot_runtime_t, dovecot_runtime_t)
+')
```
doverride, line 100: I would probably add a files_search_pids($1) here (if it is still called that)

policy/modules/services/dovecot.if
0xC0ncord: > I suspect that the rw_stream_socket_perms is a bit too much. Would just `rw_socket_perms` work instead?

Let me try it. These are rules coming from one of my machines that is set up in this way.

policy/modules/services/dovecot.if
```diff
@@ -61,6 +61,85 @@ interface(`dovecot_domtrans_deliver',`
...
+## </param>
+## <rolecap/>
+#
+interface(`dovecot_search_runtime',`
+	gen_require(`
+		type dovecot_runtime_t;
+	')
+
+	search_dirs_pattern($1, dovecot_runtime_t, dovecot_runtime_t)
+')
```
doverride, line 141: this one is redundant and can be removed as `dovecot_rw_runtime_sock_files` implies this access

policy/modules/services/postfix.te
```diff
@@ -13,6 +13,15 @@ policy_module(postfix, 1.23.2)
...
 gen_tunable(postfix_local_write_mail_spool, true)
 
+## <desc>
+##	<p>
+##	Determine whether postfix can
+##	communicate with dovecot for
+##	authentication and delivery.
+##	</p>
+## </desc>
+gen_tunable(postfix_use_dovecot, false)
```
doverride, line 23: I would personally probably make this access unconditional, but that is subjective

policy/modules/services/dovecot.if
0xC0ncord: `files_search_pids` was renamed to `files_search_runtime`, but yes I will add that here.

policy/modules/services/postfix.te
0xC0ncord: I agree that I can't personally think of any situation where one may want to have postfix and dovecot installed simultaneously yet not configured to interoperate, I think there's still some value in having this behavior conditional. What about enabling it by default instead?

policy/modules/services/dovecot.if
0xC0ncord: After some more testing, the only permissions needed by `postfix_pipe_t` here seems to be `connectto` and `write`. So it looks like we can reduce this access to `rw_socket_perms` as you said.

doverride:
0xC0ncord <notifications@github.com> writes:

>
>
> In policy/modules/services/postfix.te:
>
>> @@ -13,6 +13,15 @@ policy_module(postfix, 1.23.2)
>  ## </desc>
>  gen_tunable(postfix_local_write_mail_spool, true)
>  
> +## <desc>
> +##	<p>
> +##	Determine whether postfix can
> +##	communicate with dovecot for
> +##	authentication and delivery.
> +##	</p>
> +## </desc>
> +gen_tunable(postfix_use_dovecot, false)
>
> I agree that I can't personally think of any situation where one may want to have postfix and dovecot
> installed simultaneously yet not configured to interoperate, I think there's still some value in
> having this behavior conditional. What about enabling it by default instead?

I dont have a strong feeling on it. eitherway is fine with me and i just
thought i'd mention it for consideration.

>
> —
>

-- 
Dominick Grift

policy/modules/services/postfix.te
pebenito: I agree with Dominick on this. The access isn't really dangerous.

policy/modules/services/dovecot.if
pebenito: This should probably moved into the stream socket connection interface. This is pretty much stream_connect_pattern() but with a read on the `sock_file`.

policy/modules/services/dovecot.if
```diff
@@ -61,6 +61,68 @@ interface(`dovecot_domtrans_deliver',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dovecot_rw_stream_sockets',`
+	gen_require(`
+		type dovecot_t;
+	')
+
+	allow $1 dovecot_t:unix_stream_socket { connectto rw_socket_perms };
```
pebenito, line 80: the `rw_socket_perms` vs `connectto` seems odd. Is it receiving a fd for a socket from dovecot?

0xC0ncord: > the `rw_socket_perms` vs `connectto` seems odd. Is it receiving a fd for a socket from dovecot?

Thanks for raising this for review again. Looks like I got this access mixed up actually after another thorough look, so `rw_socket_perms` isn't needed at all here.:
```
time->Wed Feb  3 11:16:14 2021
type=PROCTITLE msg=audit(1612368974.831:29966): proctitle=2F7573722F6C6962657865632F646F7665636F742F64656C69766572002D6600746F6E79746865736C6179657240676D61696C2E636F6D002D64006D6540636F6E636F72642E7368
type=PATH msg=audit(1612368974.831:29966): item=0 name="/run/dovecot/stats-writer" inode=1652 dev=00:11 mode=0140660 ouid=0 ogid=5000 rdev=00:00 obj=system_u:object_r:dovecot_runtime_t:s0 nametype=NORMAL cap_fp=0 cap_fi=0 cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(1612368974.831:29966): cwd="/var/spool/postfix"
type=SOCKADDR msg=audit(1612368974.831:29966): saddr=01002F72756E2F646F7665636F742F73746174732D7772697465720000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
type=SYSCALL msg=audit(1612368974.831:29966): arch=c000003e syscall=42 success=yes exit=0 a0=7 a1=779e570b52d8 a2=6e a3=6f9e24fa2aa0 items=1 ppid=2010426 pid=2010427 auid=4294967295 uid=5000 gid=5000 euid=5000 suid=5000 fsuid=5000 egid=5000 sgid=5000 fsgid=5000 tty=(none) ses=4294967295 comm="deliver" exe="/usr/libexec/dovecot/dovecot-lda" subj=system_u:system_r:postfix_pipe_t:s0 key=(null)
type=AVC msg=audit(1612368974.831:29966): avc:  denied  { connectto } for  pid=2010427 comm="deliver" path="/run/dovecot/stats-writer" scontext=system_u:system_r:postfix_pipe_t:s0 tcontext=system_u:system_r:dovecot_t:s0 tclass=unix_stream_socket permissive=1
type=AVC msg=audit(1612368974.831:29966): avc:  denied  { write } for  pid=2010427 comm="deliver" name="stats-writer" dev="tmpfs" ino=1652 scontext=system_u:system_r:postfix_pipe_t:s0 tcontext=system_u:object_r:dovecot_runtime_t:s0 tclass=sock_file permissive=1
```

policy/modules/services/postfix.te
0xC0ncord: OK.

policy/modules/services/dovecot.if
0xC0ncord: OK.

0xC0ncord:
Turns out that after reworking this, the proper solution was actually much simpler.

pebenito (approved):

PR #353, opened 2021-02-05
Assortment of systemd-related denial fixes

KrissN:
Fix a number of issues related to systemd start-up in enforced mode, primarily focused on `systemd-tmpfilesd` and `systemd-networkd`.

Notable changes include:
 * Support for systemd-powered initrd (dracut). This is related to a feature in systemd, which allows it to pass certain socket file descriptors from the initrd's systemd daemon into the main systemd process. Since these are opened before the SELinux policy is loaded, all of them use the default label, which causes a number of denials, as the policy expects them to be labelled. This behaviour is controlled by a new global tunable `init_reuse_initrd_sockets`.
 * Support for factory template files in `systemd-tmpfilesd`.
 * `systemd-networkd` start-up fixes.

### DIFF
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -5078,6 +5078,25 @@ interface(`fs_relabel_tmpfs_blk_file',`
 
 ########################################
 ## <summary>
+##	Relabel named pipes on tmpfs filesystems.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_relabel_tmpfs_fifo_files',`
+	gen_require(`
+		type tmpfs_t;
+	')
+
+	allow $1 tmpfs_t:dir list_dir_perms;
+	relabel_fifo_files_pattern($1, tmpfs_t, tmpfs_t)
+')
+
+########################################
+## <summary>
 ##	Read and write, create and delete generic
 ##	files on tmpfs filesystems.
 ## </summary>
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -4999,13 +4999,28 @@ interface(`fs_dontaudit_use_tmpfs_chr_dev',`
 ##	</summary>
 ## </param>
 #
-interface(`fs_relabel_tmpfs_chr_file',`
+interface(`fs_relabel_tmpfs_chr_files',`
 	gen_require(`
 		type tmpfs_t;
 	')
 
 	allow $1 tmpfs_t:dir list_dir_perms;
 	relabel_chr_files_pattern($1, tmpfs_t, tmpfs_t)
+')
+
+########################################
+## <summary>
+##	Relabel character nodes on tmpfs filesystems.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_relabel_tmpfs_chr_file',`
+	refpolicywarn(`$0($*) has been deprecated, please use fs_relabel_tmpfs_chr_files() instead.')
+	fs_relabel_tmpfs_chr_files($1)
 ')
 
 ########################################
@@ -5037,13 +5052,28 @@ interface(`fs_rw_tmpfs_blk_files',`
 ##	</summary>
 ## </param>
 #
-interface(`fs_relabel_tmpfs_blk_file',`
+interface(`fs_relabel_tmpfs_blk_files',`
 	gen_require(`
 		type tmpfs_t;
 	')
 
 	allow $1 tmpfs_t:dir list_dir_perms;
 	relabel_blk_files_pattern($1, tmpfs_t, tmpfs_t)
+')
+
+########################################
+## <summary>
+##	Relabel block nodes on tmpfs filesystems.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_relabel_tmpfs_blk_file',`
+	refpolicywarn(`$0($*) has been deprecated, please use fs_relabel_tmpfs_blk_files() instead.')
+	fs_relabel_tmpfs_blk_files($1)
 ')
 
 ########################################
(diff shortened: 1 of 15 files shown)


policy/modules/system/systemd.te
```diff
@@ -178,6 +178,7 @@ type systemd_networkd_exec_t;
 init_system_domain(systemd_networkd_t, systemd_networkd_exec_t)
 
 type systemd_networkd_runtime_t alias systemd_networkd_var_run_t;
+files_mountpoint(systemd_networkd_runtime_t)
```
doverride, line 181: This is one of those things. Not sure if we reached any consensus on this topic. but personally i would argue that these aren't your average mountpoints and as such shouldnt be treated that way

The only thing ever mounting on systemd_networkd_runtime_t type directories will ever be systemd.

This rule is too broad in my view

KrissN: An alternative would be a direct rule granting `init_t` rights to `mounton` on `systemd_networkd_runtime_t`. This would need to go via an interface and be conditional on `init_systemd`.

KrissN: Or maybe a better one: declare an attribute `systemd_nspawn_mountpoint` and allow systemd's `init_t` `mounton` rights to it. This would allow for future extension.

KrissN: Scratch that - there is an interface for that already: `init_mountpoint`.

doverride: I guess "systemd_mountpoint_type" should be suffice. It is not only for nspawn.

doverride: as a matter of fact, nspawn has little to do with this AFAIK

doverride: oh ok, yes init_mountpoint sounds appropriate

policy/modules/system/logging.if
```diff
@@ -681,10 +681,9 @@ interface(`logging_send_syslog_msg',`
 		# Allow systemd-journald to check whether the process died
 		allow syslogd_t $1:process signull;
 
-		ifdef(`distro_redhat',`
-			kernel_dgram_send($1)
-		')
 	')
+
+	kernel_dgram_send($1)
```
pebenito, line 686: This looks like it is at the wrong level block. It should still be in `init_systemd`, right?

policy/modules/system/logging.te
```diff
@@ -501,7 +501,7 @@ auth_use_nsswitch(syslogd_t)
 
 init_use_fds(syslogd_t)
 
-# cjp: this doesnt make sense
+# needed for systemd-initrd case when syslog socket is unlabelled
 logging_send_syslog_msg(syslogd_t)
```
pebenito, line 505: In that case we should be able to move it to the `init_systemd` block.

policy/modules/system/systemd.te
```diff
@@ -437,12 +449,19 @@ kernel_use_fds(systemd_generator_t)
 kernel_read_system_state(systemd_generator_t)
 kernel_read_kernel_sysctls(systemd_generator_t)
 
+# systemd-getty-generator
+term_use_unallocated_ttys(systemd_generator_t)
+
```
pebenito, line 454: Duplicate rule.

policy/modules/system/systemd.te
```diff
@@ -1306,6 +1326,8 @@ files_manage_var_lib_dirs(systemd_tmpfiles_t)
 files_purge_tmp(systemd_tmpfiles_t)
 files_read_etc_files(systemd_tmpfiles_t)
 files_read_etc_runtime_files(systemd_tmpfiles_t)
+files_relabel_config_dirs(systemd_tmpfiles_t)
+files_relabel_config_files(systemd_tmpfiles_t)
```
pebenito, line 1330: It seems like this should also go in the factory tunable.

policy/modules/system/authlogin.te
```diff
@@ -87,6 +87,7 @@ type wtmp_t;
 logging_log_file(wtmp_t)
 
 optional_policy(`
+	systemd_tmpfilesd_managed(faillog_t, dir)
 	systemd_tmpfilesd_managed(faillog_t, file)
```
pebenito, line 91: Nit: this can be one line.

policy/modules/kernel/filesystem.if
```diff
@@ -5046,6 +5046,25 @@ interface(`fs_relabel_tmpfs_blk_file',`
...
+## <summary>
+##	Relabel named pipes on tmpfs filesystems.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_relabel_tmpfs_fifo_file',`
```
pebenito, line 5059: There is some inconsistency in the existing interfaces. It should be `fs_relabel_tmpfs_fifo_files` (plural)

pebenito (changes requested):

policy/modules/system/systemd.te
KrissN: > It seems like this should also go in the factory tunable.

Not necessarily.

According to my observations systemd will always try to restorecon the files, even if the target files exist and there is no need to copy them from the factory dir.

An alternative would be to put them - as you requested - into the factory conditional and put a dontaudit rule in the else section.

policy/modules/kernel/filesystem.if
KrissN: I can see the same inconsistency in the sibling interfaces (`fs_relabel_tmpfs_chr_file` & `fs_relabel_tmpfs_blk_file`). Would you like me to throw in a commit to align those as well?

KrissN: BTW, it would be worth to mention (for the future) on the [InterfaceNaming](https://github.com/SELinuxProject/refpolicy/wiki/InterfaceNaming) wiki page that predicates should be used in plural form.

policy/modules/system/systemd.te
pebenito: If you're not using the factory, then why would tmpfiles be acting on config files? Most general purpose systems have persistent config files and probably wouldn't be having these tmpfiles entries.

policy/modules/kernel/filesystem.if
pebenito: Thanks for note on the interface naming page. Fixing the inconsistency isn't necessary for the PR, but if you'd like to you fix that, it would be great. You'd need to have deprecated versions of the old interfaces (grep deprecated in the policy for examples).

pebenito (approved):

PR #355, opened 2021-03-05
selinux: Add a secure_mode_setbool Boolean.

### LINKED ISSUE #354: disable booleans
is there anyway to completely disable booleans support in refpolicy ?

thanks.

pebenito:
Enabling this will disable all permissions for setting SELinux Booleans,
even for unconfined domains.

This does not affect setenforce.  Enable `secure_mode_policyload` along with
`secure_mode_setbool` to fully lock the SELinux security interface.

Fixes #354

### DIFF
--- a/policy/modules/kernel/selinux.if
+++ b/policy/modules/kernel/selinux.if
@@ -392,14 +392,18 @@ interface(`selinux_read_policy',`
 interface(`selinux_set_generic_booleans',`
 	gen_require(`
 		type security_t;
+		bool secure_mode_setbool;
 	')
 
 	dev_search_sysfs($1)
 
 	allow $1 security_t:dir list_dir_perms;
-	allow $1 security_t:file rw_file_perms;
+	allow $1 security_t:file read_file_perms;
 
-	allow $1 security_t:security setbool;
+	if(!secure_mode_setbool) {
+		allow $1 security_t:file write_file_perms;
+		allow $1 security_t:security setbool;
+	}
 ')
 
 ########################################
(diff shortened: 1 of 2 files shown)


embexus:
@pebenito would be also possible to set a default state of **secure_mode_setbool** at build time ?

pebenito:
Use the `policy/booleans.conf` file. It is intended for users or distros to set default Boolean states without disturbing the policy source.

embexus:
@pebenito: I edited  `policy/booleans.conf ` to set `secure_mode_setbool = true` but after building the policy I found again :

`secure_mode_setbool = false`  !

I tried that with other booleans, this does not have any effect and it seems that  `policy/booleans.conf `  is just filled with booleasn default values.

embexus (approved):
looks good, can we merge it ?

embexus:
@pebenito I think there is a problem with with patch:

when using busybox setsebool/getsebool it's working

```
# busybox getsebool secure_mode_setbool
secure_mode_setbool --> on

# busybox getsebool user_ping 
user_ping --> off 
# busybox setsebool user_ping on
# busybox getsebool user_ping
user_ping --> off pending: on

````

But when using standard selinux getsebool/setsebool :

```
# getsebool secure_mode_setbool
secure_mode_setbool --> on

# getsebool user_ping 
user_ping --> off
#  setsebool user_ping on
# getsebool user_ping
user_ping --> on
```

**Comment by embexus** [comment]
tried also using sysfs instead of busybox :+1: 
```
$  cat /sys/fs/selinux/booleans/user_ping
0 1

$  echo 1 > /sys/fs/selinux/booleans/user_ping
$ cat /sys/fs/selinux/booleans/user_ping
0 1
$ setsebool user_ping on
$ cat /sys/fs/selinux/booleans/user_ping
1 1
```

**Comment by embexus** [comment]
I debugged the system calls using strace and found out that policycoreutils is trigering

 **/sys/fs/selinux/commit_pending_bools** 

```
$  echo 1 > /sys/fs/selinux/booleans/user_ping
$ cat /sys/fs/selinux/booleans/user_ping
0 1
$ echo 1 > /sys/fs/selinux/commit_pending_bools

$ cat /sys/fs/selinux/booleans/user_ping
1 1
```

PR #356, opened 2021-03-06
Remove additional unused modules

pebenito:
Remove more modules for programs that are no longer maintained.


etbe:
The mcelog policy might still be needed, I've found some other references to it being used last year. I now think that one should stay for the moment and we can review it next year.

PR #359, opened 2021-03-26
Many fixes for various applications and services

0xC0ncord:
This is a large collection of various fixes, appropriately broken up. Several of these commits include rules that I have found to be required in order for a systemd system to boot in enforcing. Many dontaudit rules are also in here (mostly getattr on `/proc`), scattered across the board.


policy/modules/admin/usbguard.te
```diff
@@ -27,6 +27,9 @@ logging_log_file(usbguard_log_t)
 type usbguard_rules_t;
 files_config_file(usbguard_rules_t)
 
+type usbguard_runtime_t;
+files_runtime_file(usbguard_runtime_t)
```
doverride, line 31: you should add a filecon to go with that

policy/modules/system/getty.te
```diff
@@ -123,3 +125,7 @@ optional_policy(`
 optional_policy(`
 	ppp_domtrans(getty_t)
 ')
+
+optional_policy(`
+	systemd_stream_connect_userdb(getty_t)
```
doverride, line 130: should probably just be `auth_use_nsswitch(getty_t)`

policy/modules/system/init.te
```diff
@@ -56,6 +56,12 @@ domain_entry_file(init_t, init_exec_t)
 kernel_domtrans_to(init_t, init_exec_t)
 role system_r types init_t;
 
+#
+# init_tmp_t is the type for /tmp/systemd-private-xxxx
+#
```
doverride, line 61: bad idea, unless you transition back to tmp_t on dirs with name "tmp"

ie:

`filetrans_pattern(init_t, init_tmp_t, tmp_t, dir, "tmp")`

You see these directories will be bind mounted to the service /tmp and that would then mean that any service that has PrivateTmp=yes and that maintains objects in "/tmp" will have to be able to write to i`nit_tmp_t` instead of `tmp_t`

policy/modules/system/init.te
```diff
@@ -294,6 +300,16 @@ ifdef(`init_systemd',`
...
 
+	allow init_t init_var_lib_t:dir manage_dir_perms;
+	allow init_t init_var_lib_t:file manage_file_perms;
+	allow init_t init_var_lib_t:lnk_file manage_lnk_file_perms;
+
+	allow init_t init_tmp_t:dir { manage_dir_perms relabel_dir_perms };
+	allow init_t init_tmp_t:file { manage_file_perms relabel_file_perms };
+	allow init_t init_tmp_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
+	allow init_t init_tmp_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
+	files_tmp_filetrans(init_t, init_tmp_t, { dir file })
```
doverride, line 312: this should only apply to dirs (not file, lnk_file, sock_file) AFAIK. also `relabel perms` shouldnt be needed

policy/modules/system/init.te
```diff
@@ -294,6 +300,16 @@ ifdef(`init_systemd',`
 	# /memfd:systemd-state
 	fs_tmpfs_filetrans(init_t, init_runtime_t, file)
 
+	allow init_t init_var_lib_t:dir manage_dir_perms;
+	allow init_t init_var_lib_t:file manage_file_perms;
+	allow init_t init_var_lib_t:lnk_file manage_lnk_file_perms;
```
doverride, line 306: this is probably for `/var/lib/systemd/timers`?

the lnk file is a bit questionable, but in theory I think it is a possible scenario for example `StateDirectory=systemd/timesync` in that case systemd will create the `/var/lib/systemd/timesync` symlink that points to `/var/lib/private/systemd/timesync`

policy/modules/kernel/kernel.te
```diff
@@ -347,6 +347,7 @@ files_list_root(kernel_t)
 files_list_etc(kernel_t)
 files_getattr_etc_runtime_dirs(kernel_t)
 files_mounton_etc_runtime_dirs(kernel_t)
+files_mounton_runtime_dirs(kernel_t)
```
doverride, line 350: is that `/run/initramfs`?

policy/modules/admin/bootloader.te
```diff
@@ -44,6 +44,7 @@ dev_node(bootloader_tmp_t)
 allow bootloader_t self:capability { chown dac_override dac_read_search fsetid mknod setgid sys_admin sys_rawio };
 allow bootloader_t self:process { signal_perms execmem };
 allow bootloader_t self:fifo_file rw_fifo_file_perms;
+allow bootloader_t self:netlink_selinux_socket create_socket_perms;
```
doverride, line 47: You can remove this

policy/modules/admin/bootloader.te
```diff
@@ -82,6 +84,8 @@ dev_dontaudit_rw_generic_dev_nodes(bootloader_t)
 dev_read_rand(bootloader_t)
 dev_read_urand(bootloader_t)
 dev_read_sysfs(bootloader_t)
+# for updating efi variables
+dev_write_sysfs(bootloader_t)
```
doverride, line 88: that doesnt make sense, efi variables are on efivarfs and not sysfs

policy/modules/kernel/devices.if
```diff
@@ -1937,6 +1937,25 @@ interface(`dev_rw_dlm_control',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`dev_filetrans_console_dev',`
+	gen_require(`
+		type device_t;
+	')
+
+	term_create_console($1, device_t)
```
doverride, line 1956: how about: `dev_filetrans($1, console_device_t, chr_file, "console")` instead?

policy/modules/system/init.te
```diff
@@ -468,6 +468,10 @@ ifdef(`init_systemd',`
 	selinux_compute_user_contexts(init_t)
 	selinux_use_status_page(init_t)
 
+	# systemd loads the policy itself instead of using load_policy
+	selinux_load_policy(init_t)
+	seutil_read_bin_policy(init_t)
+
```
doverride, line 482: Does not make sense though... when systemd loads policy, selinux is not even enabled yet, so it wouldnt need that permission. Why would systemd load_policy if selinux is already enabled? Seems questionable...

policy/modules/system/logging.te
```diff
@@ -234,6 +234,7 @@ optional_policy(`
 	init_list_unit_dirs(auditd_t)
 	systemd_start_power_units(auditd_t)
 	systemd_status_power_units(auditd_t)
+	systemd_stream_connect_userdb(auditd_t)
```
doverride, line 238: probably should just call `auth_use_nsswitch()` its resolving names which implies it reads /etc/passwd and /etc/nsswitch

policy/modules/system/init.te
```diff
@@ -445,6 +448,7 @@ ifdef(`init_systemd',`
 	# mount-setup
 	fs_unmount_autofs(init_t)
 	fs_getattr_pstore_dirs(init_t)
+	fs_create_pstore_dirs(init_t)
```
doverride, line 451: I wonder what this is all about:

```
[root@dssp4test2 dssp5]# mkdir /sys/fs/pstore/test
mkdir: cannot create directory ‘/sys/fs/pstore/test’: Operation not permitted
```

policy/modules/system/init.te
```diff
@@ -519,6 +523,9 @@ ifdef(`init_systemd',`
 	# for systemd to read udev status
 	udev_read_runtime_files(init_t)
 
+	udev_relabel_rules_dirs(init_t)
+	udev_relabel_rules_files(init_t)
+
```
doverride, line 528: I dont think systemd relabels /etc let alone /etc/udev?

policy/modules/system/udev.if
```diff
@@ -202,6 +202,50 @@ interface(`udev_manage_rules_files',`
...
+interface(`udev_relabel_rules_dirs',`
+	gen_require(`
+		type udev_rules_t;
+	')
+
+	relabel_dirs_pattern($1, udev_rules_t, udev_rules_t)
+
+	files_search_etc($1)
+
+	udev_search_runtime($1)
```
doverride, line 224: how is `/run/udev` related to this?

policy/modules/system/init.te
0xC0ncord: Certainly this access can be reworked. I added this primarily because I noticed some denials related to services with `PrivateTmp=yes` that look like the following:
```
time->Sat Mar 27 12:04:08 2021
type=AVC msg=audit(1616861048.420:31678): avc:  denied  { relabelfrom } for  pid=1 comm="systemd" name="tmp" dev="dm-9" ino=917506 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:tmp_t:s0 tclass=dir permissive=1
----
time->Sat Mar 27 12:04:08 2021
type=AVC msg=audit(1616861048.420:31679): avc:  denied  { relabelto } for  pid=1 comm="systemd" name="tmp" dev="dm-9" ino=917506 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:tmp_t:s0 tclass=dir permissive=1
```
Yet I'm not really sure what is happening here other than these are visible when services are started or stopped.

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -294,6 +300,16 @@ ifdef(`init_systemd',`
>  	# /memfd:systemd-state
>  	fs_tmpfs_filetrans(init_t, init_runtime_t, file)
>  
> +	allow init_t init_var_lib_t:dir manage_dir_perms;
> +	allow init_t init_var_lib_t:file manage_file_perms;
> +	allow init_t init_var_lib_t:lnk_file manage_lnk_file_perms;
> +
> +	allow init_t init_tmp_t:dir { manage_dir_perms relabel_dir_perms };
> +	allow init_t init_tmp_t:file { manage_file_perms relabel_file_perms };
> +	allow init_t init_tmp_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
> +	allow init_t init_tmp_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
> +	files_tmp_filetrans(init_t, init_tmp_t, { dir file })
>
> Certainly this access can be reworked. I added this primarily because I noticed some denials related
> to services with PrivateTmp=yes that look like the following:
>
> time->Sat Mar 27 12:04:08 2021
> type=AVC msg=audit(1616861048.420:31678): avc:  denied  { relabelfrom } for  pid=1 comm="systemd" name="tmp" dev="dm-9" ino=917506 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:tmp_t:s0 tclass=dir permissive=1
> ----
> time->Sat Mar 27 12:04:08 2021
> type=AVC msg=audit(1616861048.420:31679): avc:  denied  { relabelto } for  pid=1 comm="systemd" name="tmp" dev="dm-9" ino=917506 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:tmp_t:s0 tclass=dir permissive=1
>
> Yet I'm not really sure what is happening here other than these are visible when services are started
> or stopped.

ok so it relabels the /tmp/systemd-private-XXXX/tmp dir

That shouldnt be a problem i guess

You can transition to init_tmp_t on tmp_t dirs and then transition back
to tmp on dirs with name "tmp" in dirs with type init_tmp_t

So systemd-private-XXXX can be init_tmp_t, but systemd-private-XXXX/tmp
should be tmp_t again otherwise you break every service with
PrivateTmp=yes

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

policy/modules/kernel/devices.if
0xC0ncord: I did it this way to avoid requiring the `console_device_t` type which belongs to the `terminal` module instead, though I much prefer keeping all the required access in this interface if possible.

policy/modules/system/init.te
0xC0ncord: I only really see these accesses on `/var/lib/systemd/timers`, yes, though I'm not sure if they might occur elsewhere there. In my case I only ever observed systemd trying to `setattr` on the timer files in order to update their timestamps.

doverride:

>
>
> In policy/modules/kernel/devices.if:
>
>> +## <summary>
> +##	Automatic type transition to the type
> +##	for the console device node when created in /dev.
> +## </summary>
> +## <param name="domain">
> +##	<summary>
> +##	Domain allowed access.
> +##	</summary>
> +## </param>
> +#
> +interface(`dev_filetrans_console_dev',`
> +	gen_require(`
> +		type device_t;
> +	')
> +
> +	term_create_console($1, device_t)
>
> I did it this way to avoid requiring the console_device_t type which belongs to the terminal module
> instead, though I much prefer keeping all the required access in this interface if possible.

Right and that makes sense but why didnt you use dev_filetrans() in term_create_console()?

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -294,6 +300,16 @@ ifdef(`init_systemd',`
>  	# /memfd:systemd-state
>  	fs_tmpfs_filetrans(init_t, init_runtime_t, file)
>  
> +	allow init_t init_var_lib_t:dir manage_dir_perms;
> +	allow init_t init_var_lib_t:file manage_file_perms;
> +	allow init_t init_var_lib_t:lnk_file manage_lnk_file_perms;
>
> I only really see these accesses on /var/lib/systemd/timers, yes, though I'm not sure if they might
> occur elsewhere there. In my case I only ever observed systemd trying to setattr on the timer files in
> order to update their timestamps.

Yes but, but systemd maintains those timer files so the above is ok

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

0xC0ncord:
> Right and that makes sense but why didnt you use dev_filetrans() in term_create_console()?

Ah, that would be a mistake yes. Fixed.

policy/modules/system/init.te
0xC0ncord: I think this might be related to using the dracut SELinux module which tries to load the policy much earlier in the boot process. Still investigating though.

doverride:
> ok so it relabels the /tmp/systemd-private-XXXX/tmp dir That shouldnt be a problem i guess You can transition to init_tmp_t on tmp_t dirs and then transition back to tmp on dirs with name "tmp" in dirs with type init_tmp_t So systemd-private-XXXX can be init_tmp_t, but systemd-private-XXXX/tmp should be tmp_t again otherwise you break every service with PrivateTmp=yes

Oh i think i have an idea as to why it might relabel it and why i did not see that before.

There might be a filecon interfering:

`matchpatchcon /tmp/systemd-private-XXXX/tmp`

that *should* return `<<none>>` and not `system_u:object_r:tmp_t`

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -468,6 +468,10 @@ ifdef(`init_systemd',`
>  	selinux_compute_user_contexts(init_t)
>  	selinux_use_status_page(init_t)
>  
> +	# systemd loads the policy itself instead of using load_policy
> +	selinux_load_policy(init_t)
> +	seutil_read_bin_policy(init_t)
> +
>
> I think this might be related to using the dracut SELinux module which tries to load the policy much
> earlier in the boot process. Still investigating though.

Right, disable the module

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

doverride:
> Right, disable the module

It's old and it shouldnt be used on systemd systems. You dont want to allow systemd to load policy ideally.

doverride:
I guess that might also have been what causes your "/dev/console" labeling issue but never mind that.

0xC0ncord:
> > ok so it relabels the /tmp/systemd-private-XXXX/tmp dir That shouldnt be a problem i guess You can transition to init_tmp_t on tmp_t dirs and then transition back to tmp on dirs with name "tmp" in dirs with type init_tmp_t So systemd-private-XXXX can be init_tmp_t, but systemd-private-XXXX/tmp should be tmp_t again otherwise you break every service with PrivateTmp=yes
> 
> Oh i think i have an idea as to why it might relabel it and why i did not see that before.
> 
> There might be a filecon interfering:
> 
> `matchpathcon /tmp/systemd-private-XXXX/tmp`
> 
> that _should_ return `<<none>>` and not `system_u:object_r:tmp_t`

Looks like the fix then is a small change to `/tmp/systemd-private-[^/]+/tmp/.*	<<none>>` in `files.fc` then. If that's the case then perhaps the `init_tmp_t` type might not be needed at all. I will verify this.

policy/modules/system/init.te
0xC0ncord: Likewise, this is probably related to using the `udev-rules` dracut module.

doverride:
Right but the `init_tmp_t` on `/tmp/systemd-private-XXXX` dirs is still kind of neat to have though because that can be used to block "others" access to the private `/tmp` dirs. Its a bit weird though, and so I will leave whatever you choose to do here to others to judge

policy/modules/system/init.te
0xC0ncord: Honestly not sure either, but it comes from this denial during boot:
```
audit: type=1400 audit(1616864322.098:3): avc:  denied  { create } for  pid=1 comm="systemd" name="pstore" scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:pstore_t:s0 tclass=dir permissive=1
```
Also to note that this is after I have updated my dracut configuration such as to not include the selinux or udev-rules modules.

0xC0ncord:
> Right but the `init_tmp_t` on `/tmp/systemd-private-XXXX` dirs is still kind of neat to have though because that can be used to block "others" access to the private `/tmp` dirs. Its a bit weird though, and so I will leave whatever you choose to do here to others to judge

That's a good point. More than likely I will leave it.

doverride:
Whatever you choose to do though: to ensure that the PrivateTmp dir is labeled properly do:
```
systemd-run -p PrivateTmp=yes -t /bin/sh
....
ls -alZ /tmp
```
it should be `tmp_t`

policy/modules/system/init.te
0xC0ncord: Hm. Even after removing that module I still see this denial:
```
audit: type=1400 audit(1616864322.185:5): avc:  denied  { relabelto } for  pid=1 comm="systemd" name="rules.d" dev="tmpfs" ino=9 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:udev_rules_t:s0 tclass=dir permissive=1
```

policy/modules/system/udev.if
0xC0ncord: Looking over it again, I don't think it is. That access was copied from the above interfaces when I made these new ones. Notwithstanding the above (whether these interfaces are needed at all), this access can be removed.

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -519,6 +523,9 @@ ifdef(`init_systemd',`
>  	# for systemd to read udev status
>  	udev_read_runtime_files(init_t)
>  
> +	udev_relabel_rules_dirs(init_t)
> +	udev_relabel_rules_files(init_t)
> +
>
> Likewise, this is probably related to using the udev-rules dracut module.

okay then, I guess ...
>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -445,6 +448,7 @@ ifdef(`init_systemd',`
>  	# mount-setup
>  	fs_unmount_autofs(init_t)
>  	fs_getattr_pstore_dirs(init_t)
> +	fs_create_pstore_dirs(init_t)
>
> Honestly not sure either, but it comes from this denial during boot:

Do you have a tmpfs mounted on /sys/fs?

its trying to mkdir /sys/fs/pstore but unless there is a tmpfs mounted
on /sys/fs it wouldnt even be able to create that dir on the sysfs

I think , we have little choice but to allow this

>
> audit: type=1400 audit(1616864322.098:3): avc:  denied  { create } for  pid=1 comm="systemd" name="pstore" scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:pstore_t:s0 tclass=dir permissive=1
>
> Also to note that this is after I have updated my dracut configuration such as to not include the
> selinux or udev-rules modules.
>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

policy/modules/kernel/kernel.te
0xC0ncord: Yes, though I will probably remove this access since after removing the selinux dracut module, it seems this is no longer needed.

doverride:

>
>
> In policy/modules/system/init.te:
>
>> @@ -519,6 +523,9 @@ ifdef(`init_systemd',`
>  	# for systemd to read udev status
>  	udev_read_runtime_files(init_t)
>  
> +	udev_relabel_rules_dirs(init_t)
> +	udev_relabel_rules_files(init_t)
> +
>
> Hm. Even after removing that module I still see this denial:

Okay then i guess it should just be allowed. Not sure what is causing
this but the avc denial is pretty conclusive: its really pid1
>
> audit: type=1400 audit(1616864322.185:5): avc:  denied  { relabelto } for  pid=1 comm="systemd" name="rules.d" dev="tmpfs" ino=9 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:udev_rules_t:s0 tclass=dir permissive=1
>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

0xC0ncord:
> Do you have a tmpfs mounted on /sys/fs?

I do not.

> its trying to mkdir /sys/fs/pstore but unless there is a tmpfs mounted on /sys/fs it wouldnt even be able to create that dir on the sysfs I think , we have little choice but to allow this

OK.

doverride:

>
>
> In policy/modules/kernel/kernel.te:
>
>> @@ -347,6 +347,7 @@ files_list_root(kernel_t)
>  files_list_etc(kernel_t)
>  files_getattr_etc_runtime_dirs(kernel_t)
>  files_mounton_etc_runtime_dirs(kernel_t)
> +files_mounton_runtime_dirs(kernel_t)
>
> Yes, though I will probably remove this access since after removing the selinux dracut module, it
> seems this is no longer needed.

Aside from that, I would probably associate a private type with the
/run/initramfs dir

(filecon "/run/initramfs" dir (system_u object_r
dracut_restore_initramfs_t ((s0)(s0))))
(filecon "/run/initrafs/.*" any ())

ie. only label the dir and everything below it can be <<none>> (i think)

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

doverride:

>  Do you have a tmpfs mounted on /sys/fs?
>
> I do not.
>
>  its trying to mkdir /sys/fs/pstore but unless there is a tmpfs mounted on /sys/fs it wouldnt even
>  be able to create that dir on the sysfs I think , we have little choice but to allow this
>
> OK.

Does not make sense though as you cannot create dirs on sysfs...

Maybe its a bug

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

0xC0ncord:
> Does not make sense though as you cannot create dirs on sysfs... Maybe its a bug

I'm thinking similarly. I will drop this access temporarily from my working tree and try booting in enforcing again and see what happens. If I can boot fine, I will drop this access from this PR and we can look into it afterwards.

0xC0ncord:
> I guess that might also have been what causes your "/dev/console" labeling issue but never mind that.

Looks like this was the case, so I will drop this as well.

0xC0ncord:
> Aside from that, I would probably associate a private type with the /run/initramfs dir (filecon "/run/initramfs" dir (system_u object_r dracut_restore_initramfs_t ((s0)(s0)))) (filecon "/run/initrafs/.*" any ()) ie. only label the dir and everything below it can be <<none>> (i think)

I'm trying to find an appropriate place for something like this. The dracut module is actually part of the gentoo policy and not refpolicy. The files module has a file context for `/initrd` as `root_t`, but not `/run/initramfs`, so I'm thinking that maybe it should be labeled `root_t` instead.

doverride:

>  Aside from that, I would probably associate a private type with the /run/initramfs dir (filecon
>  "/run/initramfs" dir (system_u object_r dracut_restore_initramfs_t ((s0)(s0)))) (filecon
>  "/run/initrafs/.*" any ()) ie. only label the dir and everything below it can be <> (i think)
>
> I'm trying to find an appropriate place for something like this. The dracut module is actually part of
> the gentoo policy and not refpolicy. The files module has a file context for /initrd as root_t, but
> not /run/initramfs, so I'm thinking that maybe it should be labeled root_t instead.

/run/initramfs is used my dracut-initramfs-restore (i guess refpolicy
has not targeted that service yet?)

https://git.defensec.nl/?p=dssp3.git;a=blob_plain;f=policy/system/dracutinitramfs.cil;hb=HEAD


>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

0xC0ncord:
> /run/initramfs is used my dracut-initramfs-restore (i guess refpolicy has not targeted that service yet?) https://git.defensec.nl/?p=dssp3.git;a=blob_plain;f=policy/system/dracutinitramfs.cil;hb=HEAD

If that's the case I think that change is better suited for its own PR, though I will definitely take a look at it in the future.

policy/modules/admin/bootloader.te
0xC0ncord: The comment is probably wrong, and I'm not really sure what to make of the actual denial:
```
time->Sun Mar 14 16:42:55 2021
type=PROCTITLE msg=audit(1615754575.219:5887): proctitle=656669626F6F746D6772002D71002D620030303030002D42
type=PATH msg=audit(1615754575.219:5887): item=0 name="/sys/firmware/efi/vars/del_var" inode=20936 dev=00:0e mode=0100200 ouid=0 ogid=701 rdev=00:00 obj=system_u:object_r:sysfs_t:s0 nametype=NORMAL cap_fp=0 cap_fi=0 cap_fe=0 cap_fver=0 cap_frootid=0
type=CWD msg=audit(1615754575.219:5887): cwd="/root"
type=SYSCALL msg=audit(1615754575.219:5887): arch=c000003e syscall=257 success=no exit=-13 a0=ffffff9c a1=7b17a7cc9f60 a2=1 a3=0 items=1 ppid=228501 pid=228523 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=pts0 ses=11 comm="efibootmgr" exe="/usr/sbin/efibootmgr" subj=staff_u:sysadm_r:bootloader_t:s0-s0:c0.c1023 key="access"
type=AVC msg=audit(1615754575.219:5887): avc:  denied  { write } for  pid=228523 comm="efibootmgr" name="del_var" dev="sysfs" ino=20936 scontext=staff_u:sysadm_r:bootloader_t:s0-s0:c0.c1023 tcontext=system_u:object_r:sysfs_t:s0 tclass=file permissive=0
```

policy/modules/system/init.te
0xC0ncord: After looking at this again I found that in my case, I can still boot in enforcing without this access.

policy/modules/admin/bootloader.te
dkjii-g: /sys/firmware/efi/vars might just be mislabeled, I guess that depends on the context in fstab

policy/modules/system/init.te
dkjii-g: You should copy fedora for systemd or dracut when fixing selinux denials

policy/modules/kernel/devices.if
```diff
@@ -3391,6 +3391,25 @@ interface(`dev_setattr_null_dev',`
...
 ')
 
+########################################
+## <summary>
+##	Do not audit attempts to set the attributes of
+##	the null device nodes.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
```
pebenito, line 3401: Incorrect summary.

policy/modules/system/init.te
```diff
@@ -460,6 +490,8 @@ ifdef(`init_systemd',`
 	auth_relabel_login_records(init_t)
 	auth_relabel_pam_console_data_dirs(init_t)
 	auth_domtrans_chk_passwd(init_t)
+	# for systemd dynamic users
+	auth_rw_shadow(init_t)
```
pebenito, line 494: Needs to be conditional.

policy/modules/system/init.te
```diff
@@ -56,6 +56,12 @@ domain_entry_file(init_t, init_exec_t)
 kernel_domtrans_to(init_t, init_exec_t)
 role system_r types init_t;
 
+#
+# init_tmp_t is the type for /tmp/systemd-private-xxxx and /tmp/namespace-dev-xxxx
+#
+type init_tmp_t;
```
pebenito, line 62: I agree with Dominick, I'm not convinced this is a good change.

policy/modules/system/systemd.te
```diff
@@ -1264,6 +1285,21 @@ seutil_read_file_contexts(systemd_sessions_t)
...
+########################################
+#
+# sysctl local policy
+#
+
+dontaudit systemd_sysctl_t self:capability sys_ptrace;
+
+kernel_read_kernel_sysctls(systemd_sysctl_t)
+kernel_request_load_module(systemd_sysctl_t)
+kernel_rw_all_sysctls(systemd_sysctl_t)
```
pebenito, line 1297: This seems unnecessary, why not let it run in `initrc_t`, which is where non-systemd systems will typically be setting sysctls?

pebenito (changes requested):
Sorry for the review delay.

policy/modules/system/init.te
0xC0ncord: OK. I will remove this type and grant the necessary relabel rules instead.

policy/modules/system/systemd.te
0xC0ncord: In my testing I was getting several AVCs where `initrc_t` was not permitted to set some types of sysctls. I thought to just grant that access to `initrc_t` but felt that a new domain made more sense, and that's also how the Fedora policy handled this. I'm fine with reverting to the former however.

doverride:

>
> --------------------------------------------------------------------------------------------------
>
> In policy/modules/system/systemd.te:
>
>> @@ -1264,6 +1285,21 @@ seutil_read_file_contexts(systemd_sessions_t)
>  
>  systemd_log_parse_environment(systemd_sessions_t)
>  
> +########################################
> +#
> +# sysctl local policy
> +#
> +
> +dontaudit systemd_sysctl_t self:capability sys_ptrace;
> +
> +kernel_read_kernel_sysctls(systemd_sysctl_t)
> +kernel_request_load_module(systemd_sysctl_t)
> +kernel_rw_all_sysctls(systemd_sysctl_t)
>
> In my testing I was getting several AVCs where initrc_t was not permitted to set some types of
> sysctls. I thought to just grant that access to initrc_t but felt that a new domain made more
> sense, and that's also how the Fedora policy handled this. I'm fine with reverting to the former
> however.

should also be noted that systemd-sysctl can also be run from the
command line, i suppose as a alternative to sysctl

from man systemd-sysctl:

Example 3. Update coredump handler configuration

# /usr/lib/systemd/systemd-sysctl --prefix kernel.core_pattern

This searches all the directories listed in sysctl.d(5) for configuration files and writes /proc/sys/kernel/core_pattern.
>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

policy/modules/admin/bootloader.te
0xC0ncord: Hmm. Even in Fedora's policy these are `sysfs_t`. I'm thinking I will drop this access for now until I can investigate further.

doverride:

>
> --------------------------------------------------------------------------------------------------
>
> In policy/modules/admin/bootloader.te:
>
>> @@ -82,6 +84,8 @@ dev_dontaudit_rw_generic_dev_nodes(bootloader_t)
>  dev_read_rand(bootloader_t)
>  dev_read_urand(bootloader_t)
>  dev_read_sysfs(bootloader_t)
> +# for updating efi variables
> +dev_write_sysfs(bootloader_t)
>
> Hmm. Even in Fedora's policy these are sysfs_t. I'm thinking I will drop this access for now until
> I can investigate further.

On Fedora Rawhide (with my own policy):

***@***.*** ~]# ls -dZ `mount -t efivarfs | awk '{ print $3 }'`
sys.id:sys.role:efivar.fs:s0 /sys/firmware/efi/efivars

efivarfs is not a seclabel fs, and so it cannot be relabeled to sys.fs


>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

0xC0ncord:
> On Fedora Rawhide (with my own policy): ***@***.*** ~]# ls -dZ `mount -t efivarfs | awk '{ print $3 }'` sys.id:sys.role:efivar.fs:s0 /sys/firmware/efi/efivars efivarfs is not a seclabel fs, and so it cannot be relabeled to sys.fs

OK that makes sense

0xC0ncord:
After some discussion, the previous access for systemd to write to `shadow_t` I removed, pending a rework that I feel is outside the scope of this PR for the moment. I will probably section that off into its own patch because the scope could be fairly large.

0xC0ncord:
> After some discussion, the previous access for systemd to write to `shadow_t` I removed, pending a rework that I feel is outside the scope of this PR for the moment. I will probably section that off into its own patch because the scope could be fairly large.

After actually implementing this and testing it, the scope turned out to not be as large as I initially thought, and so I've included it in this PR.

policy/modules/system/authlogin.te
```diff
@@ -21,6 +21,9 @@ attribute nsswitch_domain;
 type auth_cache_t;
 logging_log_file(auth_cache_t)
 
+type auth_lock_t;
```
pebenito, line 24: The implication seems to be the reason for this to exist is because systemd will create the lock file on its own even though it never changes /etc/shadow?

policy/modules/system/init.te
```diff
@@ -398,6 +408,7 @@ ifdef(`init_systemd',`
 	# If /etc/localtime is missing, a watch on /etc is added.
 	files_watch_etc_dirs(init_t)
 	files_watch_etc_symlinks(init_t)
+	files_dontaudit_write_var_dirs(init_t)
```
pebenito, line 411: Are you sure this is ok? I seem to remember this happening when systemd is attempting to determine if the location is writable.

policy/modules/system/authlogin.te
0xC0ncord: That's correct. For some reason systemd needs to be able to write to `.pwd.lock`, else a service using dynamic users will fail to start. I found that once this access is granted, systemd itself never actually accesses `/etc/shadow` or `/etc/passwd`. Not sure why this is the case, but this fixes the problem.

policy/modules/system/systemd.te
```diff
@@ -610,6 +621,7 @@ allow systemd_logind_t systemd_sessions_runtime_t:dir manage_dir_perms;
 allow systemd_logind_t systemd_sessions_runtime_t:file manage_file_perms;
 allow systemd_logind_t systemd_sessions_runtime_t:fifo_file manage_fifo_file_perms;
 
+kernel_dontaudit_getattr_proc(systemd_logind_t)
```
pebenito, line 624: This might make sense to be part of systemd_log_parse_environment. Did you check if systemd was throwing errors over this? I've tried several `dontaudits` for systemd and they rarely seem to work

policy/modules/system/init.te
0xC0ncord: I'm not sure if this is actually required or not. In my testing with this dontaudited I haven't seen any problems, but I see no objections to allowing this access either.

policy/modules/system/systemd.te
0xC0ncord: In my testing I haven't seen any errors from this, but I don't see a problem with simply allowing the access either.

policy/modules/system/authlogin.te
pebenito: Here's the related code: https://github.com/systemd/systemd/blob/fee6441601c979165ebcbb35472036439f8dad5f/src/core/dynamic-user.c#L419

I'd prefer this called `shadow_lock_t` so you can see it is related to `shadow_t`. Then the `auth_manage_shadow()` and `auth_rw_shadow()` can have the rules for doing the lock too, so we don't have to call it everywhere we call those interfaces.

0xC0ncord: OK. That makes sense.

policy/modules/admin/bootloader.te
dkjii-g: efivars is mounted so fedora might have the sysfs_t label on it, but mount it with context=efivarfs (I'd have to check and dont have a EFI fedora VM rn)

pebenito: I see `genfscon efivarfs /  system_u:object_r:efivarfs_t:s0` in the Fedora policy, and it looks like something we need in refpolicy.

pebenito:
I think I'm ok with everything, pending resolution of the 3 comment threads here.

0xC0ncord:
> I think I'm ok with everything, pending resolution of the 3 comment threads here.

OK. I have changed the name of `auth_lock_t` to `shadow_lock_t` as discussed per the first comment. Can you clarify whether you think access in the second comment should be granted (write on `var_t` by init), as well as the third (change to allow `getattr` on proc)?

policy/modules/system/init.te
pebenito: I've seen this denial on a yocto-based system, but it has been a while. I assume it's valid for allow. It may be some syscall or other kernel interface that is triggering a dir creation in pstore. I haven't dug into it.

pebenito:
That's all fine, I resolved the comment threads.

policy/modules/kernel/kernel.te
0xC0ncord: Removing the SELinux dracut module in my testing environment solved this denial.

policy/modules/admin/bootloader.te
0xC0ncord: OK. I can add that here.

0xC0ncord: But it seems that's already in refpolicy? https://github.com/SELinuxProject/refpolicy/blob/master/policy/modules/kernel/filesystem.te#L106

policy/modules/system/init.te
0xC0ncord: OK. I can allow the access again.

policy/modules/admin/bootloader.te
pebenito: My mistake, I must have looked at the wrong branch.

Gentoo isn't using efivarfs?

pebenito: @perfinion do you have any input for this thread?

0xC0ncord: I should clarify: this was added because in testing I found that `grub-mkconfig` would try to write to files in `/sys/firmware/efi/varfs` which is `sysfs_t`. Apparently this is different from `/sys/firmware/efi/efivars` which is actually where the `efivarfs` filesystem lives. I found that this is also the case on RHEL. Not sure what the implications of that are.

pebenito: Ah. The sysfs entries is the old way. efivarfs is the new way. I see this in grub:
```
  /*
   * Linux uses efivarfs (mounted on /sys/firmware/efi/efivars) to access the
   * EFI variable store. Some legacy systems may still use the deprecated
   * efivars interface (accessed through /sys/firmware/efi/vars). Where both
   * are present, libefivar will use the former in preference, so attempting
   * to load efivars will not interfere with later operations.
   */
```
The change is over 2 years old, so It seems this sysfs access should be dontaudited. Or if Gentoo doesn't have a new enough grub, then the sysfs write access should live in the Gentoo policy.

0xC0ncord: I'm sure that Gentoo has a new enough Grub, so maybe this is some sort of fallback behavior that I am triggering somehow in my environment. Either way, the above change isn't in this PR anymore, but I can `dontaudit` it as you say.

pebenito (approved):

PR #360, opened 2021-04-08
Support the fingerprint daemon, fix gpg_agent and pcscd modules, new kernel USB device interface and rules, updated evolution, kernel, java, smartmon and spamassassin modules, fix build breakage and lint checks

gtrentalancia:
Support the fingerprint daemon, including optional authentication
using xscreensaver (requires PAM module and PAM configuration).

policy/modules/apps/xscreensaver.te |   17 ++++++++++++++++-
policy/modules/services/fprintd.te  |    9 +++++++++
policy/modules/system/userdomain.if |    4 ++++
3 files changed, 29 insertions(+), 1 deletion(-)
---
Introduce permissions required for gpg_agent functionality.

policy/modules/apps/gpg.te |    2 ++
1 file changed, 2 insertions(+)
---
Fix the refpolicy build currently broken in the following systemd
interfaces commit:

  commit cc8374fd24129a2a20669bda2b57d8b029945047
  Date:   Sat Mar 13 18:22:59 2021 -0500
  various: systemd user fixes and additional support

Cause: duplicate type definitions introduced by new template
            systemd_user_daemon_domain()

Example error:
  /usr/bin/checkmodule -m tmp/staff.tmp -o tmp/staff.mod
  /usr/bin/checkmodule:  loading policy configuration from tmp/staff.tmp
  policy/modules/roles/staff.te:75:ERROR 'duplicate declaration of type/attribute' at token ';' on line 50011:
  #line 75
        type staff_systemd_t;
  /usr/bin/checkmodule:  error(s) encountered while parsing configuration

policy/modules/system/systemd.if |    3 ++-
1 file changed, 2 insertions(+), 1 deletion(-)
---
Fix the pcscd daemon module by enabling detection of USB smart card
readers using libudev.
    
policy/modules/services/pcscd.te |    3 +++
1 file changed, 3 insertions(+)
---
New device interface and allow rules for the kernel module handling
USB devices plugging/unplugging.
    
policy/modules/kernel/devices.if |   18 ++++++++++++++++++
policy/modules/kernel/kernel.te |    3 +++
2 file changed, 21 insertions(+)
---
Allow running Java applications that rely on the pcscd daemon (e.g.
to use smart-card applications).

policy/modules/apps/java.te |    4 ++++
1 file changed, 4 insertions(+)
---
Let the Java domain execute ldconfig.

policy/modules/apps/java.te |    2 ++
1 file changed, 2 insertions(+)
---
Edit the kernel module to make storage permissions related to
fixed-disk devices and other common storage devices available
on all systems rather than only on systems using systemd.
    
Add permissions from other available storage device interfaces.

policy/modules/kernel/kernel.te |   19 +++++++++++++------
1 file changed, 13 insertions(+), 6 deletions(-)
---
Let evolution connect to the smart card daemon.

policy/modules/contrib/evolution.te |    4 ++++
1 file changed, 4 insertions(+)
---
Add new gpg interfaces for gpg_agent execution and to avoid
auditing search/read operations on files and directories that
are not strictly needed and might pose a security risk.
    
The new interfaces will be used in a forthcoming update to
allow fetching updates from the network for the fsdaemon
drive database.

policy/modules/apps/gpg.if |  147 +++++++++++++++++++++++++++++++++++++++++++++
1 file changed, 147 insertions(+)
---
Enable the drive database update feature for the fsdaemon: this
is needed to support new drives and/or new drive features.
    
The feature is controlled by a boolean for flexibility.
    
Execution of the update task as a cron job is also supported.

The file contexts for the data directory is updated to include
the standard drive database file location under /usr in addition
to the existing location under /var/lib.

policy/modules/roles/staff.te       |    4 +
policy/modules/roles/sysadm.te      |    1 
policy/modules/services/smartmon.fc |    2 
policy/modules/services/smartmon.if |   43 +++++++++--
policy/modules/services/smartmon.te |  133 ++++++++++++++++++++++++++++++++++++
5 files changed, 176 insertions(+), 7 deletions(-)
---
Intoduce a new interface in the mta module to manage the mail
transport agent configuration directories and files.

This interface will be used by a forthcoming update of the
rule updating feature of the spamassassin module.

policy/modules/services/mta.if |   21 +++++++++++++++++++++
1 file changed, 21 insertions(+)
---
Update the spamassassin module in order to better support
the rules updating script.

The rules update feature is now controlled by a boolean
for increased flexibility.

The specific file type for the spamassassin update feature
temporary files has been removed: just use spamd_tmp_t instead
of spamd_update_tmp_t.

policy/modules/roles/staff.te           |    1
policy/modules/roles/sysadm.te          |    1
policy/modules/services/spamassassin.if |   75 ++++++++++++++++++++++-
policy/modules/services/spamassassin.te |  101 ++++++++++++++++++++++++--------
4 files changed, 151 insertions(+), 27 deletions(-)


pebenito:
There's been several pushes. Is this ready for review? If not, please switch to draft until it is ready.

gtrentalancia:
There might be a few additional refinements, but for the time being, it's mostly complete.

So, please go ahead with the review Chris !

policy/modules/services/spamassassin.te
```diff
@@ -21,6 +21,15 @@ gen_tunable(spamassassin_can_network, false)
...
 gen_tunable(spamd_enable_home_dirs, false)
 
+## <desc>
+##	<p>
+##	Determine whether spamassassin
+##	can update the rules using th
+##	network.
+##	</p>
+## </desc>
+gen_tunable(spamassassin_can_update, true)
```
pebenito, line 31: I think `spamassassin_net_update` or `spamassassin_network_update` would be clearer naming.

policy/modules/services/spamassassin.te
```diff
@@ -95,6 +98,16 @@ files_tmpfs_file(spamd_tmpfs_t)
 type spamd_var_lib_t;
 files_type(spamd_var_lib_t)
 
+type spamd_update_t;
+type spamd_update_exec_t;
+application_domain(spamd_update_t, spamd_update_exec_t)
+role spamd_update_roles types spamd_update_t;
+
+type spamd_update_gpg_t;
```
pebenito, line 106: Last year @etbe argued that this was not necessary, which is why the gpg use was removed, please explain the need.

policy/modules/services/smartmon.te
```diff
@@ -13,10 +13,30 @@ policy_module(smartmon, 1.18.0)
...
 type fsdaemon_t;
 type fsdaemon_exec_t;
 init_daemon_domain(fsdaemon_t, fsdaemon_exec_t)
 
+type fsdaemon_update_t;
+type fsdaemon_update_exec_t;
+application_domain(fsdaemon_update_t, fsdaemon_update_exec_t)
+role fsdaemon_update_roles types fsdaemon_update_t;
+
+type fsdaemon_update_gpg_t;
```
pebenito, line 35: Please explain the gpg use here too.

policy/modules/apps/gpg.if
```diff
@@ -161,6 +161,97 @@ interface(`gpg_entry_type',`
...
+##	</summary>
+## </param>
+#
+interface(`gpg_agent_spec_domtrans',`
+	gen_require(`
+		type gpg_agent_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domain_auto_transition_pattern($1, gpg_agent_exec_t, $2)
```
pebenito, line 233: Use `domtrans_pattern`

policy/modules/system/systemd.if
```diff
@@ -40,7 +40,6 @@ template(`systemd_role_template',`
 	#
 	# Declarations
 	#
-	type systemd_user_session_type, systemd_log_parse_env_type;
```
pebenito: If the lint is coming up as this being an error, then the linter has a bug. See its use in `systemd.te`.

policy/modules/system/systemd.if
```diff
@@ -600,6 +600,25 @@ interface(`systemd_log_parse_environment',`
...
+##   Make the specified type usable as a
+##   systemd user session type.
+## </summary>
+## <param name="domain">
+##   <summary>
+##     Type to be used as a systemd user session type.
+##   </summary>
+## </param>
+#
+interface(`systemd_systemd_user_session',`
```
pebenito, line 613: I'm not clear why one would need this except since it that is what `systemd_role()` is for?

policy/modules/kernel/kernel.te
```diff
@@ -374,13 +374,20 @@ ifdef(`init_systemd',`
...
+')
 
-	optional_policy(`
-		storage_dev_filetrans_fixed_disk(kernel_t)
-		storage_setattr_fixed_disk_dev(kernel_t)
-		storage_create_fixed_disk_dev(kernel_t)
-		storage_delete_fixed_disk_dev(kernel_t)
-	')
+optional_policy(`
+	storage_dev_filetrans_fixed_disk(kernel_t)
```
pebenito, line 380: What is the non-systemd usage scenario?

pebenito (changes requested):

policy/modules/services/spamassassin.te
gtrentalancia: I can amend that, if you prefer an alternative naming

policy/modules/services/spamassassin.te
gtrentalancia: gpg is needed to verify the signatures on downloaded spamassassin rules !

without being able to run gpg (with a domain transition **BUT** without being able to access gpg_secret_t files and dirs), spamassassin is not able to verify the signature of new rules downloaded from the Internet and therefore the update fails, as it won't install anything without being able to verify the signature !

gpg signature verification is vital to update both spamassassin classification rules and smartctl drive database (see other update), even though they are not executable or binary files, it's always a good idea to verify that the downloaded files are authentic and genuine (malicious files can still cause DoS and other service disruptions) !

policy/modules/services/smartmon.te
gtrentalancia: ditto

policy/modules/apps/gpg.if
gtrentalancia: Yes, will update that.

policy/modules/system/systemd.if
gtrentalancia: It's the lint from github - the automated lint check

gtrentalancia: Probably the correction is:

@@ -27,7 +27,7 @@
 #
 template(`systemd_role_template',`
        gen_require(`
-               attribute systemd_user_session_type, systemd_log_parse_env_type;
+               attribute systemd_user_session_type;
                attribute systemd_user_activated_sock_file_type, systemd_user_unix_stream_activated_socket_type;
                type $1_systemd_t;
                type systemd_run_exec_t, systemd_analyze_exec_t;

gtrentalancia: And then I decided to remove the type declaration which seems uneeded and reedundant with the attribute declaration above...

gtrentalancia: It's not being used as a type in that interface. The use case is to use the attribute in "typeattribute" rules in other interfaces such as the one that I added in that same file.

gtrentalancia: And since you referred to systemd.te, then that confirms that it is being declared only as an attribute even there. So, the latter change that I introduced, although possibly not stricltly necessary to shut up the linter, does make sense, improves code clarity to me.

policy/modules/system/systemd.if
gtrentalancia: Please see the use of sister interface systemd_log_parse_environment() in systemd.te...

policy/modules/kernel/kernel.te
gtrentalancia: The usage scenario is other init daemons in general, see for example SysVInit:

http://www.nongnu.org/sysvinit
https://git.savannah.nongnu.org/cgit/sysvinit.git

gtrentalancia:
It will take me about 24 hours to introduce the proposed changes, in the meanwhile I strongly suggest that you push the following critical commit with urgency:

https://github.com/SELinuxProject/refpolicy/pull/360/commits/716093aba808fbd53c1c2647a7e98489eb500bd2

as it's not just a lint failure, but a proper build failure at least since the following commit triggered the build failure:

  commit cc8374fd24129a2a20669bda2b57d8b029945047
  Date:   Sat Mar 13 18:22:59 2021 -0500
  various: systemd user fixes and additional support

etbe:
On Wednesday, 14 April 2021 6:42:29 AM AEST Guido Trentalancia wrote:
> +type spamd_update_gpg_t;
> 
> gpg is needed to verify the signatures on downloaded spamassassin rules !
> 
> without being able to run gpg (with a domain transition **BUT** without
> being able to access gpg_secret_t files and dirs), spamassassin is not able
> to verify the signature of new rules downloaded from the Internet and
> therefore the update fails, as it won't install anything without being able
> to verify the signature !
> 
> gpg signature verification is vital to update both spamassassin
> classification rules and smartctl drive database (see other update), even
> though they are not executable or binary files, it's always a good idea to
> verify that the downloaded files are authentic and genuine (malicious files
> can still cause DoS and other service disruptions) !

Why do you need a separate GPG domain for this?

Why not just allow spamd_update_t to execute gpg_exec_t?

-- 
My Main Blog         http://etbe.coker.com.au/
My Documents Blog    http://doc.coker.com.au/

gtrentalancia:
@etbe the reason you can't simply use gpg_exec() but rather need to have a separate domain is as follows:

- gpg_exec() would execute gpg in its own domain and such process would then have no access rights over specific application file and dir contexts such as etc_mail in the case of the spamassassin module or fsdaemon_var_lib_t in the case of the smartmon module: in other words, the forked gpg process would not be able to create files such as the public gpg keyring;

- another reason is that gpg running in its own domain would then try to search/read gpg_secret_t files and that is undesirable.

etbe:
On Wednesday, 14 April 2021 8:50:59 PM AEST Guido Trentalancia wrote:
> @etbe the reason you can't simply use gpg_exec() but rather need to have a
> separate domain is as follows:
> 
> - gpg_exec() would execute gpg in its own domain and such process would then
> have no access rights over specific application file and dir contexts such
> as etc_mail in the case of the spamassassin module or fsdaemon_var_lib_t in
> the case of the smartmon module: in other words, the forked gpg process
> would not be able to create files such as the public gpg keyring;

Sure it can, just under a type that is writable by the calling domain.

> - another reason is that gpg running in its own domain would then try to
> search/read gpg_secret_t files and that is undesirable.

No, it would have the gpg secret files in it's own domain.

What is the need to separate the GPG program from the other programs running 
as spamd_update_t?  Are you concerned about a compromise of GPG messing with 
spamd_update_t or a compromise of the spamd updater compromising GPG?  If the 
former I think that's a low risk as the data has to go through the spamd 
updater before it gets to gpg.  If the latter then a compromise of the spamd 
updater probably makes gpg irrelevant.

-- 
My Main Blog         http://etbe.coker.com.au/
My Documents Blog    http://doc.coker.com.au/

gtrentalancia:
@etbe

If gpg is executed from a smartmon daemon (fsdaemon_t) using a gpg_exec() call, then the forked gpg process running in the gpg_t context will not be able to read/write/manage the gpg keyring files in the fsdaemon_var_lib_t context and the update will fail !

This issue is very general in refpolicy and, as far as I know, it has already been discussed extensively since long time ago: please refer to the apache module in apache.te file, in particular please see the httpd_gpg_t domain definition and usage.

gtrentalancia:
@etbe As for the gpg_security_t directory and file context matter, since gpg is only used to verify digital signatures of downloaded content, it doesn't need any access to secret key files and any attempt to access (read/write/manage) such files should be forbidden !

policy/modules/system/systemd.if
gtrentalancia: If a new such kind of attribute is introduced, such as "systemd_systemd_user_session" that has been introduced in template systemd_role_template(), then you also need to provide an interface that amkes it possible to actually use such attribute easily in the policy...

gtrentalancia:
@pebenito The changes discussed in your review have now been merged in this pull request.

policy/modules/system/systemd.if
pebenito: Yes, I missed that it said type. The correction should be to make it an attribute requirement. I don't think having the interface significantly improves readability. The worse thing is that it makes it look like other domains could be systemd user sessions, and that's not true.

policy/modules/kernel/kernel.te
pebenito: I understand there are other init systems, I'm asking can you provide an example where this is happening? code, script, etc.

pebenito:
I agree with @etbe. The purpose of the gpg domains is to tightly control access to private keys. For verifying signatures of downloads, that isn't applicable. Yes, a compromised gpg process could abuse the calling domain's access in the execute_no_trans case, but that doesn't seem like a significant enough concern, considering the data given to gpg comes from that process.

I think we shouldn't add these new gpg domains.

policy/modules/system/systemd.if
gtrentalancia: So, do you sugget removing the interface ?

policy/modules/kernel/kernel.te
gtrentalancia: Usually scripts that run at bootup need to carry out operations on storage devices.

gtrentalancia: The new permissions introduced do not seem partilarly risky when executed from a kernel context: at the end, it seems quite normal to me that the kernel should at least be able able to [gs]etattr storage devices.

gtrentalancia:
If you don't introduce the new spamassassin and smartmon gpg domains then the two applications will not be able to perform the updates because they will not be able to verify the gpg signatures and create the needed keyrings in their own directories.

etbe:
On Friday, 16 April 2021 6:31:25 AM AEST Guido Trentalancia wrote:
> If you don't introduce the new spamassassin and smartmon gpg domains then
> the two applications will not be able to perform the updates because they
> will not be able to verify the gpg signatures and create the needed
> keyrings in their own directories.

Why would those domins have such problems when other domains have gpg_exec() 
and it works for them?

-- 
My Main Blog         http://etbe.coker.com.au/
My Documents Blog    http://doc.coker.com.au/

gtrentalancia:
All other domains or modules currently implemented in the policy that use gpg are implemented in a similar manner if they have to perform gpg signature verification.

Namely, the other modules currently implementing gpg_exec() are: apache, rkhunter, kdump and portage.

I have already mentioned the apache module as a long-time reference for this problem: https://github.com/SELinuxProject/refpolicy/pull/360#issuecomment-819581424 !!!

Portage is using the same kind of approach as apache and the one I am proposing for the two spamassassin and smartmon modules (i.e. gpg_exec() in a new specialized domain along with gpg_entry_type()) !

As for the other two remaining modules named above , the situation is as follows: neither of rkhunter or kdump are effectively using gpg for signature verification or even worse they might not be using gpg at all thus making the gpg_exec() reference in their respective module misplaced and unneeded (therefore they might appear to "work", when effectively they are not using gpg at all).

I hope this helps and closes the issue.

pebenito:
> If you don't introduce the new spamassassin and smartmon gpg domains then the two applications will not be able to perform the updates because they will not be able to verify the gpg signatures and create the needed keyrings in their own directories.

@gtrentalancia If you just do gpg_exec() and gpg can't write to the right place files/dirs, then why would it be bad to allow the access? You may have to provide a concrete example with real paths and files to illustrate this.

gtrentalancia:
The updates for both spamassassin and smartmon are performed by two scripts, named "sa-update" and "update-smart-drivedb" respectively.

The two scripts download non-binary non-executable data (text files) from the network to directories labeled as spamd_var_lib_t and fsdaemon_var_lib_t respectively.

Before actually installing the files, they verify that their digital gpg signature is valid, in order to make sure that the updates are genuine and have neither been tampered with nor are the result of a Man-In-The-Middle malicious file injection operation.

The operation performed is something like "gpg --verify". During such operation gpg uses the two spamd_var_lib_t and fsdaemon_var_lib_t mentioned above, therefore the gpg process should have read/write/create permissions for files in those directories. The specific paths varies but are generally /var/lib/spamassassin and /usr/share/smartmontools.

Without signature verification, although quite sophisticate to be of any practical convenience, a possible attack scenario could render the spamassassin ineffective by: DNS hijacking to fake update server -> injection of bogus spam detection rules -> spamassassin would not be able to detect spam anymore.

And similarly for smartmon: DNS hijacking to fake update server -> injection of malicious hard drive data -> could potentially cause Denial of Service to the data storage failure monitoring infrastructure.

etbe:
On Saturday, 17 April 2021 05:23:00 AEST Guido Trentalancia wrote:
> The updates for both spamassassin and smartmon are performed by two scripts,
> named "sa-update" and "update-smart-drivedb" respectively.
> 
> The two scripts download non-binary non-executable data (text files) from
> the network to directories labeled as spamd_var_lib_t and
> fsdaemon_var_lib_t respectively.
> 
> Before actually installing the files, they verify that their digital gpg
> signature is valid, in order to make sure that the updates are genuine and
> have neither been tampered with nor are the result of a Man-In-The-Middle
> malicious file injection operation.

Basic GPG operation.

> The operation performed is something like "gpg --verify". During such
> operation gpg uses the two spamd_var_lib_t and fsdaemon_var_lib_t mentioned
> above, therefore the gpg process should have read/write/create permissions
> for files in those directories. The specific paths varies but are generally
> /var/lib/spamassassin and /usr/share/smartmontools.

So GPG needs read/write access to the directories that the calling scripts 
have read/write access to.

> Without signature verification, although quite sophisticate to be of any
> practical convenience, a possible attack scenario could render the
> spamassassin ineffective by: DNS hijacking to fake update server ->
> injection of bogus spam detection rules -> spamassassin would not be able
> to detect spam anymore.

Basic GPG operation.

> And similarly for smartmon: DNS hijacking to fake update server -> injection
> of malicious hard drive data -> could potentially cause Denial of Service
> to the data storage failure monitoring infrastructure.

Basic GPG operation.

So you have said a dozen times that basic GPG operation is needed.  But why 
does that need a different domain than the calling scripts?

-- 
My Main Blog         http://etbe.coker.com.au/
My Documents Blog    http://doc.coker.com.au/

gtrentalancia:
I have now opened the following to issues so that it is easier to track down the problem:

- spamassassin: https://github.com/SELinuxProject/refpolicy/issues/362

- smartctl (smartmon module): https://github.com/SELinuxProject/refpolicy/issues/363

gtrentalancia:
> Basic GPG operation.

Not exactly - and this is critiical point: basic gpg operation *with the exclusion of secret gpg files access* (which is sensitive data)

> So you have said a dozen times that basic GPG operation is needed.  But why
> does that need a different domain than the calling scripts?

As already explained a few days ago here:

https://github.com/SELinuxProject/refpolicy/pull/360#issuecomment-819425344

pebenito:
I don't think we're making any progress on this discussion. Lets try a different way. Can you do the commands in permissive mode with gpg_exec() and provide a list of denials? Also note which denials you feel justify the separate gpg domain.

gtrentalancia:
The pull request and all associated patches are provided as they are and as courtesy.

So, you can either accept it as it is or carry out further testing on your own if required or just drop it.

I hope this clarifies the matter.

pebenito:
Very well. Since I am not yet convinced the new gpg domains are necessary, I must reject this PR. Please feel free to submit a new PR with those changes removed, if you like.

gtrentalancia:
The solution proposed is the best possible solution to sort out the problems reported in:

https://github.com/SELinuxProject/refpolicy/issues/363
https://github.com/SELinuxProject/refpolicy/issues/362

Closing the pull request won't solve the underlying issues but simply degrade the level of usability of the current SELinux policy and it might even decrease the level of security of systems using such policy.

Feel free to merge it at any other time, since it's provided as a courtesy I have no further time to spend on it.

PR #361, opened 2021-04-15
Resolve compile issue on RHEL7

dsugar100:
I was seeing the following error when compiling on RHEL7:

/usr/bin/checkmodule -m tmp/auditadm.tmp -o tmp/auditadm.mod
/usr/bin/checkmodule:  loading policy configuration from tmp/auditadm.tmp
policy/modules/roles/auditadm.te:42:ERROR 'duplicate declaration of type/attribute' at token 'auditadm_systemd_t' on line 31749:
		type auditadm_systemd_t;
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
make: *** [tmp/auditadm.mod] Error 1

This is seen with the changes in cc8374fd24129a2a20669bda2b57d8b029945047

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -1527,6 +1527,11 @@ type_transition systemd_user_session_type systemd_user_runtime_t:sock_file syste
 allow systemd_user_session_type systemd_user_tmpfs_t:file manage_file_perms;
 fs_tmpfs_filetrans(systemd_user_session_type, systemd_user_tmpfs_t, file)
 
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "generator.early")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "generator.late")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "transient")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "user")
+
 # Run generators in /usr/lib/systemd/user-environment-generators with no domain transition
 can_exec(systemd_user_session_type, systemd_generator_exec_t)
 
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -679,6 +679,13 @@ template(`userdom_common_user_template',`
 		')
 	')
 
+	#DJS - 4-May-2021 moved here due to issues compiling on RHEL7.  I was getting
+	# errors related to multiple declaration of auditadm_systemd_t.  Moving outside
+	# an optional block seemed very important.
+	ifdef(`init_systemd',`
+		systemd_role_template($1, $1_r, $1_t)
+	')
+
 	tunable_policy(`user_direct_mouse',`
 		dev_read_mouse($1_t)
 	')
@@ -859,10 +866,6 @@ template(`userdom_common_user_template',`
 
 	optional_policy(`
 		slrnpull_search_spool($1_t)
-	')
-
-	optional_policy(`
-		systemd_role_template($1, $1_r, $1_t)
 	')
 
 	optional_policy(`
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -118,8 +118,10 @@ template(`systemd_role_template',`
 	systemd_search_user_runtime_unit_dirs($1_systemd_t)
 	systemd_read_user_unit_files($1_systemd_t)
 
-	dbus_system_bus_client($1_systemd_t)
-	dbus_spec_session_bus_client($1, $1_systemd_t)
+	optional_policy(`
+		dbus_system_bus_client($1_systemd_t)
+		dbus_spec_session_bus_client($1, $1_systemd_t)
+	')
 
 	# userdomain rules
 	allow $3 $1_systemd_t:process signal;
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -81,11 +81,6 @@ template(`systemd_role_template',`
 	allow $1_systemd_t $3:file read_file_perms;
 	allow $1_systemd_t $3:lnk_file read_lnk_file_perms;
 
-	filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "generator.early")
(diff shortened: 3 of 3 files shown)


0xC0ncord:
@gtrentalancia also reported this. I'm wondering why the build tests are reporting failure here if this indeed fixes a build problem. I recall I had similar issues when originally writing [`cc8374f`](https://github.com/SELinuxProject/refpolicy/commit/cc8374fd24129a2a20669bda2b57d8b029945047). I don't have a RHEL system handy to actually test on.

gtrentalancia:
The problem you report has already been fixed here:

https://github.com/SELinuxProject/refpolicy/pull/360/commits/716093aba808fbd53c1c2647a7e98489eb500bd2

gtrentalancia:
It's just the the pull request https://github.com/SELinuxProject/refpolicy/pull/360 has not been merged yet...

dsugar100:
@gtrentalancia Thanks for pointing this out.  I will close this merge request and hope that yours gets merged soon.  
@0xC0ncord It turns out the build tests fail when building monolithic.  It worked in my tests on RHEL7 prior to pushing but I was testing modular.

dsugar100:
It looks like #360 has been closed, but the problem I am trying to solve still remains. 
This patch obviously needs some work, but I'm not exactly sure how to resolve the issue.

I don't understand why the build checks here are not complaining without this patch.
When building modular policy with this patch RHEL7 and Fedora 33 I have success.
When building monolithic policy I have errors (I think similar to what is reported by the checks).

I will keep working, but would appreciate some suggestions.

gtrentalancia:
As already said, as far as I know, the problem has been fixed in this pull request:

https://github.com/SELinuxProject/refpolicy/pull/360

and more specifically in the following commit that belongs to the above mentioned pull request:

https://github.com/SELinuxProject/refpolicy/commit/716093aba808fbd53c1c2647a7e98489eb500bd2

So, please try the above mentioned patch and then let me know.

dsugar100:
@gtrentalancia yes, I looked a that, but had similar problems with that change also.  The fact that only $1_systemd_t was moved into the gen_require block but not the attributes it is associated with didn't make sense to me either.  

I will keep working on it, but it is pretty deep inside templates and hard to follow exactly what the complaint is making it hard for me to follow.

0xC0ncord:
I fail to understand how 716093a fixes the issue, when that moves the declaration of `$1_systemd_t` in such a way that it never gets declared at all. I'm also looking into this and I'm not sure what's going on here. Obviously there is an issue with the tests in that this wasn't picked up, but I don't see how changing the policy build from modular to monolithic would suddenly call that interface to be called twice for the `auditadm` user, since it is only ever called in `userdom_common_user_template()`.

I will see if I can reproduce this and find out where that type is getting created twice.

pebenito:
It sounds like an ordering issue. The complier can only handle
1. declare type
2. require type
3. require type
4. ...

There cannot be any requires for a type before the type is declared. Moving the declaration in `systemd_role_template` to a require means it will never be declared.

dsugar100:
Ok, I think this is now ready for review.  Sorry it took so long for me to get back to it, other stuff has been going on.  These should probably get squashed into a single commit, but I left them separate to show the different errors that I was correcting.  I'm happy to squash if desired.

policy/modules/system/userdomain.if
```diff
@@ -679,6 +679,10 @@ template(`userdom_common_user_template',`
 		')
 	')
 
+	ifdef(`init_systemd',`
+		systemd_role_template($1, $1_r, $1_t)
```
pebenito, line 683: Did you try putting this in the `dbus_system_bus_client` block to avoid making dbus optional in `systemd.if`? I don't think this can work w/o dbus.

Also, please put a comment for why it is out of order
1. when we forget about this we don't inadvertently put it back
2. when RHEL7 finally goes EOL we can restore it to the original location

pebenito (changes requested):

policy/modules/system/userdomain.if
dsugar100: I hadn't tried this in dbus_system_bus_client block (until you suggested it).  But, I get the same error about duplicate definition when building.  It seems like putting systemd_role_template in an optional is causing some problem, but I really don't understand why.  

It seems to also be resolved by commenting out the use of 'systemd_user_daemon_domain' in 'dbus_role_template' which is what is requiring the 'auditadm_systemd_t'.  But doing that would change functionality.  

I will add a comment and update the pull request.  
If there is anything else you would like me to try, let me know.

pebenito: Please provide the error you're getting.

dsugar100: I'm sorry @pebenito , are you asking me for the error seen when moving 'systemd_role_template' into the 'dbus_system_bus_client' optional block?  It is the same:

/usr/bin/checkmodule -m tmp/auditadm.tmp -o tmp/auditadm.mod
/usr/bin/checkmodule:  loading policy configuration from tmp/auditadm.tmp
policy/modules/roles/auditadm.te:42:ERROR 'duplicate declaration of type/attribute' at token 'auditadm_systemd_t' on line 32039:
#line 42
		type auditadm_systemd_t;
/usr/bin/checkmodule:  error(s) encountered while parsing configuration
make: *** [tmp/auditadm.mod] Error 1


If you are asking for it to be clearer in the comment in systemd.if, I can do that also.

pebenito: Unless there is some distro other than RHEL7 that hits this, I'm afraid I have to reject this piece of the PR. RHEL7 is not officially supported by refpolicy, and this is over the line of what I feel comfortable carrying. If it was just the change in `auditadm.te`, it would be fine, but I draw the line at making dbus optional when it's not.

I still do want the other pieces.

policy/modules/system/systemd.te
```diff
@@ -1527,6 +1527,11 @@ type_transition systemd_user_session_type systemd_user_runtime_t:sock_file syste
 allow systemd_user_session_type systemd_user_tmpfs_t:file manage_file_perms;
 fs_tmpfs_filetrans(systemd_user_session_type, systemd_user_tmpfs_t, file)
 
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "generator.early")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "generator.late")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "transient")
+filetrans_pattern(systemd_user_session_type, systemd_user_runtime_t, systemd_user_runtime_unit_t, dir, "user")
```
pebenito, line 1533: Please move these up to line 1526 (`runtime_unit` comes after `runtime_notify`)

PR #366, opened 2021-04-24
${ROLE}_dbusd_t: allow domain transition from ${ROLE}_systemd_t

topimiettinen:
Allow dbus-daemon started by systemd running as user manager (with
e.g. type user_systemd_t) to transition to appropriate
type (e.g. user_dbusd_t). Also let dbus-daemon access user manager's
socket.

### DIFF
--- a/policy/modules/services/dbus.if
+++ b/policy/modules/services/dbus.if
@@ -64,6 +64,7 @@ template(`dbus_role_template',`
 		type system_dbusd_t, dbusd_exec_t;
 		type session_dbusd_tmp_t, session_dbusd_home_t;
 		type session_dbusd_runtime_t;
+		type $1_systemd_t;
 	')
 
 	##############################
@@ -97,6 +98,9 @@ template(`dbus_role_template',`
 	userdom_user_home_dir_filetrans($3, session_dbusd_home_t, dir, ".dbus")
 
 	domtrans_pattern($3, dbusd_exec_t, $1_dbusd_t)
+
+	domtrans_pattern($1_systemd_t, dbusd_exec_t, $1_dbusd_t)
+	allow $1_dbusd_t $1_systemd_t:unix_stream_socket { accept getattr getopt read write };
 
 	ps_process_pattern($3, $1_dbusd_t)
 	allow $3 $1_dbusd_t:process { ptrace signal_perms };


policy/modules/services/dbus.if
```diff
@@ -98,6 +99,9 @@ template(`dbus_role_template',`
 
 	domtrans_pattern($3, dbusd_exec_t, $1_dbusd_t)
 
+	domtrans_pattern($1_systemd_t, dbusd_exec_t, $1_dbusd_t)
+	allow $1_dbusd_t $1_systemd_t:unix_stream_socket { accept getattr getopt read write };
```
doverride, line 103: Are you sure about this? I believe that this is not right.

systemd --user should create the socket activated socket with setsockcreate if all is well, and dbus should not have to accept and read/write systemd sockets

doverride: Aside from that you are not allowed to reference the external $1_systemd_t type here directly. Not to mention that systemd should be optional to dbus

topimiettinen: The `access` is weird, but here's the AVC entry:
```
Apr 22 20:21:58 audit[1648]: AVC avc:  denied  { accept } for  pid=1648 comm="dbus-daemon" path="/run/user/1000/bus" scontext=user_u:user_r:user_dbusd_t:s0 tcontext=user_u:user_r:user_systemd_t:s0 tclass=unix_stream_socket permissive=1
```
I think `read` & `write` rules are needed since dbus-daemon can start services with the help of systemd when systemd activation with `SystemdService=` is used. Maybe `getopt` could be ignored.

topimiettinen: Actually, just a few lines down there's `systemd_user_daemon_domain($1, dbusd_exec_t, $1_dbusd_t)` which should do exactly what this patch is supposed to do, with the sole exception of `accept`. Why is that not then working?

topimiettinen: I see, for some reason I didn't have the correct version, which includes cc8374fd24 when testing. Sorry for the noise.

doverride: make sure that $1_systemd_t has access to `self:process setsockcreare`

because it should create, and listen on,  that $1_dbus_t socket with type $1_dbus_t instead of $1_systemd_t, and then $1_dbus_t should have to accept on $1_systemd_t sockets, it will however probably want to read/write $1_systemd_t sockets because its systemd forked (and its stdout is caught be systemd for journal using its unix stream socket i think)

topimiettinen: Hmm, if you meant this below, it didn't help:
```
diff --git a/policy/modules/system/systemd.if b/policy/modules/system/systemd.if
index 8462a1420..1c9e7bb43 100644
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -53,7 +53,7 @@ template(`systemd_role_template',`
        # This domain is per-role because of the below transitions.
        # See the systemd --user section of systemd.te for the
        # remainder of the rules.
-       allow $1_systemd_t self:process { getsched signal };
+       allow $1_systemd_t self:process { setsockcreate getsched signal };
        allow $1_systemd_t self:netlink_kobject_uevent_socket create_socket_perms;
        allow $1_systemd_t self:unix_stream_socket create_stream_socket_perms;
        allow $1_systemd_t $3:process { setsched rlimitinh signal_perms };
```

I still get this:
```
Apr 24 15:29:47 audit[1648]: AVC avc:  denied  { accept } for  pid=1648 comm="dbus-daemon" path="/run/user/1000/bus" scontext=user_u:user_r:user_dbusd_t:s0 tcontext=user_u:user_r:user_systemd_t:s0 tclass=unix_stream_socket permissive=1
Apr 24 15:29:47 audit[1648]: SYSCALL arch=c000003e syscall=288 success=yes exit=23 a0=3 a1=316519cf5880 a2=316519cf587c a3=80000 items=0 ppid=1601 pid=1648 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=(none) ses=2 comm="dbus-daemon" exe="/usr/bin/dbus-daemon" subj=user_u:user_r:user_dbusd_t:s0 key=(null)
```

doverride: what version of systemd is that? Maybe a `systemctl daemon-reexec` helps?

Other than that I dont know what is going on but systemd should create the dbus socket with the context of the dbus process and "listen" on that

```
allow $1_systemd_t self:process setsockcreate;
allow $1_systemd_t $1_dbusd_t:unix_stream_socket { create_socket_perms listen } ;
```

and the dbus should "just work" because it already can accept on $1_dbusd_t unix domain sockets

doverride: is systemd compiled with `+SELinux`?

topimiettinen: Found the root cause of problem: mode of `/etc/selinux` was 0700 (because of my overeager administration), so without access to SELinux policy, `systemd --user` didn't use correct type (default `user_runtime_t` instead of expected `session_dbusd_runtime_t`) when creating the socket for dbus. After fixing the permissions, no patch is needed and there's no error anymore.

topimiettinen: Thanks for the help!

PR #369, opened 2021-04-27
Irc: SOCK proxy and screen fixes

jpds:
Fixes to allow IRC clients to connect to optionally connect to SOCKS proxies and also the screen status attachment status.


pebenito:
I'm ok with the rules, but can you explain why they should be conditional?

policy/modules/apps/irc.te
```diff
@@ -112,6 +128,10 @@ userdom_user_content_access_template(irc, irc_t)
 
 xdg_manage_downloads(irc_t)
 
+tunable_policy(`irc_allow_screen_access',`
```
pebenito, line 131: Regardless if we keep the tunables, this should be in an optional block, otherwise it will always need screen.

pebenito (changes requested):

jpds:
@pebenito I was unsure if it should be implicitly allowed as it's not core functionality - however I've now realized that it's simpler to just do so.

pebenito:
It's not a core functionality, but I don't see an issue if it is available all the time.

PR #370, opened 2021-05-05
domain: Revise ruleset so domain_base_type() has near no rules.

### LINKED ISSUE #328: Provide barebones versions of existing domains
Currently, `domain` adds more permissions than are actually needed.

It would be nice to have a type (`barebones_domain`?) that includes exactly the permissions needed to run a program that does nothing, and no more.  Similarly, there would be a `barebones_daemon` with the minimum permissions for a daemon that just immediately exits, and so on.

This would be useful for those who want a strict default-deny policy, since it ensures that access is not accidentally granted.

pebenito:
Based on the current design there are a couple rules that still
get added to domain due to domain interfaces being called.
```
allow domain livecd_t:process sigchld;
allow domain netlabel_peer_t:peer recv;
allow domain netlabel_peer_t:tcp_socket recvfrom;
allow domain sysadm_t:process sigchld; [ allow_ptrace ]:True
allow domain unconfined_domain_type:association recvfrom;
allow domain unconfined_domain_type:peer recv;
allow domain unconfined_domain_type:tcp_socket recvfrom;
```
This also moves kernel_t and chromium_renderer_t to full
domains to ensure no rules are lost.

Fixes #328


pebenito:
> Is `chromium_renderer_t` still actually used? IIRC Chromium ripped out the SELinux sandbox quite a while ago. That said, I would very much like to add it back in, as it works even when a user cannot use SUID binaries or user namespaces.

@etbe can you speak to the above question?

DemiMarie:
This looks good, though I have not had a chance to try it locally.

PR #371, opened 2021-05-06
init: Add support for systemd StandardInputText.

pebenito:
This normally uses memfd which uses posix shm under the hood. Additionally, a direct shm use is a fallback if memfd is not available.


pebenito:
cc @bluca

bluca (approved):
LGTM

PR #374, opened 2021-05-12
roles: move dbus_role_template out of the condition block for sysadm/…

yizhao1:
…staff/unprivuser

After commit cc8374fd24129a2a20669bda2b57d8b029945047 (various: systemd
user fixes and additional support), the dbus_role_template is required
for all roles. But for sysadm/staff/unprivuser, this piece of code is
not included if the DISTRO is set to redhat. Move it out of the
'ifndef distro_redhat' condition block, otherwise the label of the
instance is not correct.

Before the patch if set DISTRO=redhat:
root@qemux86-64:~# ps xZ | grep "systemd --user"
root:sysadm_r:sysadm_t  240 ? Ss 0:00 /lib/systemd/systemd --user

After the patch:
root@qemux86-64:~# ps xZ | grep "systemd --user"
root:sysadm_r:sysadm_systemd_t  218 ? Ss 0:00 /lib/systemd/systemd --user

### DIFF
--- a/policy/modules/roles/auditadm.te
+++ b/policy/modules/roles/auditadm.te
@@ -40,10 +40,6 @@ optional_policy(`
 ')
 
 optional_policy(`
-	dbus_role_template(auditadm, auditadm_r, auditadm_t)
-')
-
-optional_policy(`
 	screen_role_template(auditadm, auditadm_r, auditadm_t)
 ')
 
--- a/policy/modules/roles/secadm.te
+++ b/policy/modules/roles/secadm.te
@@ -49,10 +49,6 @@ optional_policy(`
 ')
 
 optional_policy(`
-	dbus_role_template(secadm, secadm_r, secadm_t)
-')
-
-optional_policy(`
 	dmesg_exec(secadm_t)
 ')
 
--- a/policy/modules/roles/staff.te
+++ b/policy/modules/roles/staff.te
@@ -98,19 +98,15 @@ ifndef(`distro_redhat',`
 	')
 
 	optional_policy(`
-		dbus_role_template(staff, staff_r, staff_t)
+		gnome_role_template(staff, staff_r, staff_t)
+	')
 
-		optional_policy(`
-			gnome_role_template(staff, staff_r, staff_t)
-		')
+	optional_policy(`
+		telepathy_role_template(staff, staff_r, staff_t)
+	')
 
-		optional_policy(`
-			telepathy_role_template(staff, staff_r, staff_t)
-		')
-
-		optional_policy(`
-			wm_role_template(staff, staff_r, staff_t)
-		')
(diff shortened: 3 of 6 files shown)


pebenito (changes requested):
It seems like the better choice would be to move the systemd block in userdom_common_user_template() into the optional with all of the other dbus calls.

policy/modules/roles/auditadm.te
```diff
@@ -39,10 +39,6 @@ optional_policy(`
 	dmesg_exec(auditadm_t)
 ')
 
-optional_policy(`
-	dbus_role_template(auditadm, auditadm_r, auditadm_t)
```
pebenito, line 43: This and the below ones shouldn't be removed. The only templates that call `userdom_common_user_template()` is `userdom_unpriv_user_template()` and `userdom_admin_user_template()'.

pebenito (changes requested):
Specifically, I think what needs to be done in `userdom_admin_user_template()` is:

1. replace the `dbus_system_bus_client()` with `dbus_role_template()`
2. move the `systemd_role_template()` optional block nested into the dbus optional block, similar to `accountsd_dbus_chat()`, `bluetooth_dbus_chat()`, etc. optional blocks. This reinforces that the `systemd_role_template()` needs dbus.

policy/modules/roles/auditadm.te
pebenito: My mistake, auditadm and secadm call the unpriv user template, but the others don't. Please check with sediff to make sure no rules are lost.

yizhao1:
> Specifically, I think what needs to be done in `userdom_admin_user_template()` is:
> 
>     1. replace the `dbus_system_bus_client()` with `dbus_role_template()`
> 
I'm not sure if dbus_role_template() can replace dbus_system_bus_client().

In userdom_restricted_xwindows_user_template(), both of these 2 lines exists:

1097     optional_policy(`
1098         dbus_role_template($1, $1_r, $1_t)
1099         dbus_system_bus_client($1_t)


//Yi
>     2. move the `systemd_role_template()` optional block nested into the dbus optional block, similar to `accountsd_dbus_chat()`, `bluetooth_dbus_chat()`, etc. optional blocks. This reinforces that the `systemd_role_template()` needs dbus.

pebenito:
You are correct about `dbus_system_bus_client()` not being in `dbus_role_template()`, but I think it should be like this:
```
optional_policy(`
	dbus_role_template($1, $1_r, $1_t)
	dbus_system_bus_client($1_t)

	optional_policy(`
		accountsd_dbus_chat($1_t)
	')
	...
	optional_policy(`
		systemd_role_template($1, $1_r, $1_t)
	')
')
```

pebenito (approved):

PR #375, opened 2021-05-13
Use correct interface or template declaration

cgzones:
Following the guideline of interfaces not allowed to declare anything and not use prefix parameters, declare interfaces doing so as templates.

Also declare templates not using those features and not calling templates themselves as interfaces.

These changes originate from the discussion in https://github.com/TresysTechnology/selint/issues/205 and are found by new proposed SELint checks at https://github.com/TresysTechnology/selint/pull/206.

### DIFF
--- a/policy/modules/admin/kismet.if
+++ b/policy/modules/admin/kismet.if
@@ -15,7 +15,7 @@
 ##	</summary>
 ## </param>
 #
-template(`kismet_role',`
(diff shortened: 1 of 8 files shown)


pebenito (approved):

pebenito:
You have RFC, so I'm waiting on merging, if you have any other changes.

PR #377, opened 2021-05-28
logrotate.te: Added boolean for allowing logrotate to rotate the audit log

jpds:


policy/modules/admin/logrotate.te
```diff
@@ -5,6 +5,14 @@ policy_module(logrotate, 1.25.1)
 # Declarations
 #
 
+## <desc>
+##      <p>
+##      Determine whether logrotate can manage
+##      audit log files
+##      </p>
+## </desc>
+gen_tunable(allow_logrotate_manage_audit_log, false)
```
pebenito, line 14: Please remove the `allow_` from the name.

pebenito (changes requested):

policy/modules/admin/logrotate.te
```diff
@@ -121,6 +129,10 @@ logging_send_syslog_msg(logrotate_t)
 logging_send_audit_msgs(logrotate_t)
 logging_exec_all_logs(logrotate_t)
 
+tunable_policy(`logrotate_manage_audit_log',`
+	logging_manage_audit_log(logrotate_t)
+')
```
pebenito, line 134: Sorry, I missed this in my first read, this block should be after the distro_debian block below (before the optionals)

doverride:
Chris PeBenito ***@***.***> writes:

> @pebenito commented on this pull request.
>
> -------------------------------------------------------------------------------------------------------------------
>
> In policy/modules/admin/logrotate.te:
>
>> +tunable_policy(`logrotate_manage_audit_log',`
> +	logging_manage_audit_log(logrotate_t)
> +')
>
> Sorry, I missed this in my first read, this block should be after the distro_debian block below (before the
> optionals)

I am just going to drop this here for whatever it is worth.

auditd has its own internal way of rotating logs (SIGUSR1). Audit is for
a special kind of audience that prioritizes auditing over anything
else (and its pretty much self-contained). I suspect that this audience probably does not like the idea of
some external process "managing" audit logs. I realize that making this
tunable eleviates some of this but it still feels wrong to me to
"manage" to audit logs outside of auditd.

When I brought this up on IRC I got the impression that using auditd in
this scenario is not appropriate anyway.

Regardless, I just wanted to express my thoughts here. I am fine with
whatever is decided as long as these aspects are weighed

>
> —
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

pebenito:
I agree that this is not normal procedure, for the reasons you note. But I'm ok with that as long as its tunable and default false.

pebenito:
@jpds can you elaborate why you use logrotate instead of auditd's own rotation?

jpds:
@pebenito Sure, I want all my logs to be handled in a consistent way, it makes it far easier to know which day a log is from and thus correlate events across logs/services at the host level:

```
$ sudo ls -lZ /var/log/audit/
-rw-------. 1 root root system_u:object_r:auditd_log_t 163241 Jun  7 15:21 audit.log
-rw-------. 1 root root system_u:object_r:auditd_log_t  75255 May 28 13:40 audit.log-20210528.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t 111165 May 29 03:08 audit.log-20210529.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  27349 May 30 03:44 audit.log-20210530.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t 149510 May 31 03:45 audit.log-20210531.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  22025 Jun  1 03:40 audit.log-20210601.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  23063 Jun  2 03:33 audit.log-20210602.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  37828 Jun  3 03:39 audit.log-20210603.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  13408 Jun  4 03:11 audit.log-20210604.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  14530 Jun  5 03:39 audit.log-20210605.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  13238 Jun  6 03:14 audit.log-20210606.gz
-rw-------. 1 root root system_u:object_r:auditd_log_t  65733 Jun  7 03:24 audit.log-20210607.gz
# ls /var/log/syslog* -lZ
-rw-r-----. 1 root root system_u:object_r:var_log_t 3465080 Jun  7 15:28 /var/log/syslog
-rw-r-----. 1 root root system_u:object_r:var_log_t 4093777 May 16 03:38 /var/log/syslog-20210516.gz
-rw-r-----. 1 root root system_u:object_r:var_log_t 1240442 May 23 03:48 /var/log/syslog-20210523.gz
-rw-r-----. 1 root root system_u:object_r:var_log_t 1883631 May 30 03:50 /var/log/syslog-20210530.gz
-rw-r-----. 1 root root system_u:object_r:var_log_t 1612803 Jun  6 03:35 /var/log/syslog-20210606.gz
```

As opposed to having logs rotated at arbitrary times, which means that events for a single day/week could potentially be scattered across multiple files. Also auditd doesn't do compression.

Other examples of people doing this:

- https://access.redhat.com/solutions/661603
- https://bgstack15.wordpress.com/2018/02/13/logrotate-audit-log-selinux-cron-and-ansible/

doverride:
My argument was that in that cause one would probably be better off routing the messages to syslog for consistency. Seemed overkill to me to use audit for this. Unfortunately me argument does not hold up since utils like auditctl etc are bundled with the daemon. So the daemon is likely installed anyway, I guess might as well run it.

pebenito:
I get that, but I think that's more compelling argument for systems that have systemd-journald.

PR #378, opened 2021-05-28
virt: handle urandom unlink

jpds:
libvirt adds the following XML to VM definitions by default:

```
    <rng model='virtio'>
      <backend model='random'>/dev/urandom</backend>
      <address type='pci' domain='0x0000' bus='0x06' slot='0x00' function='0x0'/>
    </rng>
```

Which causes the VM to fail to start with:

```
error: Failed to start domain 'test1'
error: Unable to create device /dev/urandom: File exists
```

The AVC logged is:

```
type=AVC msg=audit(1622221825.029:3649): avc:  denied  { unlink } for  pid=6879 comm="rpc-worker" name="urandom" dev="tmpfs" ino=5 scontext=system_u:system_r:virtd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:urandom_device_t:s0 tclass=chr_file permissive=0
```

### DIFF
--- a/policy/modules/kernel/devices.if
+++ b/policy/modules/kernel/devices.if
@@ -3834,6 +3834,24 @@ interface(`dev_create_rand_dev',`
 
 ########################################
 ## <summary>
+##  Delete the random device (/dev/random).
+## </summary>
+## <param name="domain">
+##  <summary>
+##  Domain allowed access.
+##  </summary>
+## </param>
+#
+interface(`dev_delete_rand_dev',`
+	gen_require(`
+		type device_t, random_device_t;
+	')
+
+	delete_chr_files_pattern($1, device_t, random_device_t)
+')
+
+########################################
+## <summary>
 ##	Read the realtime clock (/dev/rtc).
 ## </summary>
 ## <param name="domain">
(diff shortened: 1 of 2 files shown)


pebenito:
If it's trying to create /dev/urandom, won't it need create perm too?

jpds:
@pebenito I looked through the code again, and it seems that they only have functions for `virChrdevFree` and `virChrdevOpen`:

- https://gitlab.com/libvirt/libvirt/-/blob/master/src/conf/virchrdev.c

`virChrdevLockFileCreate` is handled by the existing `dev_read_[u]rand_dev()` interfaces. I also think the first half of the error message is a red herring in that it is a generic error handler for the "creating" virtual devices for the virtual machine.

Whilst scanning the code, I also found that the default RNG is set to `/dev/random` if no backend file path is specified: https://gitlab.com/libvirt/libvirt/-/blob/master/src/conf/domain_conf.c#L5299 so I've added policy for that as well and tested the behavior on a box.

pebenito:
I don't think I follow your logic. From what I see it looks like it's actually operating on files that have `LCK` in it, due to `virChrdevLockFilePath()`

perfinion:
@jpds shouldnt this just have something like `dev_manage_rand_dev` instead of read+delete?

Issue #365, opened 2021-04-21
Proposal: add an attribute for application execute and transition permissions


0xC0ncord:
With the introduction of systemd user support, access needs to be added to `$1_systemd_t` for various applications if we want these to be run and transitioned properly. Other applications normally run by users such as window managers may also require such access. Instead of adding calls to `myapp_run()` for each of these applications, I think an attribute for this kind of access may be more suitable.

Such an attribute, `staff_app_runner_domain` for example, would have all the necessary access granted by interface calls like `chromium_run()`, and all that would be needed to ensure some domain has the same access to run applications would be to associate the `staff_app_runner_domain` to it, such as `staff_systemd_t`. That way, any application that can normally be run by `staff_t` can also be run by `staff_systemd_t`. Of course, explicitly allowing access to `staff_t` or `staff_systemd_t` can be used where appropriate.

I feel that this also has the advantage of making local policy development significantly easier to do, as one would not need to call the appropriate interfaces for every application that `staff_t` can normally run to whatever local policy module is being written. On the contrary, as pointed out in earlier discussion, this may overcomplicate refpolicy somewhat.

pebenito:
I agree that something is needed to solve this issue, and I think the Fedora policy has this type of attribute for a long time, but I've resisted it for a long time. I think this is finally a reason to do it.

Note: this would also apply to the per-user cronjob domains too.

0xC0ncord:
Looking at Fedora's policy, they create a `$1_usertype` attribute in `userdom_base_user_template()` that gets the same access that `$1_t` in refpolicy would. However, I only see that attribute being associated with `$1_t` and `$1_wine_t` or similar. I think that for what I would like to accomplish, that's a lot more permissive than I would like, and I think that an app's access to things like the user's home content should continue to be conditional. I think for starters, we can give access for executing other applications as well as allow `$1_t` and `$1_systemd_t` to have `signal_perms` on these application domains and work from there.

0xC0ncord:
I am anticipating getting to this soon, but I already worry that this might end up in needing to change the signatures of the interfaces for things like `chromium_role()` to include the prefix, which would then turn them into templates instead, or other such weirdness.

0xC0ncord:
Was kinda hoping that linking this issue in the [PR](#381) would not cause the bot to get upset, so here's an explicit activity bump.

0xC0ncord:
Not stale, waiting review on #412

PR #381, opened 2021-06-24
Add an attribute for user applications to run other applications

0xC0ncord:
This addresses the issue raised in #365. With the advent of `systemd --user` support, we should support users running various applications through user units and other means. The problem is that any other such applications normally started by a user that transition to their own domain (i.e. `user_systemd_t`) do not have the proper permissions to run other applications.

Many applications' policies have their own way of granting run access to users. Some policies grant access directly with `myapp_domtrans`, `myapp_run`, or by granting an attribute. I think the best way to do this would be to streamline all such interfaces and templates to take exactly 3 parameters: the role prefix, the role, and the type of the user domain. This way we can have a single "proper" way to declare interfaces and templates like this, and we can ensure that calls to things like `screen_role_template` have everything they need.

~~In this PR I have only modified `apache` and `screen` to use this approach, and they are now using the proposed attribute for user application run access. I am opening this PR to hopefully get some early feedback on this approach, as the final result of this effort may be quite large.~~

This is a two-part PR. This PR addresses adding the required infrastructure for this feature, while the upcoming second PR will hook up this infrastructure into the application policies to support it. 

This PR only adds the aforementioned attribute, the interfaces required to support it, and assigns it to the `$1_systemd_t` and `$1_wm_t` domains.

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -42,7 +42,7 @@ template(`systemd_role_template',`
 	type $1_systemd_t, systemd_user_session_type, systemd_log_parse_env_type;
 	init_pgm_spec_user_daemon_domain($1_systemd_t)
 	domain_user_exemption_target($1_systemd_t)
-	userdom_spec_user_application_exec_domain($1_systemd_t, $1)
+	userdom_application_exec_domain($1_systemd_t, $1)
 	ubac_constrained($1_systemd_t)
 	role $2 types $1_systemd_t;
 
@@ -198,8 +198,7 @@ template(`systemd_user_daemon_domain',`
 
 	domtrans_pattern($1_systemd_t, $2, $3)
 
-	allow $1_systemd_t $3:process signal_perms;
-	allow $3 $1_systemd_t:unix_stream_socket rw_socket_perms;
+	systemd_user_app_status($1, $3)
 ')
 
 ######################################
@@ -245,6 +244,46 @@ interface(`systemd_user_unix_stream_activated_socket',`
 
 	typeattribute $1 systemd_user_unix_stream_activated_socket_type;
 	systemd_user_activated_sock_file($2)
+')
+
+######################################
+## <summary>
+##   Allow the target domain to be monitored and have its output
+##   captured by the specified systemd user instance domain.
+## </summary>
+## <param name="prefix">
+##	<summary>
+##	Prefix for the user domain.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Domain to allow the systemd user instance to monitor.
+##	</summary>
+## </param>
+#
+template(`systemd_user_app_status',`
+	gen_require(`
+		type $1_systemd_t;
+	')
+
+	# This interface does not imply permissions for the systemd user
+	# domain to run the application. Instead, these rules are necessary
+	# for a systemd --user instance to monitor an application as well as
+	# any child processes of that application, children of those, and so
+	# on.
+
+	# systemd --user instances need to be able to read the state of the
+	# app it runs and capture its output to journald
+	ps_process_pattern($1_systemd_t, $2)
+	allow $1_systemd_t $2:process signal_perms;
+	allow $2 $1_systemd_t:fd use;
+	allow $2 $1_systemd_t:unix_stream_socket rw_socket_perms;
+
+	# apps run by systemd --user instances need to be able to read the
+	# state of the systemd --user instance
+	ps_process_pattern($2, $1_systemd_t)
+	allow $2 $1_systemd_t:process sigchld;
 ')
 
 ######################################
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -42,6 +42,7 @@ template(`systemd_role_template',`
 	type $1_systemd_t, systemd_user_session_type, systemd_log_parse_env_type;
 	init_pgm_spec_user_daemon_domain($1_systemd_t)
 	domain_user_exemption_target($1_systemd_t)
+	userdom_spec_user_application_exec_domain($1_systemd_t, $1)
 	ubac_constrained($1_systemd_t)
 	role $2 types $1_systemd_t;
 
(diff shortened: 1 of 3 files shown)


policy/modules/apps/screen.if
```diff
@@ -48,31 +49,31 @@ template(`screen_role_template',`
 
 	dontaudit $1_screen_t self:capability sys_tty_config;
 
-	domtrans_pattern($3, screen_exec_t, $1_screen_t)
+	domtrans_pattern($1_user_exec_domain, screen_exec_t, $1_screen_t)
```
pebenito, line 52: This won't work for *_role() interfaces that don't use a prefix. The $1_user_exec_domain needs to be passed in as $1.

0xC0ncord: >The $1_user_exec_domain needs to be passed in as $1.

The problem with this is that some `*_role()` interfaces associate an attribute with `$1`, and since `$1_user_exec_domain` is an attribute this will never work, which is my rationale for reworking all `*_role()` interfaces to take `$3` as the user prefix, or maybe just the user domain (i.e. `$3_t`).

pebenito: hmm, maybe we should expand it to be *_role([prefix], domain, exec_domain_attr, role) to cover both cases?

0xC0ncord: I think that would be a good solution. As long as we can ensure that `*_role()` interfaces can have the prefix, the domain, and the attribute, then there won't be any issues with creating the required rules. Updating this PR soon so we can see what that will look like.

pebenito:
I'm not sure where you are with this due to the weird github timestamping, but at a minimum, there are CI errors.

0xC0ncord:
> I'm not sure where you are with this due to the weird github timestamping, but at a minimum, there are CI errors.

I have a lot of local work that I haven't pushed yet. It will likely be a little bit more time before I think it is ready for review, so I am OK with closing this until then if you'd like.

pebenito:
You can set the PR to draft and leave it open.

0xC0ncord:
This is at the point now where every `*_role()` call I could find by the user domains has been reworked, and just needs some review and tuning to determine exactly what permissions and how those permissions should be granted to `$1_t` and `$1_user_exec_domain`. Initially I wanted the exec domain attribute to only be used for transition access, but there are a few edge cases where it might make sense to grant additional rules to be able to manage the applications' data.

As an aside, `$1_user_exec_domain` might need to be renamed, because `user_user_exec_domain` for the unprivuser is confusing.

0xC0ncord:
After some discussion, I have decided to split this PR into two parts. This PR will address the infrastructure needed to support this feature, and the upcoming second will support actually hooking up this infrastructure into the applications. I think this part is finally ready for review.

policy/modules/system/systemd.if
```diff
@@ -246,6 +246,46 @@ interface(`systemd_user_unix_stream_activated_socket',`
...
+##	Prefix for the user domain.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Domain to allow the systemd user instance to monitor.
+##	</summary>
+## </param>
+#
+interface(`systemd_user_app_status',`
```
pebenito, line 265: Why does this need to be split out of systemd_user_app_status?

policy/modules/system/systemd.if
```diff
@@ -246,6 +246,46 @@ interface(`systemd_user_unix_stream_activated_socket',`
...
+	gen_require(`
+		type $1_systemd_t;
+	')
+
+    # This interface does not imply permissions for the systemd user
+    # domain to run the application. Instead, these rules are necessary
```
pebenito, line 271: nit: tabs vs spaces

policy/modules/system/systemd.if
pebenito: This is also a template since it uses prefixes.

policy/modules/system/userdomain.if
```diff
@@ -242,6 +243,34 @@ template(`userdom_user_content_access_template',`
...
+## </param>
+## <param name="userdomain_prefix">
+##	<summary>
+##	The prefix of the user domain (e.g., user
+##	is the prefix for user_t).
+##	</summary>
+## </param>
+## <rolebase/>
+#
+interface(`userdom_spec_user_application_exec_domain',`
```
pebenito, line 266: I'm not fond of the name of this, but don't immediately have a better name. Also, this is a template.

policy/modules/system/systemd.if
0xC0ncord: I decided to split it out since this interface grants a subset of the same access that `systemd_user_daemon_domain()` gives. `systemd_user_daemon_domain()` allows the systemd --user instance to execute the app and monitor it, while `systemd_user_app_status` only allows the systemd --user instance to monitor it.

I don't really see a problem with allowing the exec access though, since the systemd --user instance is one of these domains that we will allow to run other applications like a userdomain, but figured it would allow for more fine-grained access control if needed.

pebenito: Where else would this be called? It's only called one place in this patch. That's the confusion.

0xC0ncord: > Where else would this be called? It's only called one place in this patch. That's the confusion.

This will need to be called by interfaces and templates that grant role or run access to an application, i.e. `screen_role_template()`. I can submit a draft PR for the hooking up of this infrastructure sooner if it would help visualize how this is to be used.

pebenito: I follow now.

policy/modules/system/userdomain.if
pebenito: `spec` is supposed to be for cases where setexec() is used to initiate a domain transition.

pebenito: how about `userdom_application_exec_domain`

pebenito (approved):

0xC0ncord:
Sorry, I forgot to change `systemd_user_app_status()` to a template in my last push. Corrected just now.

PR #384, opened 2021-07-02
cleanup: Missing requires

maage:
Several interfaces were missing requires.

### DIFF
--- a/policy/modules/kernel/files.if
+++ b/policy/modules/kernel/files.if
@@ -3226,6 +3226,7 @@ interface(`files_exec_etc_files',`
 interface(`files_get_etc_unit_status',`
 	gen_require(`
 		type etc_t;
+		class service status;
 	')
 
 	allow $1 etc_t:service status;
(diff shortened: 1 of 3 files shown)


policy/modules/system/systemd.if
```diff
@@ -1417,6 +1417,7 @@ interface(`systemd_status_networkd',`
 interface(`systemd_relabelfrom_networkd_tun_sockets',`
 	gen_require(`
 		type systemd_networkd_t;
+		class tun_socket relabelfrom;
```
pebenito, line 1420: This is not necessary.

pebenito (changes requested):

policy/modules/system/systemd.if
pebenito: To further clarify, `tun_socket` is a kernel object class, so it is always requried.

maage: Yes, I made some mistake here. Now when I checked these, I was not able to trigger the issue.

maage:
Dropper unnecessary patch as required.

pebenito (approved):

PR #386, opened 2021-07-02
cleanup: policy avahi: avahi_filetrans_pid: doc: add missing params

maage:
Even if interface is deprecated, still use all documented parameters.

Signed-off-by: Markus Linnala <Markus.Linnala@cybercom.com>

### DIFF
--- a/policy/modules/services/avahi.if
+++ b/policy/modules/services/avahi.if
@@ -255,7 +255,7 @@ interface(`avahi_dontaudit_search_pid',`
 #
 interface(`avahi_filetrans_pid',`
 	refpolicywarn(`$0($*) has been deprecated, please use avahi_filetrans_runtime() instead.')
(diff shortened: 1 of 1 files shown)


policy/modules/services/avahi.if
```diff
@@ -255,7 +255,7 @@ interface(`avahi_dontaudit_search_pid',`
 #
 interface(`avahi_filetrans_pid',`
 	refpolicywarn(`$0($*) has been deprecated, please use avahi_filetrans_runtime() instead.')
-	avahi_filetrans_runtime($1)
+	avahi_filetrans_runtime($1, $2, $3)
```
cgzones, line 258: maybe `avahi_filetrans_runtime($*)`?

maage: Changed and force pushed

pebenito (approved):

PR #389, opened 2021-07-02
Makefile fixes and build time speedups

maage:
Use actual directory targets to create them and then use order-only-presequites to ensure directory is created.
Do not use needlessly tmp files if can use pipe.
Couple of fixes.
Use m4 frozen state to speed up module processing.

As far as I can see, result is exactly the same except removed tmp file and added frozen state files in tmp. Tested couple of random build.conf configurations and this seems to be valid.

Frozen state can cause regressions if m4 defines are changed between rules. But that is not the case at the moment and hence exactly same results. And this would most probably generate bogus policy or fail anyway.

You can only use frozen state in rule if there is no additional` M4PARAM += -D` or `$(M4) -D`.

Speedup in my machine using: make -j$(nproc)
28,45s real (77,28s:12,46s)=315% 0K (16K)
->
8,97s real (18,51s:8,97s)=306% 0K (15K)

Here $(nproc) is 4. With speedup usually you can get away with: `make -j` as jobs are completed about as fast as they are started.


maage:
It seems, I need to iterate some build.conf settings more.

maage:
Okay, now I tested some configurations more and results seems to be exactly the same.
Previous version used frozen state when generating files with additional variables. Now as I do not use frozen state any more there and added M4PARAM to all_interfaces, then config should be safe.

Rules.modular
```diff
@@ -138,26 +146,25 @@ $(tmpdir)/global_bools.conf: M4PARAM += -D self_contained_policy
...
 	$(verbose) $(M4) $(M4PARAM) $^ > $@
 
-$(tmpdir)/all_interfaces.conf: $(m4support) $(all_interfaces) $(m4iferror)
-	@test -d $(tmpdir) || mkdir -p $(tmpdir)
+$(tmpdir)/all_interfaces.conf: | $(tmpdir)
+$(tmpdir)/all_interfaces.conf: $(tmpdir)/m4support.m4f $(all_interfaces) $(m4iferror)
 	@cat $(m4divert) > $@
-	$(verbose) $(M4) $^ > $(tmpdir)/$(@F).tmp
-	$(verbose) $(SED) -e s/dollarsstar/\$$\*/g $(tmpdir)/$(@F).tmp >> $@
+	$(verbose) $(M4) $(M4PARAM) -R $^ | $(SED) -e s/dollarsstar/\$$\*/g >> $@
```
fishilico, line 152: Hello, quick comment for the change "use pipe instead of tmp file": if I understand correctly how `make` and `sh` usually work, when a pipe is used to run `cmd1 | cmd2` and `cmd1` fails but `cmd2` succeeds, the whole command is seen as "success" (and `set -o pipefail` in bash prevents such an issue).

Here if the `m4` invocation fails, the build will no longer fail. This seems to be a regression. Is my interpretation correct? I might have missed where the "pipefail" option configured in this Makefile.

maage: You are right
```
$ cat Makefoo 
a:
	m4 <<< "m4exit(\`1')" | sed -e s/1/2/
b:
	m4 <<< "m4exit(\`1')" > tmp1
	sed -e s/1/2/ tmp1
$ make -f Makefoo 
m4 <<< "m4exit(\`1')" | sed -e s/1/2/
$ echo $?
0
$ make -f Makefoo b
m4 <<< "m4exit(\`1')" > tmp1
make: *** [Makefoo:5: b] Error 1
$ echo $?
2
```

cgzones: I'd prefer to not use any bashisms and stay POSIX compliant.

pebenito: I'd also prefer to have intermediate files since there is so much processing. If something goes wrong it will be harder to find. Build performance isn't a concern for refpolicy; correctness is.

pebenito (changes requested):

Rules.modular
maage: Ok, testfiles without <<< bashism
```
$ cat testmake
# .SHELLFLAGS := -c -o pipefail

a:
	echo "m4exit(\`1')" | m4 - | sed -e s/1/2/
b:
	echo "m4exit(\`1')" | m4 - > tmp1
	sed -e s/1/2/ tmp1
$ cat testmake-pipefail 
.SHELLFLAGS := -c -o pipefail

a:
	echo "m4exit(\`1')" | m4 - | sed -e s/1/2/
b:
	echo "m4exit(\`1')" | m4 - > tmp1
	sed -e s/1/2/ tmp1
```

Now without -o pipefail, using pipe does not fail:
```
$ make -f testmake -k a b
echo "m4exit(\`1')" | m4 - | sed -e s/1/2/
echo "m4exit(\`1')" | m4 - > tmp1
make: *** [testmake:6: b] Error 1
```

With -o pipefail, both examples fail:
```
$ make -f testmake-pipefail -k a b
echo "m4exit(\`1')" | m4 - | sed -e s/1/2/
make: *** [testmake-pipefail:4: a] Error 1
echo "m4exit(\`1')" | m4 - > tmp1
make: *** [testmake-pipefail:6: b] Error 1
```

As far as I can see, any current pipe with any possibility of failure is not okay without -o pipefail.

pebenito: I'm fine with changing this to a pipe as long as the suggested `.SHELLFLAGS` are added.

Rules.modular
```diff
@@ -138,26 +146,25 @@ $(tmpdir)/global_bools.conf: M4PARAM += -D self_contained_policy
...
+$(tmpdir)/all_te_files.conf: | $(tmpdir)
 $(tmpdir)/all_te_files.conf: M4PARAM += -D self_contained_policy
 $(tmpdir)/all_te_files.conf: $(m4support) $(tmpdir)/generated_definitions.conf $(tmpdir)/all_interfaces.conf $(base_te_files) $(m4terminate)
 ifeq "$(strip $(base_te_files))" ""
 	$(error No enabled modules! $(notdir $(mod_conf)) may need to be generated by using "make conf")
 endif
-	@test -d $(tmpdir) || mkdir -p $(tmpdir)
 	$(verbose) $(M4) $(M4PARAM) -s $^ > $@
 
+$(tmpdir)/post_te_files.conf: | $(tmpdir) $(tmpdir)/all_te_files.conf
```
cgzones, line 163: This is a change of the dependency type for `$(tmpdir)/all_te_files.conf` from *check if older* to *check if exists*.

PR #395, opened 2021-07-14
virt: Defined virt_runtime_common_t type

jpds:
For the new common/system.token file in libvirt 4.7.0 and added permissions to virtd_t and virtlogd_t.

Modelled on: https://github.com/fedora-selinux/selinux-policy/commit/1f761d0bbdc08d21a91cdcbd1909ddb53858e354
libvirt change introducing this: https://gitlab.com/libvirt/libvirt/-/commit/cbfebfc74741a00bddf67b7fa10892b757fffd6

### DIFF
--- a/policy/modules/services/virt.te
+++ b/policy/modules/services/virt.te
@@ -129,6 +129,9 @@ type virt_image_t; # customizable
 virt_image(virt_image_t)
 files_mountpoint(virt_image_t)
 
+type virt_common_runtime_t;
+files_runtime_file(virt_common_runtime_t)
+
 type virt_content_t; # customizable
 virt_image(virt_content_t)
 userdom_user_home_content(virt_content_t)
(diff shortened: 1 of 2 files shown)


policy/modules/services/virt.te
```diff
@@ -143,6 +143,9 @@ mls_trusted_object(virt_log_t)
 type virt_runtime_t alias virt_var_run_t;
 files_runtime_file(virt_runtime_t)
 
+type virt_runtime_common_t alias virt_common_var_run_t;
```
pebenito, line 146: I'd prefer to have it named `virt_common_runtime_t`. Also, no need for the alias since there is nothing to be backwards compatibility issue.

pebenito (changes requested):

policy/modules/services/virt.te
```diff
@@ -143,6 +143,9 @@ mls_trusted_object(virt_log_t)
 type virt_runtime_t alias virt_var_run_t;
 files_runtime_file(virt_runtime_t)
 
+type virt_common_runtime_t;
+files_runtime_file(virt_common_runtime_t)
```
pebenito, line 147: Needs to move above the `virt_content_t` lines.

PR #403, opened 2021-08-10
remove MongoDB policies from ref policy

brushless-glitch:
Due to a [known issue](https://jira.mongodb.org/browse/SERVER-38704) with MongoDB reference policy which prevents mongod from starting on RHEL when SELinux is enabled, there is an effort underway to create an official SELinux policy for MongoDB. It will be open sourced in [official MongoDB repository](https://github.com/mongodb/mongo) and supported by MongoDB. This pull request is to remove MongoDB policy from reference repository so that there is no expectation that there is a working reference policy.


pebenito:
Red Hat/Fedora hard forked their policy from refpolicy many years ago. Do you have known issues with refpolicy's rules? Otherwise I'm not inclined to remove it.

brushless-glitch:
The issue is that the daemon will not start; but I see your point. If RHEL does not follow current repository, feel free to reject the PR

PR #404, opened 2021-08-31
chronyd.te: Grant chronyd_t access to the net_admin capability

jpds:
This is required for its `hwtimestamp` option, which otherwise returns:

    ioctl(SIOCSHWTSTAMP) failed : Operation not permitted

### DIFF
--- a/policy/modules/services/chronyd.te
+++ b/policy/modules/services/chronyd.te
@@ -5,6 +5,14 @@ policy_module(chronyd, 1.8.0)
 # Declarations
 #
 
+## <desc>
+##      <p>
+##      Determine whether chronyd can access NIC hardware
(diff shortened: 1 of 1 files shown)


pebenito (changes requested):
Please add a comment in the file for why this is needed.

pebenito:
Also, is this this required for chronyd in all cases?

jpds:
@pebenito Added a comment, and this is only needed when the `hwtimestamp` option is added to the chronyd configuration.

pebenito (changes requested):
In that case, please move the net_admin into a tunable called `chronyd_hwtimestamp`.

pebenito (approved):

PR #406, opened 2021-09-06
git: add policy for git application

0xC0ncord:
The git application cannot normally be used by confined users as it requires the `map` permission on files in a repository. To address this, create a policy for the application and allow users to transition to it. `git_client_role_template()` is used to create the access needed for users, and the permissions for the git client are separate from those of the git daemon and git CGI.

### DIFF
--- a/policy/modules/services/git.fc
+++ b/policy/modules/services/git.fc
@@ -1,4 +1,6 @@
 HOME_DIR/public_git(/.*)?		gen_context(system_u:object_r:git_user_content_t,s0)
+HOME_DIR/\.config/git(/.*)?		gen_context(system_u:object_r:git_xdg_config_t,s0)
+HOME_DIR/\.gitconfig	--	gen_context(system_u:object_r:git_xdg_config_t,s0)
 
 /usr/lib/git-core/git-daemon	--	gen_context(system_u:object_r:gitd_exec_t,s0)
 
--- a/policy/modules/services/git.if
+++ b/policy/modules/services/git.if
@@ -49,6 +49,73 @@ interface(`git_role',`
 
 ########################################
 ## <summary>
+##	Role access for Git client.
+## </summary>
+## <param name="role_prefix">
+##	<summary>
+##	The prefix of the user role (e.g., user
+##	is the prefix for user_r).
+##	</summary>
+## </param>
+## <param name="user_role">
+##	<summary>
+##	The role associated with the user domain.
+##	</summary>
+## </param>
+## <param name="user_domain">
+##	<summary>
+##	The type of the user domain.
+##	</summary>
+## </param>
+#
+template(`git_client_role_template',`
+	gen_require(`
+		attribute git_client_domain;
+		type git_exec_t, git_home_t, git_home_hook_t;
+	')
+
+	########################################
+	#
+	# Declarations
+	#
+
+	type $1_git_t, git_client_domain;
+	userdom_user_application_domain($1_git_t, git_exec_t)
+	role $2 types $1_git_t;
+
+	########################################
+	#
+	# Policy
+	#
+	
+	domtrans_pattern($3, git_exec_t, $1_git_t)
+
+	allow $3 git_home_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $3 git_home_t:file { manage_file_perms relabel_file_perms };
+	userdom_user_home_dir_filetrans($3, git_home_t, dir, ".git")
+
+	allow $3 git_home_hook_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $3 git_home_hook_t:file { exec_file_perms manage_file_perms relabel_file_perms };
+	filetrans_pattern($3, git_home_t, git_home_hook_t, dir, "hooks")
+
+	allow $3 $1_git_t:process { ptrace signal_perms };
+	ps_process_pattern($3, $1_git_t)
+
+	auth_use_nsswitch($1_git_t)
+
+	# allow userdomains to exec git hooks
+	exec_files_pattern($3, git_home_t, git_home_t)
+	# transition back to the user domain when executing git hooks
+	domtrans_pattern($1_git_t, git_home_t, $3)
+
+	# transition to ssh client domain when performing ssh operations
+	optional_policy(`
+		ssh_client_domtrans($1_git_t)
+	')
+')
+
+########################################
+## <summary>
 ##	Read generic system content files.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/services/git.te
+++ b/policy/modules/services/git.te
@@ -79,6 +79,15 @@ gen_tunable(git_system_use_cifs, false)
 ## </desc>
 gen_tunable(git_system_use_nfs, false)
 
+## <desc>
+##	<p>
+##	Determine whether Git client domains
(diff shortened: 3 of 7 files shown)


pebenito:
> The git application cannot normally be used by confined users as it requires the `map` permission on files in a repository. To address this, create a policy for the application and allow users to transition to it. `git_client_role_template()` is used to create the access needed for users, and the permissions for the git client are separate from those of the git daemon and git CGI.

What is the negative consequence of allowing map on the userdomain instead? It doesn't seem concerning to me.

0xC0ncord:
> What is the negative consequence of allowing map on the userdomain instead? It doesn't seem concerning to me.

I can't personally think of any, but maybe someone with a better understanding of the `map` permission may object. @doverride do you have any thoughts?

0xC0ncord:
I also want to point out that with this policy, git index files are given their own private type and we can more easily restrict access to only allow git and userdomains to read them.

doverride:
there is no instead. you can give shell domains map access but it is still useful to confine the client as it needs to create sockets for network access, git-grep actually needs execmem in some scenario (grep itself too though btw). but yes for me personally i like to protect the .git metadata and only allow git as a way in there (just like gpg is the only way into the private key)

pebenito:
Sorry, I'm not yet convinced. The main concerns for map are surrounding revocation, because you can't revoke the mmap by a policy or boolean change once the map is in place. I don't think network sockets alone are a reason to add a new domain, otherwise we'd have domains for stuff like wget.

0xC0ncord:
I think allowing userdomains to `map` their own home content should be fine to add, and might resolve the original issue I had with the git client not working as a result (I will test this). However, I still think there is some value in making git index files a private type.

0xC0ncord:
> I think allowing userdomains to `map` their own home content should be fine to add, and might resolve the original issue I had with the git client not working as a result (I will test this). However, I still think there is some value in making git index files a private type.

To add to this, git's xdg config file might also need to have a private type (I added this in this PR). The git config file may contain sensitive data such as credentials in plaintext, which would be another reason to confine this access.

policy/modules/services/git.if
```diff
@@ -47,6 +47,67 @@ interface(`git_role',`
...
+	#
+
+	allow $3 git_home_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $3 git_home_t:file { exec_file_perms manage_file_perms relabel_file_perms };
+	userdom_user_home_dir_filetrans($3, git_home_t, dir, ".git")
+
+	allow $3 $1_git_t:process { ptrace signal_perms };
+	ps_process_pattern($3, $1_git_t)
+
+	auth_use_nsswitch($1_git_t)
```
pebenito, line 98: This should be in the .te using `git_client_domain`.

policy/modules/services/git.if
```diff
@@ -47,6 +47,67 @@ interface(`git_role',`
...
+
+	# allow userdomains to exec git hooks
+	exec_files_pattern($3, git_home_t, git_home_t)
+	# transition back to the user domain when executing git hooks
+	domtrans_pattern($1_git_t, git_home_t, $3)
```
pebenito, line 103: I understand the logic of this, but it allows the domain to execute arbitrary code. It could write a `git_home_t` file and then execute it. Since it runs in the user's domain, it could access other secrets.

If we keep the line, please put at the top of the rule block (line 90)

pebenito (changes requested):
> The git config file may contain sensitive data such as credentials in plaintext, which would be another reason to confine this access.

This is a compelling reason.

policy/modules/services/git.fc
```diff
@@ -1,4 +1,6 @@
 HOME_DIR/public_git(/.*)?		gen_context(system_u:object_r:git_user_content_t,s0)
+HOME_DIR/\.config/git		gen_context(system_u:object_r:git_xdg_config_t,s0)
```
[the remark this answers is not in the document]
0xC0ncord: The git config can exist in two locations: in `$HOME/.gitconfig` and `$HOME/.config/git/config`, with the latter taking precedence. I'm not sure if anything else goes in that dir except for the config file.

policy/modules/services/git.if
0xC0ncord: I tried that previously, but was getting a compiler error because `auth_use_nsswitch()` uses `typeattribute $1 nsswitch_domain`, and `git_client_domain` is another attribute.

policy/modules/services/git.if
0xC0ncord: I would like to keep this logic because it allows the git client to operate in a way that I as a user would expect. Do you think putting this behind a tunable makes more sense?

0xC0ncord: Alternatively we could create a separate type for git hooks and only allow those to be executed, but not written to by the client.

policy/modules/services/git.if
pebenito: Ah, yes, I forgot that.

policy/modules/services/git.fc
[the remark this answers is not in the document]
pebenito: In that case, shouldn't the regex be `$HOME/.config/git(/.*)?`?

policy/modules/services/git.if
pebenito: I think a type for hooks is a better option.

policy/modules/services/git.fc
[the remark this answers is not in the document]
0xC0ncord: You are correct.

policy/modules/services/git.if
0xC0ncord: I have added the `git_home_hook_t` type for git hooks. This type is pretty much identical to `git_home_t` except `git_client_domain` does not have write access to it (and we also remove git's access to execute `git_home_t`). Unfortunately, we can't easily allow the git client to create these files because they are copied from `/usr/share/git-core/templates`, and thus `write` access is required. Instead, a user will need to manually relabel git hooks to this type in order to allow git to execute them.

policy/modules/services/git.te
```diff
@@ -92,12 +101,25 @@ type git_session_t, git_daemon;
...
 type git_sys_content_t;
 files_type(git_sys_content_t)
 
 type git_user_content_t;
 userdom_user_home_content(git_user_content_t)
 
+type git_home_t;
+userdom_user_home_content(git_home_t)
+
+type git_home_hook_t;
```
pebenito, line 117: Is there anywhere else the hook could be? If not, git_hook_t would be fine.

policy/modules/services/git.te
```diff
@@ -273,3 +295,55 @@ files_read_usr_files(git_daemon)
...
+
+########################################
+#
+# Git client policy
+#
+
+allow git_client_domain self:fifo_file rw_fifo_file_perms;
+
+corenet_tcp_connect_git_port(git_client_domain)
+corenet_tcp_connect_http_port(git_client_domain)
```
pebenito, line 307: Needs to move down by corecmd

policy/modules/services/git.te
0xC0ncord: I anticipate that in the future there may be a need for a hook type in system git repositories, that being a `git_sys_hook_t`.

PR #412, opened 2021-10-08
Hookup attribute for user applications to run other applications

### LINKED ISSUE #365: Proposal: add an attribute for application execute and transition permissions

0xC0ncord:
This is the second part of the effort previously started in #381.

This PR modifies the calls to every `_role()` and `_role_template()` to use the new `$1_application_exec_domain` where appropriate, while still standardizing these role calls to use this format: `myapp_role[_template](user, user_t, user_application_exec_domain, user_r)`. This PR does not touch any role interfaces that do not have any callers.

`$1_application_exec_domain` only gets permissions needed to execute and transition to other applications, as well as using `stream_connect()` on them if it was previously given to userdomains. `$1_application_exec_domain` does not get access to read or modify application data explicitly.

Additionally, `$1_systemd_t` gets the permissions needed to monitor applications started by users in order to capture output to `journald`, check their state, and so on.

Lastly, `sudo`, `su`, and `shutdown` get a new tunable (default off) to control whether `$1_application_exec_domain` should have access to execute these administrative applications.

Fixes #365

### DIFF
--- a/policy/modules/admin/shutdown.if
+++ b/policy/modules/admin/shutdown.if
@@ -28,13 +28,23 @@
 #
 template(`shutdown_role',`
 	gen_require(`
+		attribute_role shutdown_roles;
 		type shutdown_t;
 	')
 
-	shutdown_run($3, $4)
+	roleattribute $4 shutdown_roles;
 
-	allow $3 shutdown_t:process { ptrace signal_perms };
-	ps_process_pattern($3, shutdown_t)
+	tunable_policy(`shutdown_allow_user_exec_domains',`
+		shutdown_domtrans($3)
+
+		allow $3 shutdown_t:process { ptrace signal_perms };
+		ps_process_pattern($3, shutdown_t)
+	',`
+		shutdown_domtrans($2)
+
+		allow $2 shutdown_t:process { ptrace signal_perms };
+		ps_process_pattern($2, shutdown_t)
+	')
 
 	optional_policy(`
 		systemd_user_app_status($1, shutdown_t)
--- a/policy/modules/admin/shutdown.if
+++ b/policy/modules/admin/shutdown.if
@@ -4,26 +4,41 @@
 ## <summary>
 ##	Role access for shutdown.
 ## </summary>
-## <param name="role">
+## <param name="role_prefix">
 ##	<summary>
-##	Role allowed access.
+##	The prefix of the user role (e.g., user
+##	is the prefix for user_r).
 ##	</summary>
 ## </param>
-## <param name="domain">
+## <param name="user_domain">
 ##	<summary>
 ##	User domain for the role.
 ##	</summary>
 ## </param>
+## <param name="user_exec_domain">
+##	<summary>
+##	User exec domain for execute and transition access.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access
+##	</summary>
+## </param>
 #
-interface(`shutdown_role',`
+template(`shutdown_role',`
 	gen_require(`
 		type shutdown_t;
 	')
 
-	shutdown_run($2, $1)
+	shutdown_run($3, $4)
 
-	allow $2 shutdown_t:process { ptrace signal_perms };
-	ps_process_pattern($2, shutdown_t)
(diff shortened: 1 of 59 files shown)


policy/modules/admin/shutdown.if
```diff
@@ -4,26 +4,51 @@
 ## <summary>
 ##	Role access for shutdown.
 ## </summary>
-## <param name="role">
+## <param name="role_prefix">
```
pebenito, line 7: Changing the interface to use a prefix turns it into a template.

policy/modules/apps/mono.if
```diff
@@ -63,7 +63,7 @@ template(`mono_role_template',`
 	optional_policy(`
 		fs_dontaudit_rw_tmpfs_files($1_mono_t)
 
-		xserver_role($1_r, $1_mono_t)
+		xserver_role($1, $1_mono_t, $1_application_exec_domain, $1_r)
```
pebenito, line 66: This needs the same changes you  made elsewhere, instead of using `$1_application_exec_domain` directly.

policy/modules/apps/pulseaudio.if
```diff
@@ -55,12 +71,9 @@ interface(`pulseaudio_role',`
 #
 interface(`pulseaudio_domtrans',`
 	gen_require(`
-		attribute pulseaudio_client;
 		type pulseaudio_t, pulseaudio_exec_t;
 	')
 
-	typeattribute $1 pulseaudio_client;
```
pebenito, line 62: Why is this change needed?

policy/modules/apps/wine.if
```diff
@@ -97,7 +97,7 @@ template(`wine_role_template',`
 	')
 
 	optional_policy(`
-		xserver_role($1_r, $1_wine_t)
+		xserver_role($1, $1_wine_t, $1_application_exec_domain, $1_r)
```
pebenito, line 100: Same as above.

pebenito (changes requested):

pebenito:
From the source diff it seems ok, but I noticed something concerning from sediff:
```
$ sediff policy.33{.orig,} | grep xguest | wc -l
175
```
These are all additions.

0xC0ncord:
> From the source diff it seems ok, but I noticed something concerning from sediff:
> 
> ```
> $ sediff policy.33{.orig,} | grep xguest | wc -l
> 175
> ```
> 
> These are all additions.

It seems that most of these new rules are to `xguest_wm_t`, and most of those are allowing `xguest_wm_t` to transition to various httpd user scripts. Everything else looks fine to me.

I'm not sure what the use case is for xguest having the ability to run httpd user scripts. If you think we should lock this down, I can add a template for httpd user script guest access, or we can remove this call altogether: https://github.com/SELinuxProject/refpolicy/blob/master/policy/modules/roles/xguest.te#L89

pebenito:
Lets remove the access. I don't see a need for a guest to have this access. If there is a good reason, we can reconsider.

0xC0ncord:
OK. I added another commit to remove `apache` access from xguest. I also went ahead and did the same thing for normal guest for the same reason.

pebenito (approved):

PR #415, opened 2021-10-13
mcs: pull in changes from Fedora policy

0xC0ncord:
Pull in some changes from the Fedora policy's MCS constraints.

Most notably, the MCS override attributes were deprecated in favor of `mcs_constrained_type`. This means that domains will have unchecked access to objects with categories UNLESS the domain is `mcs_constrained_type`. This alleviates confusion between the MCS overrides and `mcs_constrained_type` to imply that a domain must be MCS-constrained to have MCS checks at all.

Other changes include additional constraints to miscellaneous IPC objects, node "write" operations, and netif egress/ingress operations.

### DIFF
--- a/policy/mcs
+++ b/policy/mcs
@@ -69,53 +69,47 @@ gen_levels(1,mcs_num_cats)
 #  - /proc/pid operations are not constrained.
 
 mlsconstrain file { read ioctl lock execute execute_no_trans }
-	(( h1 dom h2 ) or ( t1 == mcsreadall ) or
-	(( t1 != mcs_constrained_type ) and (t2 == domain)));
+	(( h1 dom h2 ) or ( t1 != mcs_constrained_type ));
 
 mlsconstrain file { write setattr append unlink link rename }
-	(( h1 dom h2 ) or ( t1 == mcswriteall ) or
-	(( t1 != mcs_constrained_type ) and (t2 == domain)));
+	(( h1 dom h2 ) or ( t1 != mcs_constrained_type ));
 
 mlsconstrain dir { search read ioctl lock }
-	(( h1 dom h2 ) or ( t1 == mcsreadall ) or
-	(( t1 != mcs_constrained_type ) and (t2 == domain)));
+	(( h1 dom h2 ) or ( t1 != mcs_constrained_type ));
 
 mlsconstrain dir { write setattr append unlink link rename add_name remove_name }
-	(( h1 dom h2 ) or ( t1 == mcswriteall ) or
-	(( t1 != mcs_constrained_type ) and (t2 == domain)));
+	(( h1 dom h2 ) or ( t1 != mcs_constrained_type ));
 
 mlsconstrain fifo_file { open }
-	(( h1 dom h2 ) or ( t1 == mcsreadall ) or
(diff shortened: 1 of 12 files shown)


cgzones:
SELinux by design has the principle *everything is denied except explicitly allowed*. (Within refpolicy this is already circumvented in UBAC by the choice of `ubac_constrained_type`.) So the switch from deny everything except `mcs*all` to allow everything except `mcs_constrained_type` is quite a paradigm change (which I am not a huge fan of). Also it seems only `netlabel_peer_t` is constrained so far, but probably far more domains, especially `userdomain`'s and user application domains, should be.

policy/mcs
```diff
@@ -138,6 +135,9 @@ mlsconstrain { db_database db_schema db_table db_sequence db_view db_procedure d
 mlsconstrain { db_tuple } { insert relabelto }
 	(( h1 dom h2 ) and ( l2 eq h2 ));
 
+mlsconstrain context contains
+	(( h1 dom h2 ) and ( l1 domby l2 ));
+
```
pebenito, line 140: This needs to move up to line 126 since it is not a db constraint. Also, why not have `or ( t1 != mcs_constrained_type )` condition too?

policy/mcs
```diff
@@ -138,6 +135,9 @@ mlsconstrain { db_database db_schema db_table db_sequence db_view db_procedure d
 mlsconstrain { db_tuple } { insert relabelto }
 	(( h1 dom h2 ) and ( l2 eq h2 ));
```
pebenito, line 136: I would think all the db classes should have the `or ( t1 != mcs_constrained_type )` condition too.

pebenito (changes requested):

policy/mcs
0xC0ncord: I will move it up.

Not sure why it doesn't have `or ( t1 != mcs_constrained_type )`, as I pulled it directly from Fedora's policy as-is. It was added in this nondescript commit: https://github.com/fedora-selinux/selinux-policy/commit/f9ba5d862027233ab129ea021dfa340437d3b08e

Maybe someone else might know.
cc @doverride

policy/mcs
0xC0ncord: Same as above. I pulled these directly out of Fedora's policy. The majority of these were added in: https://github.com/fedora-selinux/selinux-policy/commit/82c32d5cf4d6df42ca16513cde2b3a28df41015f

doverride:

>
> -------------------------------------------------------------------------------------------
>
> In policy/mcs:
>
>> +mlsconstrain context contains
> +	(( h1 dom h2 ) and ( l1 domby l2 ));
> +
>
> I will move it up.
>
> Not sure why it doesn't have or ( t1 != mcs_constrained_type ), as I pulled it directly
> from Fedora's policy as-is. It was added in this nondescript commit:
> ***@***.***
>
> Maybe someone else might know.
> cc @doverride

Don't know. Should likely be there. Then again not sure whether its even
worth constraining that in the first place.

>
> —
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

pebenito (approved):
The only comments were in this PR, none from the mail list.

PR #429, opened 2021-11-07
More various fixes

0xC0ncord:
Fixes for various services including:

- Allowing several services to read certbot TLS certificates and keys
- Allow dovecot to tell spamd to learn spam or ham
- Let sysadm interact with usbguard at runtime
- Let sysadm interact with systemd-networkd
- Let sshd create polyinstantiated tmp directories for remote user logins
- and a couple others


policy/modules/services/bind.te
```diff
@@ -81,7 +81,8 @@ allow named_t self:fifo_file rw_fifo_file_perms;
 allow named_t self:unix_stream_socket { accept listen };
 allow named_t self:tcp_socket { accept listen };
 
-allow named_t dnssec_t:file read_file_perms;
+filetrans_pattern(named_t, named_conf_t, dnssec_t, dir, "cache")
+allow named_t dnssec_t:file manage_file_perms;
```
pebenito, line 85: Nit: the `allow` should be above the `filetrans_pattern`.

policy/modules/services/spamassassin.if
```diff
@@ -196,6 +196,7 @@ interface(`spamassassin_exec_client',`
 
 	corecmd_search_bin($1)
 	can_exec($1, spamc_exec_t)
+	read_lnk_files_pattern($1, spamc_exec_t, spamc_exec_t)
```
pebenito, line 199: There should be no `spamc_exec_t` symlinks. The labeling should be fixed instead.

pebenito (changes requested):

policy/modules/services/spamassassin.if
0xC0ncord: OK. I will split that out into its own commit, since it is primarily for `rspamd`.

pebenito (approved):

0xC0ncord:
Fixed the documentation in the `certbot_read_lib` interface just now. Caught it at the last second.

PR #432, opened 2021-11-12
Report warning on duplicate definition of interface

vmojzis:
Duplicate definition of interface or template should only be a warning,
since it does not block the compilation.
The warning should also specify which file contains the original
definition.

Old:
 ipa.if:284: Error: duplicate definition of
  ipa_cert_filetrans_named_content(). Original definition on 284.
New:
 ipa.if:284: Warning: duplicate definition of
  ipa_cert_filetrans_named_content(). Original definition on
  /usr/share/selinux/devel/include/contrib/ipa.if:284.

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>

### DIFF
--- a/policy/support/loadable_module.spt
+++ b/policy/support/loadable_module.spt
@@ -60,7 +60,7 @@ define(`policy_m4_comment',`
 # template(name,rules)
 #
 define(`template',` dnl
-	ifdef(`$1',`refpolicyerr(`duplicate definition of $1(). Original definition on '$1.) define(`__if_error')',`define(`$1',__line__)') dnl
+	ifdef(`$1',`refpolicyerr(`duplicate definition of $1(). Original definition on '$1.) define(`__if_error')',`define(`$1',__file__:__line__)') dnl
 	`define(`$1',` dnl
 	pushdef(`policy_call_depth',incr(policy_call_depth)) dnl
 	policy_m4_comment(policy_call_depth,begin `$1'(dollarsstar)) dnl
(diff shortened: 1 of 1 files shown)


pebenito:
Please elaborate why you feel this should be a warning. If you have multiple definitions, you may not get the rules you want, which sounds like an error condition to me.

vmojzis:
I'd expect an error to be a non-recoverable issue, or an issue preventing some major functionality. In this case yes, you may not get the rules you want. But I would expect users to mostly encounter this message when the new definition is in working directory. In which case you get the rules specified locally (most likely what you want) and the resulting policy works perfectly fine (as if the duplicate definition didn't exist).

pebenito:
My discomfort is more about all the other build scenarios. If a user builds the entire policy for an embedded system in their CI, they may never see the message if this is a warning. They may be getting more permissions than they want.

cgzones:
> If a user builds the entire policy for an embedded system in their CI, they may never see the message if this is a warning.

As last resort with the build configuration `WERROR=y` it is still a hard error.

pebenito:
I considered that, but then I think I'd want WERROR=y to be the default. The problem with that is that deprecated interfaces warnings would become errors, and those definitely are warnings, since they still have rules. No need to have them error out by default.

vmojzis:
So, should I change the commit to only add the file containing the original definition (and leave the message as error)?

pebenito:
Yes, please.

PR #442, opened 2021-11-29
samba and ntp fixes

yizhao1:


pebenito:
What fails when `net_admin` is denied on samba?

yizhao1:
This avc message only happens on systemd-based system but not on sysvinit-based system. And seems it doesn't  impact the smb/nmb daemon running.

yizhao1:
> What fails when `net_admin` is denied on samba?

This avc message only happens on systemd-based system but not on sysvinit-based system. And seems it doesn't impact the smb/nmb daemon running.

pebenito (changes requested):
Please change the samba `net_admin` to dontaudits.

pebenito (approved):

PR #451, opened 2021-12-21
Kmod fixes

yizhao1:


policy/modules/system/modutils.te
```diff
@@ -40,6 +40,8 @@ dontaudit kmod_t self:capability sys_admin;
 
 allow kmod_t self:udp_socket create_socket_perms;
 allow kmod_t self:rawip_socket create_socket_perms;
+allow kmod_t self:lockdown confidentiality;
```
pebenito, line 43: I'd prefer not to add anymore `lockdown` rules, since the checks will be removed in newer kernels, so it should be removed.

Optionally, remove all lockdown rules from the policy and then add `allow domain self:lockdown { confidentiality integrity };` in `domain.te`.

pebenito (changes requested):

policy/modules/system/modutils.te
yizhao1: Thanks. It's OK for me. How about the 2nd patch ?

pebenito: The key write is ok.

pebenito (approved):

PR #454, opened 2022-01-04
obj_perm_sets.spt: Fixed typo in rw_netlink_socket_perms.

jpds:
Untested, but I think I came across this typo whilst writing policy for something else.

### DIFF
--- a/policy/support/obj_perm_sets.spt
+++ b/policy/support/obj_perm_sets.spt
@@ -100,7 +100,7 @@ define(`create_netlink_socket_perms', `{ create_socket_perms nlmsg_read nlmsg_wr
 #
 # Permissions for using netlink sockets for operations that modify state.
 #
-define(`rw_netlink_socket_perms', `{ create_socket_perms nlmsg_read nlmsg_write }')
+define(`rw_netlink_socket_perms', `{ rw_socket_perms nlmsg_read nlmsg_write }')
 
 #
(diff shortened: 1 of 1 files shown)


pebenito:
My main concern is with the domains that use this macro. The `create_socket_perms` should be added to the existing macro users, unless we're confident it isn't needed (per each domain).

pebenito (changes requested):
I think we can take this change if we add `create_socket_perms` to all of the current users.

jpds:
@pebenito I've replaced `rw_netlink_socket_perms` with `create_netlink_socket_perms` everywhere,

policy/modules/kernel/kernel.if
```diff
@@ -380,7 +380,7 @@ interface(`kernel_rw_netlink_audit_sockets',`
 		type kernel_t;
 	')
 
-	allow $1 kernel_t:netlink_audit_socket { rw_netlink_socket_perms };
+	allow $1 kernel_t:netlink_audit_socket { create_netlink_socket_perms };
 ')
```
pebenito, line 384: Since this is explicitly `rw`, I think we shouldn't change this. Since it's not a `self` permission and no setsockcreate, the create wouldn't be used.

pebenito (changes requested):

pebenito (approved):

PR #458, opened 2022-01-06
RFE: convert README to README.md

pcmoore:
This is a pretty basic conversion, but it is better than the plaintext README when viewed in the GitHub web UI.

Signed-off-by: Paul Moore <paul@paul-moore.com>


pcmoore:
The rendered README.md can be viewed here:

* https://github.com/pcmoore/misc-refpolicy/tree/working-docs#readme

pebenito (changes requested):
Thanks for the patch. I'd prefer to keep the tabular format that the README has.

pcmoore:
I understand the desire to stick with a tabular formatting, but unfortunately that doesn't render well at all in Markdown (it appears that all horizontal whitespace ends up rendering as a single space).  I tried going with a table format to try and stick with the spirit behind the tab-based formatting but Markdown's limited table support made it ugly both when rendered and as a raw text file.

I'm okay if you want to just close/shelve this PR and/or use this as a basis for a future conversion; you're not going to hurt my feelings either way :)  I just had a few moments near the end of my day and thought I would take a crack at a rough conversion.

PR #459, opened 2022-01-07
Remove MCS categories from default users

### LINKED ISSUE #453: Consider removing MCS categories from default users contexts in seusers
We should consider changing the default seusers to not include MCS categories since #415 means that the user domains no longer need them to access resources with MCS categories.

cc @doverride

0xC0ncord:
Remove MCS categories from the default users and remove most MCS ranged transitions as a result.

I didn't see any ranged transitions that I think needed to stay, as none of the domains I removed them from are `mcs_constrained()`.

Fixes #453

### DIFF
--- a/policy/modules/admin/samhain.te
+++ b/policy/modules/admin/samhain.te
@@ -35,10 +35,6 @@ role samhain_roles types samhain_t;
 samhain_service_template(samhaind)
 init_system_domain(samhaind_t, samhain_exec_t)
 
-ifdef(`enable_mcs',`
-	init_ranged_system_domain(samhaind_t, samhain_exec_t, mcs_systemhigh)
-')
-
 ifdef(`enable_mls',`
 	init_ranged_system_domain(samhaind_t, samhain_exec_t, mls_systemhigh)
 ')
--- a/policy/modules/apps/vmware.te
+++ b/policy/modules/apps/vmware.te
@@ -42,10 +42,6 @@ userdom_user_tmp_file(vmware_tmp_t)
 type vmware_tmpfs_t;
 userdom_user_tmpfs_file(vmware_tmpfs_t)
 
-ifdef(`enable_mcs',`
-	init_ranged_daemon_domain(vmware_host_t, vmware_host_exec_t, s0 - mcs_systemhigh)
-')
-
 optional_policy(`
 	wm_application_domain(vmware_t, vmware_exec_t)
 ')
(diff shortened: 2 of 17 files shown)


doverride (commented):
I am not sure about this but if this file is also used for MLS then one should probably ensure that this change does not affect MLS? (ifdef_MLS/ifndef_MLS)

0xC0ncord:
> I am not sure about this but if this file is also used for MLS then one should probably ensure that this change does not affect MLS? (ifdef_MLS/ifndef_MLS)

Good point. I'll look into it.

0xC0ncord:
I checked all of the `ifdef_MLS/ifndef_MLS` definitions and didn't see anything that stuck out as potentially problematic to me.

0xC0ncord:
One thing maybe to consider is that most of the MLS range transitions are to `mls_systemhigh` which includes MCS categories. Maybe those MCS categories should be removed as well for any domains that aren't MCS-constrained?

doverride:

> One thing maybe to consider is that most of the MLS range transitions are to mls_systemhigh which
> includes MCS categories. Maybe those MCS categories should be removed as well for any domains that
> aren't MCS-constrained?

No, if it is in a ifdef_MLS block then it need to stay the same
regardless. (MLS needs categories)

As for the other comment about the "users" file. I think there may be a
misunderstanding.

take for example this:
https://github.com/SELinuxProject/refpolicy/blob/master/policy/users#L29

I was suggesting that you might need something like this

ifdef(`enable_mls',`
gen_user(sysadm_u, sysadm, sysadm_r, s0, s0 - mls_systemhigh, mcs_allcats)
',`
gen_user(sysadm_u, sysadm, sysadm_r, s0, s0)
')

similar for the other entries in that file.

But again I might be wrong

Cc. @pebenito

>
> —
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
>

-- 
gpg --locate-keys ***@***.***
Dominick Grift

pebenito:
> I was suggesting that you might need something like this
> 
> ifdef(`enable_mls',`
> gen_user(sysadm_u, sysadm, sysadm_r, s0, s0 - mls_systemhigh, mcs_allcats)
> ',`
> gen_user(sysadm_u, sysadm, sysadm_r, s0, s0)
> ')
> 
> similar for the other entries in that file.

That's not necessary. gen_user() already handles that; see the comments earlier in the file.

pebenito (approved):

PR #462, opened 2022-01-13
Systemd updates including systemd-homed and systemd-userdbd.

pebenito:
- Add systemd-homed and systemd-userdbd.
- Various other systemd fixes.

The systemd-homed policy has only had the most basic testing and is just a starting point.

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -29,6 +29,8 @@
 /usr/lib/systemd/systemd-binfmt		--	gen_context(system_u:object_r:systemd_binfmt_exec_t,s0)
 /usr/lib/systemd/systemd-cgroups-agent	--	gen_context(system_u:object_r:systemd_cgroups_exec_t,s0)
 /usr/lib/systemd/systemd-coredump	--	gen_context(system_u:object_r:systemd_coredump_exec_t,s0)
+/usr/lib/systemd/systemd-homed  	--	gen_context(system_u:object_r:systemd_homed_exec_t,s0)
+/usr/lib/systemd/systemd-homework       --      gen_context(system_u:object_r:systemd_homework_exec_t,s0)
 /usr/lib/systemd/systemd-hostnamed	--	gen_context(system_u:object_r:systemd_hostnamed_exec_t,s0)
 /usr/lib/systemd/systemd-localed	--	gen_context(system_u:object_r:systemd_locale_exec_t,s0)
 /usr/lib/systemd/systemd-logind		--	gen_context(system_u:object_r:systemd_logind_exec_t,s0)
@@ -43,6 +45,8 @@
 /usr/lib/systemd/systemd-update-done	--	gen_context(system_u:object_r:systemd_update_done_exec_t,s0)
 /usr/lib/systemd/systemd-user-runtime-dir	--	gen_context(system_u:object_r:systemd_user_runtime_dir_exec_t,s0)
 /usr/lib/systemd/systemd-user-sessions	--	gen_context(system_u:object_r:systemd_sessions_exec_t,s0)
+/usr/lib/systemd/systemd-userdbd	--	gen_context(system_u:object_r:systemd_userdbd_exec_t,s0)
+/usr/lib/systemd/systemd-userwork	--	gen_context(system_u:object_r:systemd_userdbd_exec_t,s0)
 
 # Systemd unit files
 HOME_DIR/\.config/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home_t,s0)
@@ -62,6 +66,7 @@ HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data
 /usr/lib/systemd/system/systemd-networkd.*		gen_context(system_u:object_r:systemd_networkd_unit_t,s0)
 /usr/lib/systemd/system/systemd-rfkill.*	--	gen_context(system_u:object_r:systemd_rfkill_unit_t,s0)
 /usr/lib/systemd/system/systemd-socket-proxyd\.service	--	gen_context(system_u:object_r:systemd_socket_proxyd_unit_file_t,s0)
+/usr/lib/systemd/system/systemd-userdbd\.(service|socket)		--	gen_context(system_u:object_r:systemd_userdbd_unit_t,s0)
 /usr/lib/systemd/system/user@\.service	--	gen_context(system_u:object_r:systemd_user_manager_unit_t,s0)
 
 /usr/share/factory(/.*)?	gen_context(system_u:object_r:systemd_factory_conf_t,s0)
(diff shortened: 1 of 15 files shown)


policy/modules/system/systemd.fc
```diff
@@ -29,6 +29,8 @@
 /usr/lib/systemd/systemd-binfmt		--	gen_context(system_u:object_r:systemd_binfmt_exec_t,s0)
 /usr/lib/systemd/systemd-cgroups-agent	--	gen_context(system_u:object_r:systemd_cgroups_exec_t,s0)
 /usr/lib/systemd/systemd-coredump	--	gen_context(system_u:object_r:systemd_coredump_exec_t,s0)
+/usr/lib/systemd/systemd-homed  	--	gen_context(system_u:object_r:systemd_homed_exec_t,s0)
+/usr/lib/systemd/systemd-homework   --  gen_context(system_u:object_r:systemd_homed_exec_t,s0)
```
doverride, line 33: s/systemd_homed_exec_t/systemd_homework_exec_t/

policy/modules/system/systemd.te
```diff
@@ -115,6 +115,28 @@ typealias systemd_generator_t alias { systemd_fstab_generator_t systemd_gpt_gene
...
+type systemd_homed_exec_t;
+init_daemon_domain(systemd_homed_t, systemd_homed_exec_t)
+
+type systemd_homework_t;
+type systemd_homework_exec_t;
+domain_type(systemd_homework_t)
+domain_entry_file(systemd_homework_t, systemd_homework_exec_t)
+role system_r types systemd_homework_t;
+
+type systemd_homed_runtime_t;
```
doverride, line 128: I could not find and corresponding fc specs for this and possibly others. This may or may not affect any existing unix_stream_connect_userdbd() calls (as the socket may or may not have been mislabeled at the time)

pebenito: It should work fine without them, as I intentionally test with as few fcs as possible to make sure clean installs create files with correct labels. I forgot to add them after I finished.

policy/modules/system/systemd.if
```diff
@@ -1046,12 +1064,13 @@ interface(`systemd_manage_userdb_runtime_sock_files', `
...
 	gen_require(`
-		type systemd_userdb_runtime_t;
+		type systemd_userdbd_t, systemd_userdbd_runtime_t;
 	')
 
 	init_search_runtime($1)
-	allow $1 systemd_userdb_runtime_t:dir list_dir_perms;
-	allow $1 systemd_userdb_runtime_t:sock_file write_sock_file_perms;
+	allow $1 systemd_userdbd_runtime_t:dir list_dir_perms;
+	allow $1 systemd_userdbd_runtime_t:sock_file write_sock_file_perms;
```
doverride, line 1072: The write_sock_file_perms rule is redundant and implied with the `unix_stream_connect_pattern()` below.

policy/modules/system/systemd.te
```diff
@@ -489,6 +520,125 @@ optional_policy(`
...
+allow systemd_homed_t self:netlink_kobject_uevent_socket create_socket_perms;
+
+nnp_domtrans_pattern(systemd_homed_t, systemd_homework_exec_t, systemd_homework_t)
+
+allow systemd_homed_t systemd_homed_tmpfs_t:file manage_file_perms;
+fs_tmpfs_filetrans(systemd_homed_t, systemd_homed_tmpfs_t, file)
+
+manage_sock_files_pattern(systemd_homed_t, systemd_userdbd_runtime_t, systemd_homed_runtime_t)
+manage_dirs_pattern(systemd_homed_t, systemd_homed_runtime_t, systemd_homed_runtime_t)
+filetrans_pattern(systemd_homed_t, systemd_userdbd_runtime_t, systemd_homed_runtime_t, sock_file)
```
doverride, line 537: This is highly subjective but I thought I at least mention it briefly.

I am not sure if it is worth the trouble to label the content in `/run/systemd/userdb` with private types. Mainly because most of these sockets are used for name and group resolving (whether the name is dynamic, a containerr name, or a name of a user managed by systemd-homed) eventually it boils down to either nss-user/group equivalent resolving or nss-hosts equivalent resolving.

The userdbd socket (or atleast one of them) is a notable exception to this rule as it is used to provide the functionality that systemd-userdbctl provides either directly or indirectly AFAIK.

Regardless though the socket end points are always different and so you  could just allow writes to all sockets in `/run/systemd/userdb` for simplicity and then still use the `connectto` pattern to narrow down how the socket writes can effectively be used.

Its a matter of taste, this is not wrong in any way but it is expensive

policy/modules/system/systemd.te
```diff
@@ -489,6 +520,125 @@ optional_policy(`
...
+')
+
+#######################################
+#
+# systemd-homework policy
+#
+
+allow systemd_homework_t self:capability { chown fowner fsetid sys_admin };
+dontaudit systemd_homework_t self:capability sys_resource;
+allow systemd_homework_t self:key { search write };
```
doverride, line 587: This is also interesting and a matter of taste. The key access is (also) related to device mapper (cryptsetupl/uks2) That code essentially does two things.

1. it broadly searches keyrings
2. it writes its own keyring

So instead of what you have above I would do:

```
allow systemd_homework_t self:key write;

domain_search_all_keyrings(systemd_homework_t)
```

Alternatively you could, I guess, use `domain_dontaudit_search_all_keyrings(systemd_homework_t)` but not sure if that is worth a dontaudit rule or whetter that affects anything.

policy/modules/admin/aide.te
```diff
@@ -58,3 +58,7 @@ tunable_policy(`aide_mmap_files',`
 optional_policy(`
 	seutil_use_newrole_fds(aide_t)
 ')
+
+optional_policy(`
+	systemd_stream_connect_userdb(aide_t)
```
doverride, line 63: I would for now not address the stream connects because chances are that this can be implemented in a more efficient and scalable way eventually.

But, and I an not 100% sure of this, I vaguely recall that this might or might not be proper..

doverride:
Unfortunately I lost my systemd-homed/systemd-userdbd policy when i abandoned my dssp3 project and so I do not have anything to compare this with. I actually used this functionality for a short while but I forgot quite a bit about the specific of my policy implementation. It is clear to me though that this iteration only scratches the surface (which is fine but I do believe that for example the `systemd-stream_connect_userdbd()` related rules are probably pre-mature.

policy/modules/system/systemd.fc
```diff
@@ -29,6 +29,8 @@
 /usr/lib/systemd/systemd-binfmt		--	gen_context(system_u:object_r:systemd_binfmt_exec_t,s0)
 /usr/lib/systemd/systemd-cgroups-agent	--	gen_context(system_u:object_r:systemd_cgroups_exec_t,s0)
 /usr/lib/systemd/systemd-coredump	--	gen_context(system_u:object_r:systemd_coredump_exec_t,s0)
+/usr/lib/systemd/systemd-homed  	--	gen_context(system_u:object_r:systemd_homed_exec_t,s0)
```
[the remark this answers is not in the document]
pebenito: If this module is split up, it would be in a separate PR.

policy/modules/admin/aide.te
pebenito: I'll keep it until we can determine the pattern and update it or revert if it is found to be improper.

pebenito: Nevermind, I'll take it out.

policy/modules/system/systemd.te
pebenito: Makes sense.

policy/modules/system/systemd.te
pebenito: I'll keep it as-is. It's a small size and we can revise to make it only the sockets that filetrans in the future, if necessary.

PR #468, opened 2022-01-28
node_exporter: Added initial policy

jpds:
Policy for https://github.com/prometheus/node_exporter (also goes with https://github.com/prometheus/node_exporter/pull/2205).

@0xC0ncord also tested this on systemd for me.

### DIFF
--- a/policy/modules/services/node_exporter.fc
+++ b/policy/modules/services/node_exporter.fc
@@ -1,0 +1,6 @@
+/run/node_exporter\.pid	--	gen_context(system_u:object_r:node_exporter_runtime_t,s0)
+
+/usr/sbin/node_exporter	--	gen_context(system_u:object_r:node_exporter_exec_t,s0)
+
+/var/lib/node_exporter(/.*)?	gen_context(system_u:object_r:node_exporter_var_lib_t,s0)
+/var/log/node_exporter(/.*)?	gen_context(system_u:object_r:node_exporter_log_t,s0)
--- a/policy/modules/services/node_exporter.te
+++ b/policy/modules/services/node_exporter.te
@@ -1,0 +1,73 @@
+policy_module(node_exporter)
+
+########################################
+#
+# Declarations
+#
+
+type node_exporter_t;
+type node_exporter_exec_t;
+init_daemon_domain(node_exporter_t, node_exporter_exec_t)
+
+type node_exporter_runtime_t;
+files_runtime_file(node_exporter_runtime_t)
+
+type node_exporter_var_lib_t;
+files_type(node_exporter_var_lib_t)
+
+type node_exporter_log_t;
+logging_log_file(node_exporter_log_t)
+
+########################################
+#
+# Local policy
+#
+
+allow node_exporter_t self:fifo_file rw_fifo_file_perms;
+allow node_exporter_t self:process { getsched signal };
+allow node_exporter_t self:netlink_route_socket r_netlink_socket_perms;
+allow node_exporter_t self:tcp_socket create_stream_socket_perms;
+allow node_exporter_t self:udp_socket create_socket_perms;
+
+manage_files_pattern(node_exporter_t, node_exporter_runtime_t, node_exporter_runtime_t)
+files_runtime_filetrans(node_exporter_t, node_exporter_runtime_t, file)
+
+manage_dirs_pattern(node_exporter_t, node_exporter_var_lib_t, node_exporter_var_lib_t)
+manage_files_pattern(node_exporter_t, node_exporter_var_lib_t, node_exporter_var_lib_t)
+files_var_lib_filetrans(node_exporter_t, node_exporter_var_lib_t, { dir file })
+
+append_files_pattern(node_exporter_t, node_exporter_log_t, node_exporter_log_t)
+create_files_pattern(node_exporter_t, node_exporter_log_t, node_exporter_log_t)
+setattr_files_pattern(node_exporter_t, node_exporter_log_t, node_exporter_log_t)
+logging_log_filetrans(node_exporter_t, node_exporter_log_t, { dir file })
+
+# Also uses port 9100
+corenet_tcp_bind_hplip_port(node_exporter_t)
+corenet_tcp_bind_generic_node(node_exporter_t)
+
+dev_read_sysfs(node_exporter_t)
+
+fs_getattr_all_fs(node_exporter_t)
+
+init_read_state(node_exporter_t)
+
+kernel_read_fs_sysctls(node_exporter_t)
+kernel_read_kernel_sysctls(node_exporter_t)
+kernel_read_net_sysctls(node_exporter_t)
+kernel_read_network_state(node_exporter_t)
+kernel_read_software_raid_state(node_exporter_t)
+kernel_read_system_state(node_exporter_t)
+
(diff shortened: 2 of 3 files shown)


policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,65 @@
+policy_module(node_exporter, 1.0.0)
```
0xC0ncord, line 1: The version field is no longer used in refpolicy and can be removed.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,65 @@
...
+files_type(node_exporter_var_lib_t)
+
+type node_exporter_var_log_t;
+logging_log_file(node_exporter_var_log_t)
+
+allow node_exporter_t self:fifo_file rw_fifo_file_perms;
+allow node_exporter_t self:process { getsched signal };
+allow node_exporter_t self:netlink_route_socket r_netlink_socket_perms;
+allow node_exporter_t self:tcp_socket create_stream_socket_perms;
+allow node_exporter_t self:udp_socket create_stream_socket_perms;
```
0xC0ncord, line 20: I think this should just be `create_socket_perms`.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,65 @@
...
+
+corenet_tcp_bind_all_ports(node_exporter_t)
+corenet_tcp_bind_generic_node(node_exporter_t)
+
+dev_read_sysfs(node_exporter_t)
+
+fs_getattr_dos_fs(node_exporter_t)
+fs_getattr_rpc_pipefs(node_exporter_t)
+fs_getattr_tmpfs(node_exporter_t)
+fs_getattr_xattr_fs(node_exporter_t)
```
0xC0ncord, line 42: This is subjective but in my local policy I consolidated this down to `fs_getattr_all_fs`. Truthfully I do not remember why, but I suspect I did that in order to cover some edge cases like how I have a FAT32 ESP/boot partition that is labeled `boot_t` and to allow `node_exporter` to properly get metrics on such filesystems. Wondering what your thoughts are on doing the same here.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,65 @@
...
+
+init_read_state(node_exporter_t)
+
+# Search kernel sysctl rpc attributes
+kernel_read_rpc_sysctls(node_exporter_t)
```
0xC0ncord, line 47: Is this access required only when also using the RPC service? If so I think this should be moved into the below optional block so that this access isn't enabled when that service isn't in use.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,65 @@
...
+kernel_read_net_sysctls(node_exporter_t)
+kernel_read_network_state(node_exporter_t)
+kernel_read_software_raid_state(node_exporter_t)
+kernel_read_system_state(node_exporter_t)
+
+ifdef(`init_systemd',`
+	dbus_system_bus_client(node_exporter_t)
+
+	init_dbus_chat(node_exporter_t)
+	init_get_system_status(node_exporter_t)
```
0xC0ncord, line 60: I forgot to tell you earlier that we also need `init_get_all_units_status(node_exporter_t)` here.

policy/modules/services/node_exporter.te
jpds: Good catch, my EFI `/boot`'s are just `dosfs_t`.

policy/modules/services/node_exporter.fc
```diff
@@ -0,0 +1,6 @@
+/run/node_exporter\.pid	-- gen_context(system_u:object_r:node_exporter_runtime_t,s0)
+
+/usr/sbin/node_exporter	-- gen_context(system_u:object_r:node_exporter_exec_t,s0)
+
+/var/lib/node_exporter(/.*)?	gen_context(system_u:object_r:node_exporter_var_lib_t,s0)
+/var/log/node_exporter(/.*)?	gen_context(system_u:object_r:node_exporter_var_log_t,s0)
```
pebenito, line 6: Nit: whitespace change to make the gen_context() line up

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,62 @@
+policy_module(node_exporter)
+
```
pebenito, line 2: Needs a declarations heading.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,62 @@
...
+type node_exporter_exec_t;
+init_daemon_domain(node_exporter_t, node_exporter_exec_t)
+
+type node_exporter_runtime_t;
+files_runtime_file(node_exporter_runtime_t)
+
+type node_exporter_var_lib_t;
+files_type(node_exporter_var_lib_t)
+
+type node_exporter_var_log_t;
```
pebenito, line 13: `node_explorer_log_t` instead. We're trying to avoid encoding the path in the type name.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,62 @@
...
+logging_log_file(node_exporter_var_log_t)
+
+allow node_exporter_t self:fifo_file rw_fifo_file_perms;
```
pebenito, line 16: Needs a local policy heading.

policy/modules/services/node_exporter.te
```diff
@@ -0,0 +1,62 @@
...
+manage_dirs_pattern(node_exporter_t, node_exporter_var_lib_t, node_exporter_var_lib_t)
+manage_files_pattern(node_exporter_t, node_exporter_var_lib_t, node_exporter_var_lib_t)
+files_var_lib_filetrans(node_exporter_t, node_exporter_var_lib_t, { dir file })
+
+append_files_pattern(node_exporter_t, node_exporter_var_log_t, node_exporter_var_log_t)
+create_files_pattern(node_exporter_t, node_exporter_var_log_t, node_exporter_var_log_t)
+setattr_files_pattern(node_exporter_t, node_exporter_var_log_t, node_exporter_var_log_t)
+logging_log_filetrans(node_exporter_t, node_exporter_var_log_t, { dir file })
+
+corenet_tcp_bind_all_ports(node_exporter_t)
```
pebenito, line 34: This seems far too broad. Isn't there some reasonable subset of this?

pebenito (changes requested):

policy/modules/services/node_exporter.te
0xC0ncord: The default port for node_exporter is `9100/tcp` which is currently assigned to `hplip_t`. We can just replace this with `corenet_tcp_bind_hplip_port()` but it feels wrong. Otherwise it accomplishes the intended goal without giving too much access.

jpds: The `node_exporter` folks indeed decided to use a reserved port (and `hplip` also has more ports than would be usual). I picked the compromise of having it be able to listen on any port and then rely on the firewall to control access.

pebenito: I think it's OK to use hplip here, as long as there is a comment that says the node exporter also uses 9100. Hplip came first.

pebenito (approved):

PR #474, opened 2022-02-03
Add udica templates

0xC0ncord:
Add CIL templates for usage with [udica](https://github.com/containers/udica) that are compatible with refpolicy.

These templates are based on the ones provided by [container-selinux](https://github.com/containers/container-selinux/tree/main/udica-templates) but I cleaned them up a little ~~and opted to use macros for better readability~~. (cc @doverride for CIL comments).

I'm not super keen on the location for these, so if there are any suggestions on where they should be placed in the refpolicy repository, please comment.

### DIFF
--- a/udica-templates/base_container.cil
+++ b/udica-templates/base_container.cil
@@ -1,0 +1,77 @@
+;;
+;; Permission sets definitions
+;;
+
+(classpermission search_dir_perms)
+(classpermissionset search_dir_perms (dir (getattr search)))
+
+(classpermission list_dir_perms)
+(classpermissionset list_dir_perms (dir (getattr search open read lock ioctl)))
+
+(classpermission rw_dir_perms)
+(classpermissionset rw_dir_perms (dir (open read getattr lock search ioctl add_name remove_name write)))
+
+(classpermission manage_dir_perms)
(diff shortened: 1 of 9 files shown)


udica-templates/base_container.cil
```diff
@@ -0,0 +1,94 @@
+;;
+;; Macros
+;;
+
+(macro search_dirs ((type ARG1) (type ARG2))
+	(allow ARG1 ARG2 (dir (getattr search open)))
+)
```
doverride, line 7: search_dirs does not need "open" (i was reminded of that by @cgzones)

doverride: Also i would have used classpermissionset (template from the bottom up)

```
(classpermission·search_dir)$
(classpermissionset·search_dir·(dir·(getattr·search)))

(macro search_dirs ((type ARG1)(type ARG2))
            (allow ARG1 ARG2 search_dir))
```

0xC0ncord: Thanks. I opted to replace the macro calls with classpermissionsets instead as I think they are more clear.

Also removed the `open` permission from `search_dir_perms`.

pebenito:
Sorry, refpolicy does not support CIL modules.

pebenito:
I've reconsidered this. I will take these templates if you want to contribute them. Please open a new PR in that case.

PR #478, opened 2022-03-11
Further merge policies from containers-selinux

JAORMX:
This takes relevant file contexts into use from `containers-selinux`.
Thus enabling CRI-O support and taking some extra kubernetes paths that
are relevant for containerized deployments.

The `kubernetes_file_t` definition was taken into use as well and
containers are allowed to use it.

### DIFF
--- a/policy/modules/services/container.fc
+++ b/policy/modules/services/container.fc
@@ -18,18 +18,51 @@ HOME_DIR/\.local/share/docker/init(/.*)?		gen_context(system_u:object_r:containe
 HOME_DIR/\.local/share/docker/fuse-overlayfs(/.*)?		gen_context(system_u:object_r:container_ro_file_t,s0)
 HOME_DIR/\.local/share/docker/volumes(/.*)?		gen_context(system_u:object_r:container_file_t,s0)
 
-/usr/bin/crun	--	gen_context(system_u:object_r:container_engine_exec_t,s0)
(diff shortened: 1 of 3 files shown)


0xC0ncord:
Refpolicy's container module doesn't include specific types for different container engines (i.e. `docker` or `podman`). Instead, those applications have their own modules. The container module only sets up the base attributes and rules for containers and "generic" engines/runtimes.

There are a few file contexts you've added like `/var/lib/registry` and `/srv/containers` that I think are OK, but the rest should be added into their own modules to follow in the current direction.

Also, some of the types in refpolicy are named differently than in container-selinux, namely:
- `container_runtime_t` -> `container_engine_t`
- `container_runtime_exec_t` -> `container_engine_exec_t`
- `container_var_run_t` -> `container_runtime_t`

0xC0ncord:
Also, I think it would be best not to just copy stuff directly from `container-selinux` and instead write most of the container policies from scratch, using `container-selinux` as inspiration or a source of truth. `container-selinux` has been around for a while and is built on a lot of legacy cruft that hasn't been fully cleaned up. If you really delve into it you'll start to see parts of it that are quite a mess, or worse, are not and won't be compatible with refpolicy at all.

JAORMX:
Thanks for the clarification! I'll come back with a a PR with more targeted rules if needed then.

JAORMX:
Would be nice to get CRI-O for contexts here too. As there are containerd contexts

Issue #329, opened 2020-12-24
Confined user dbus instances cannot activate services


DemiMarie:
This is due to missing `service start` rules.

DemiMarie:
Here is the patch I used for my own policy:

```diff
diff --git a/qubes.fc b/qubes.fc
index ab251f1..adb479b 100644
--- a/qubes.fc
+++ b/qubes.fc
@@ -23,3 +23,14 @@
 `/rw/config/firewall\.nft'		gen_context(system_u:object_r:etc_t,s0)
 `/rw/config/tinyproxy'		-d	gen_context(system_u:object_r:etc_t,s0)
 `/rw/config/tinyproxy/[^/]+'	--	gen_context(system_u:object_r:tinyproxy_config_t,s0)
+`/run/user/[^/]+/systemd'	-d	gen_context(`system_u:object_r:init_var_run_t',`s0')
+`/run/user/[^/]+/systemd/generator(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/run/user/[^/]+/systemd/generator\.early(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/run/user/[^/]+/systemd/generator\.late(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/run/user/[^/]+/systemd/user(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/var/run/user/[^/]+/systemd'	-d	gen_context(`system_u:object_r:init_var_run_t',`s0')
+`/var/run/user/[^/]+/systemd/generator(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/var/run/user/[^/]+/systemd/generator\.early(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/var/run/user/[^/]+/systemd/generator\.late(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/var/run/user/[^/]+/systemd/user(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
+`/var/run/user/[^/]+/systemd/transient(/.*)?'	gen_context(`system_u:object_r:systemd_unit_file_t',`s0')
diff --git a/qubes.te b/qubes.te
index 785401e..835e3db 100644
--- a/qubes.te
[...]

PR #479, opened 2022-03-14
Add type for systemd runtime units and add dbus-broker support

0xC0ncord:
Add a separate type for systemd runtime units and add tunables to control access to using dbus-broker as the systemd dbus daemon.

The `dbus_broker_system_bus` tunable adds the rules necessary to allow dbus-broker to be used as the system dbus daemon, and `dbus_broker_run_runtime_units` controls whether dbus-broker is allowed to run systemd runtime units. The latter tunable was added to address dbus-broker's default behavior of creating and running arbitrary units itself (instead of asking systemd) if the requested service doesn't exist.

### DIFF
--- a/policy/modules/services/dbus.te
+++ b/policy/modules/services/dbus.te
@@ -55,6 +55,9 @@ init_mountpoint(system_dbusd_runtime_t)
 type system_dbusd_tmp_t;
 files_tmp_file(system_dbusd_tmp_t)
 
+type system_dbusd_tmpfs_t;
+files_tmpfs_file(system_dbusd_tmpfs_t)
+
 type system_dbusd_var_lib_t;
 files_type(system_dbusd_var_lib_t)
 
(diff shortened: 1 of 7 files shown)


pebenito (changes requested):
I don't think the tunables are needed. We should also be careful about naming the type, since this is adding transient units but naming the type "runtime". Runtime implies the units that go in `/run/systemd/system`.

0xC0ncord:
> I don't think the tunables are needed. We should also be careful about naming the type, since this is adding transient units but naming the type "runtime". Runtime implies the units that go in `/run/systemd/system`.

I added the `dbus_broker_run_runtime_units` tunable to address [this comment](https://github.com/SELinuxProject/refpolicy/issues/329#issuecomment-755418651) with regards to potentially allowing dbus-broker to transition to `initrc_t` in some circumstances. Truthfully I'm not sure how much of a problem this would be if we simply allowed the access.

As for the type naming, good point. Do you think `init_transient_unit_t` would suffice?

pebenito:
> potentially allowing dbus-broker to transition to `initrc_t`

You can run any daemon from `initrc_t` and do other transitions to privileged domains like `fsadm_t`. Now that I think about it, this is effectively arbitrary command execution with the privilege of init, so let's keep the tunables.

I'd prefer `systemd_transient_unit_t`.

pebenito (approved):

PR #480, opened 2022-03-14
Add support for static usermode helper config file

0xC0ncord:
Add a config file type for the userspace side of Linux's `CONFIG_STATIC_USERMODEHELPER` and allow kernel threads to read it. This optional config file can be referenced by a usermode binary used for `CONFIG_STATIC_USERMODEHELPER` that may contain options for various settings such as argument filtering. One such example of this is [huldufolk](https://github.com/tych0/huldufolk).

### DIFF
--- a/policy/modules/kernel/kernel.fc
+++ b/policy/modules/kernel/kernel.fc
@@ -1,2 +1,3 @@
+/etc/usermode-helper\.conf	--	gen_context(system_u:object_r:usermode_helper_conf_t,s0)
 /sys/kernel/debug	-d	gen_context(system_u:object_r:debugfs_t,s0)
 /sys/kernel/debug/.*		<<none>>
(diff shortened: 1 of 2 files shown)


pebenito:
Why not leave it `etc_t` and add the appropriate rule?

0xC0ncord:
I thought it would be a better idea to make this a private type since that means we can also ensure that no other domains with read access to `etc_t` can read this specific config file, even if that domain is running as root.

pebenito:
What data in the file is that sensitive?

0xC0ncord:
Using huldufolk as an example, the [config file](https://github.com/tych0/huldufolk/blob/master/sample-usermode-helper.toml) contains entries for things the kernel calls into userspace with `call_usermodehelper()`. While currently primitive at best, huldufolk allows the administrator to apply a sort of policy on the calls that can be made through the static usermode helper binary. At the moment that only applies to a static limit on `argc`, but the way forward would allow for setting capabilities, namespacing, argument filtering (possibly via regex), etc. Preventing processes other than kernel threads from reading this file is just closing any potential information leaks.

pebenito:
It sounds like a full policy for Huldufólk would be more appropriate.

PR #481, opened 2022-03-14
Various fixes, mostly systemd-related

0xC0ncord:
Various scattered fixes, mostly systemd-related.

Some of these are:
- Allow init to `nnp_transition` to any init daemon domain (systemd's `NoNewPrivileges` unit option)
- Allowing domains which can use nsswitch to read resolvconf symlinks (for systemd-resolved's stub configuration)
- Polyinstantiation fixes
- Various dontaudits

### DIFF
--- a/policy/modules/services/networkmanager.te
+++ b/policy/modules/services/networkmanager.te
@@ -160,6 +160,7 @@ storage_getattr_fixed_disk_dev(NetworkManager_t)
 init_read_utmp(NetworkManager_t)
 init_dontaudit_write_utmp(NetworkManager_t)
 init_domtrans_script(NetworkManager_t)
+init_get_system_status(NetworkManager_t)
 
 auth_use_nsswitch(NetworkManager_t)
 
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -654,6 +654,9 @@ udev_read_runtime_files(systemd_homework_t)
 
 allow systemd_hostnamed_t self:capability sys_admin;
 
+fs_getattr_cgroup(systemd_hostnamed_t)
+fs_getattr_xattr_fs(systemd_hostnamed_t)
+
 kernel_read_kernel_sysctls(systemd_hostnamed_t)
 kernel_dontaudit_getattr_proc(systemd_hostnamed_t)
 
@@ -992,6 +995,9 @@ optional_policy(`
 #
 # modules-load local policy
 #
+
+fs_getattr_cgroup(systemd_modules_load_t)
+fs_getattr_xattr_fs(systemd_modules_load_t)
 
 kernel_load_module(systemd_modules_load_t)
 kernel_read_kernel_sysctls(systemd_modules_load_t)
@@ -1345,6 +1351,10 @@ manage_dirs_pattern(systemd_rfkill_t, systemd_rfkill_var_lib_t, systemd_rfkill_v
 manage_files_pattern(systemd_rfkill_t, systemd_rfkill_var_lib_t, systemd_rfkill_var_lib_t)
 init_var_lib_filetrans(systemd_rfkill_t, systemd_rfkill_var_lib_t, dir)
 
+fs_getattr_cgroup(systemd_rfkill_t)
+fs_getattr_xattr_fs(systemd_rfkill_t)
+
+kernel_getattr_proc(systemd_rfkill_t)
 kernel_read_kernel_sysctls(systemd_rfkill_t)
 
 dev_read_sysfs(systemd_rfkill_t)
@@ -1559,6 +1569,7 @@ dev_setattr_all_sysfs(systemd_tmpfiles_t)
 dev_write_sysfs(systemd_tmpfiles_t)
 
 files_create_lock_dirs(systemd_tmpfiles_t)
+files_dontaudit_getattr_lost_found_dirs(systemd_tmpfiles_t)
 files_manage_all_runtime_dirs(systemd_tmpfiles_t)
 files_delete_usr_files(systemd_tmpfiles_t)
 files_list_home(systemd_tmpfiles_t)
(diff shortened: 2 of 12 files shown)


policy/modules/services/networkmanager.te
```diff
@@ -192,6 +192,10 @@ userdom_write_user_tmp_sockets(NetworkManager_t)
 userdom_dontaudit_use_unpriv_user_fds(NetworkManager_t)
 userdom_dontaudit_use_user_ttys(NetworkManager_t)
 
+ifdef(`init_systemd',`
+	init_get_system_status(NetworkManager_t)
```
pebenito, line 196: I don't think the ifdef is needed.

policy/modules/system/init.fc
```diff
@@ -7,6 +7,8 @@
 /etc/rc\.d/init\.d/.*	--	gen_context(system_u:object_r:initrc_exec_t,s0)
 /etc/sysconfig/network-scripts/ifup-ipsec -- gen_context(system_u:object_r:initrc_exec_t,s0)
 
+/etc/systemd/system(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
```
pebenito, line 10: This should already be handled by the `file_contexts.subs_dist`.

policy/modules/system/sysnetwork.fc
```diff
@@ -21,6 +21,7 @@ ifdef(`distro_debian',`
 /etc/hostname		--	gen_context(system_u:object_r:net_conf_t,s0)
 /etc/denyhosts.*	--	gen_context(system_u:object_r:net_conf_t,s0)
 /etc/resolv\.conf.*	--	gen_context(system_u:object_r:net_conf_t,s0)
+/etc/resolv\.conf.*	-l	gen_context(system_u:object_r:net_conf_t,s0)
```
pebenito, line 24: Normally we leave symlinks as the general label, `etc_t` in this case.

pebenito (changes requested):

policy/modules/system/init.fc
0xC0ncord: You are right. I think this is just a simple oversight.

pebenito (approved):

0xC0ncord:
Sorry for the push after the approval.

I fixed a small conflict with rules added to `systemd_sessions_t` since a rule was added allowing it to `getattr` on all filesystems, so I just removed the rules I added for specific ones.
Also fixed the merge conflict on `sudo.if`.

policy/modules/system/systemd.te
```diff
@@ -671,6 +674,7 @@ sysnet_manage_config(systemd_hostnamed_t)
 systemd_log_parse_environment(systemd_hostnamed_t)
 
 # Allow reading /run/udev/data/+dmi:id
+udev_search_runtime(systemd_hostnamed_t)
 udev_read_runtime_files(systemd_hostnamed_t)
```
pebenito, line 678: I guess I missed this before, but this indicates that `udev_read_runtime_files()` is missing perms, since it should already provide `udev_search_runtime()`.

policy/modules/system/systemd.te
```diff
@@ -1451,9 +1466,8 @@ allow systemd_sessions_t systemd_sessions_runtime_t:file manage_file_perms;
 files_runtime_filetrans(systemd_sessions_t, systemd_sessions_runtime_t, file)
 
 fs_getattr_all_fs(systemd_sessions_t)
+fs_getattr_cgroup(systemd_sessions_t)
 fs_search_cgroup_dirs(systemd_sessions_t)
-fs_search_tmpfs(systemd_sessions_t)
-fs_search_ramfs(systemd_sessions_t)
```
pebenito, line 1456: I don't understand the deletions. There's the getattr all, but no search all. I think you meant to remove `fs_getattr_cgroup(systemd_sessions_t)` from your patch?

pebenito (changes requested):

policy/modules/system/systemd.te
0xC0ncord: That appears to be the case. I will move this access into that interface.

policy/modules/system/systemd.te
0xC0ncord: Ah yes, sorry for the confusion. That is exactly what I wanted to do.

policy/modules/system/udev.if
```diff
@@ -503,6 +503,7 @@ interface(`udev_read_runtime_files',`
 	')
 
 	files_search_runtime($1)
+	udev_search_runtime($1)
 	read_files_pattern($1, udev_runtime_t, udev_runtime_t)
```
pebenito, line 507: Hmm, because of the `read_files_pattern()` there should already be the equivalent of `udev_search_runtime()` access in this interface.

0xC0ncord: Hmm. Unfortunately I didn't save the AVC for why I added this. I think for now I'll drop it and if I can reproduce it in the future I will look into it further.

policy/modules/system/systemd.te
```diff
@@ -1377,6 +1387,10 @@ init_runtime_filetrans(systemd_resolved_t, systemd_resolved_runtime_t, dir)
 
 dev_read_sysfs(systemd_resolved_t)
 
+fs_getattr_cgroup(systemd_resolved_t)
```
pebenito, line 1390: Please double check the fs getattr rules, as some of the domains have `fs_getattr_all()` but unfortunately the lines are out of order.

0xC0ncord: Good call. There were some duplicate rules added since the `fs_getattr_all()`s were added some time after I originally worked on this.

pebenito (approved):

0xC0ncord:
I think I pushed changes while Github was having internal issues which could explain why the checks aren't running.

PR #482, opened 2022-03-14
podman: add explicit range transition for conmon

0xC0ncord:
Ensure that when conmon is started, it runs in s0 and is able to
communicate with the container.

This is in support of https://github.com/containers/podman/pull/13093

### DIFF
--- a/policy/modules/services/podman.te
+++ b/policy/modules/services/podman.te
@@ -190,6 +190,14 @@ container_engine_tmp_filetrans(podman_conmon_t, { file sock_file })
 container_manage_engine_tmp_files(podman_conmon_t)
 container_manage_engine_tmp_sock_files(podman_conmon_t)
 
+# Ensure conmon runs in s0 so that it can talk to the container
+ifdef(`enable_mcs',`
+	range_transition podman_t podman_conmon_exec_t:process s0;
+')
+ifdef(`enable_mls',`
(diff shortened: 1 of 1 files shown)


policy/modules/services/podman.te
```diff
@@ -28,6 +28,16 @@ application_domain(podman_conmon_t, podman_conmon_exec_t)
...
 
+# Ensure conmon runs in s0 so that it can talk to the container
+ifdef(`enable_mcs',`
+	range_transition podman_t podman_conmon_exec_t:process s0;
+	range_transition podman_user_t podman_conmon_exec_t:process s0;
+')
+ifdef(`enable_mls',`
+	range_transition podman_t podman_conmon_exec_t:process s0;
+	range_transition podman_user_t podman_conmon_exec_t:process s0;
+')
```
pebenito, line 39: These should not be part of the declaration block. Please move them.

pebenito (changes requested):

policy/modules/services/podman.te
0xC0ncord: OK. I've moved these to the conmon block.

pebenito: The rules should go in the existing `podman_t` and `podman_user_t` blocks. The two enable_* blocks should go above the init_systemd block.

pebenito (approved):

PR #486, opened 2022-03-22
Remove unused role argument from userdom interfaces

dburgener:
Three userdom interfaces appear to have previously used role support
which has since been refactored out.  Rename this interfaces to remove
the role mention and drop the unused roll argument.


pebenito:
These were done intentionally from the beginning, so they'd mirror other interfaces meant to build up a role. If we drop them we'd need back compat interfaces.

dburgener:
Are you saying you're okay with dropping the _role suffix as long as we add back compat interfaces?  (And deprecation warnings?)

PR #491, opened 2022-04-01
container: minor additions

0xC0ncord:
Missed this in my last PR. If the tunables for containers to access public content are enabled, also allow adding a watch on public content directories.


0xC0ncord:
Also add rules to allow `podman logs -f` to function.

pebenito (approved):

PR #492, opened 2022-04-01
Replace "user runtime unit" wording with "user transient unit"

0xC0ncord:
Deprecate the current interfaces for systemd user transient units in favor of renaming them to be more consistent with the "transient" unit terminology in the `init` module. Alias `systemd_user_runtime_unit_t` to `systemd_user_transient_unit_t`.

Something to consider: systemd user runtime units *do* exist (`/run/systemd/user` and `/run/user/%{USERID}/user`), so if rules specific to these units were to be added later (they do not currently exist), the type should probably still be called `systemd_user_runtime_unit_t`. If we added this now we would be effectively changing the meaning of `systemd_user_runtime_unit_t` and I'm not sure what the best path for communicating this to consumers of refpolicy would be.

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -84,11 +84,11 @@ HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data
 /run/nologin	--	gen_context(system_u:object_r:systemd_sessions_runtime_t,s0)
 
 /run/user/%{USERID}/systemd	-d	gen_context(system_u:object_r:systemd_user_runtime_t,s0)
-/run/user/%{USERID}/systemd/generator(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/generator\.early(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/generator\.late(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/transient(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
+/run/user/%{USERID}/systemd/generator(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/generator\.early(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/generator\.late(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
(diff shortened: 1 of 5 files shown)


pebenito:
The interfaces that haven't been in a release don't need backwards compat.

0xC0ncord:
All of the interfaces that are deprecated in this PR have been released since [RELEASE_2_20210908](https://github.com/SELinuxProject/refpolicy/releases/tag/RELEASE_2_20210908).

policy/modules/system/systemd.fc
```diff
@@ -84,11 +84,11 @@ HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data
...
-/run/user/%{USERID}/systemd/generator(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/generator\.early(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/generator\.late(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/transient(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
-/run/user/%{USERID}/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_user_runtime_unit_t,s0)
+/run/user/%{USERID}/systemd/generator(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/generator\.early(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/generator\.late(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/transient(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
+/run/user/%{USERID}/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_user_transient_unit_t,s0)
```
pebenito, line 91: I've been debating this to myself for a while. Unlike systemd --system, I'm not sure that systemd --user gains much benefit from differentiating between transient and runtime units. One thing for certain is in the above lines, all but the transient dir is a runtime unit.

0xC0ncord: I agree I don't think there's much difference between the two as far as the policy is concerned, but I think that the fact that there is a difference is important enough to treat them differently in case anything changes in the future.

What I really want to do is redefine the user "runtime" unit that is already in the policy so that we have the appropriate types for both the runtime and transient units.

0xC0ncord: I had another look at the systemd docs about runtime vs transient units and I think we should do this instead: keep the current naming as-is, but add a new `systemd_user_transient_unit_t` type for `/run/user/%{USERID}/systemd/transient` and adjust the rules accordingly.

What do you think?

pebenito: Sounds good to me.

0xC0ncord:
I will open a new PR with the changes discussed when it is ready.

PR #494, opened 2022-04-21
seusers: Remove sddm.

### LINKED ISSUE #488: Cannot load modular policy without xserver module due to sddm user
6e5a6bffdb09992f962a6ecb09a0a90fe3e76269 adds the sddm user to the default users. However, if we are using a modular policy on a headless server and do not include the `xserver` module, we cannot compile the policy due to the `xdm_t` type not existing. Instead, we get an error like the following:
```
# semodule -i /usr/share/selinux/mcs/*.pp                                                                   
Failed to resolve selinuxuser statement at /var/lib/selinux/mcs/tmp/modules/400/base/cil:5618                         
Failed to resolve AST                                                                                                 
semodule:  Failed!   
```

Also, if the `sddm` user does not exist on the system we get this non-fatal warning:
```
libsemanage.add_user: user sddm not in password file
```

pebenito:
This breaks systems that do not have the xserver module.

This partially reverts 6e5a6bffdb09992f962a6ecb09a0a90fe3e76269.

Fixes #488

### DIFF
--- a/config/appconfig-mcs/seusers
+++ b/config/appconfig-mcs/seusers
@@ -1,3 +1,2 @@
 root:root:s0-mcs_systemhigh
 __default__:user_u:s0
-sddm:xdm:s0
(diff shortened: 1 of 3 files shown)


dburgener:
LGTM

PR #496, opened 2022-04-27
[WIP] Add policy for k3s

0xC0ncord:
Initial policy import for k3s. Add a new type for k3s, building upon the container module.

Still very much a work in progress in need of testing, but this PR is open for early commentary.


0xC0ncord:
In the future it may be a good idea to split out some of these rules into a separate kubernetes module and add appropriate templates. The templates would grant the necessary rules for various kubernetes distributions that other policy modules like this one would call.

PR #499, opened 2022-04-29
Add udica templates

0xC0ncord:
This is a reopen of #474 to add base container templates for [udica](https://github.com/containers/udica). Additional comments can be found in the previous PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,6 +133,9 @@ htmldir := $(LOCAL_ROOT)/doc/html
 doctmpdir := $(LOCAL_ROOT)/doc/tmp
 endif
 
+# udica templates path
+udicatemplates := udica-templates
+
 # config file paths
 globaltun := $(poldir)/global_tunables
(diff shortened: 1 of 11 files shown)


pebenito (changes requested):
Please add an install target in the `Makefile`.

0xC0ncord:
> Please add an install target in the `Makefile`.

Install target added as well as a test in the Github workflow.

Makefile
```diff
@@ -655,6 +668,7 @@ bare: clean
 	$(verbose) rm -f $(mod_conf)
 	$(verbose) rm -f $(booleans)
 	$(verbose) rm -fR $(htmldir)
+	$(verbose) rm -fR $(udicatemplatesdir)
```
pebenito, line 671: I don't think this should be added because `make bare` doesn't delete installed files.

0xC0ncord: Good point. I've removed this.

PR #489, opened 2022-04-01
Another round of various fixes

0xC0ncord:
Not many major things this time. Just a bunch of minor fixes.

PR #501, opened 2022-05-17
Another round of various fixes (reopen)

0xC0ncord:
This is a reopen of #489 that I inadvertently lost track of while moving branches around in my local working repo.

This PR is unchanged from #489 save for 4 extra commits on top.


yizhao1:
> > Hi, I have tested these patches and found that systemd-network-generator should also be set to systemd_networked_exec_t. Otherwise the directory /run/systemd/network/ still can not get the correct label.
> > [...]
> > But I got another AVC:
> 
> Thanks! I'll take a look at this and test it out. I'll see if I can work through that AVC as well if I can reproduce it. Is that the only AVC you got with these patches?

I only test systemd related patches and only found this so far.

0xC0ncord:
That access looks to be needed by udev as it searches the networkd runtime for `*.link` files which are apparently used as an alternative to custom udev rules. I've added the access in my test policy and will test it soon. Thanks for the finding!

policy/modules/services/ssh.if
```diff
@@ -174,6 +174,14 @@ template(`ssh_server_template', `
 		attribute ssh_server;
 		type sshd_exec_t, sshd_key_t;
 	')
+
+	## <desc>
+	## <p>
+	##	Allow sshd to use remote port forwarding (bind to any TCP port)
+	## </p>
+	## </desc>
+	gen_tunable($1_use_remote_port_forwarding, false)
```
pebenito, line 183: Let's go a bit simpler for the name; how about `$1_port_forwarding`?

policy/modules/system/init.if
```diff
@@ -1375,6 +1375,24 @@ interface(`init_read_var_lib_links',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`init_search_var_lib_dirs',`
+	gen_require(`
+		type init_var_lib_t;
+	')
+
+	allow $1 init_var_lib_t:dir search_dir_perms;
```
pebenito, line 1393: Needs parent dir access.

pebenito (changes requested):

policy/modules/services/ssh.if
0xC0ncord: Definitely much simpler. Let's use that.

0xC0ncord:
I've slightly tweaked the rules for the `/run/systemd/network` file transition after some more testing, thanks to @yizhao1's findings.

pebenito (approved):

PR #503, opened 2022-05-18
Do not transition to container engines for unconfined users

0xC0ncord:
Add an unconfined role template for unconfined users such that these users can use container resources but without transitioning to the container engine domains. Containers created by unconfined users will still transition as expected.

This PR also consolidates some of the required role access into their own interfaces for convenience, namely access for container files and container home content.

### DIFF
--- a/policy/modules/services/container.if
+++ b/policy/modules/services/container.if
@@ -130,7 +130,6 @@ interface(`container_user_engine',`
 #
 template(`container_base_role',`
 	gen_require(`
-		type container_file_t, container_ro_file_t;
 		type container_config_t;
 	')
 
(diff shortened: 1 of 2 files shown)


policy/modules/services/container.if
```diff
@@ -1079,6 +1091,119 @@ interface(`container_manage_home_data_sock_files',`
...
+## <summary>
+##	Administrate all container files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`container_admin_all_files',`
```
pebenito, line 1104: If the interface is `admin`, then it should be using `admin_pattern`.

pebenito (changes requested):

policy/modules/services/container.if
0xC0ncord: OK, I've reworked most of the `admin` interfaces to use `admin_pattern` but renamed the home content access interface to `manage` because it's more general than just typical admin access.

pebenito (approved):

PR #504, opened 2022-05-18
Rework conmon rules

0xC0ncord:
This is to lay some groundwork for an upcoming kubernetes policy module, specifically with CRI-O. CRI-O also uses conmon to monitor container processes, but currently this has the side effect of conmon transitioning back to `podman_t` when it executes `runc`, a generic container engine.

To fix this, rework the current conmon policy rules to use a template to generate conmon types. This is so that `podman_t`, `podman_user_t`, and eventually `crio_t` get their own conmon types that transition back to the proper source domain.

This has the side effect of renaming `podman_conmon_user_t` to `podman_user_conmon_t`, ~~but there hasn't been a release with the container module yet so this is OK,~~ and I like the new name better.

Also, add a file context for conmon in `/usr/libexec/conmon` which exists in Gentoo and consolidate the range transition of podman to conmon in an interface. This interface is also being used by the WIP CRI-O module.

### DIFF
--- a/policy/modules/services/podman.fc
+++ b/policy/modules/services/podman.fc
@@ -1,2 +1,4 @@
 /usr/bin/podman	--	gen_context(system_u:object_r:podman_exec_t,s0)
 /usr/bin/conmon	--	gen_context(system_u:object_r:conmon_exec_t,s0)
+
+/usr/libexec/podman/conmon	--	gen_context(system_u:object_r:conmon_exec_t,s0)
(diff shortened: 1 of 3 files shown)


pebenito:
> This has the side effect of renaming `podman_conmon_user_t` to `podman_user_conmon_t`, but there hasn't been a release with the container module yet so this is OK, and I like the new name better.

Sorry, I missed this before releasing, so you'll need compat. :(

0xC0ncord:
> Sorry, I missed this before releasing, so you'll need compat. :(

No worries. I've added the alias in the top commit.

PR #507, opened 2022-05-24
Various updates 20220524

pebenito:


policy/modules/services/container.te
```diff
@@ -414,12 +415,15 @@ corecmd_dontaudit_exec_all_executables(container_engine_domain)
...
 corenet_tcp_connect_http_port(container_engine_domain)
+corenet_tcp_connect_http_cache_port(container_engine_domain)
 corenet_tcp_bind_all_ports(container_engine_domain)
 corenet_udp_bind_all_ports(container_engine_domain)
 corenet_rw_tun_tap_dev(container_engine_domain)
 
 dev_getattr_all_blk_files(container_engine_domain)
 dev_getattr_all_chr_files(container_engine_domain)
+dev_getattr_generic_blk_files(container_engine_domain)
+dev_getattr_generic_chr_files(container_engine_domain)
```
0xC0ncord, line 426: My immediate reaction to this was "doesn't `dev_getattr_all_{blk,chr}_files()` already grant this access? Turns out it doesn't because `device_t` doesn't have the `device_node` attribute. I'm wondering if maybe one of these interface pairs should be renamed.

policy/modules/services/docker.te
```diff
@@ -45,6 +46,9 @@ container_runtime_named_socket_activation(dockerd_t)
 files_read_kernel_symbol_table(dockerd_t)
 files_dontaudit_write_usr_dirs(dockerd_t)
 
+init_get_system_status(dockerd_t)
+init_stop_generic_units(dockerd_t)
```
0xC0ncord, line 50: I think this go into an `ifdef('init_systemd')` block for consistency.

policy/modules/system/locallogin.te
```diff
@@ -145,6 +145,8 @@ ifdef(`init_systemd',`
 	auth_manage_faillog(local_login_t)
 
 	init_dbus_chat(local_login_t)
+        init_use_fds(local_login_t)
```
0xC0ncord, line 148: nit: spaces vs tabs.

policy/modules/services/container.te
0xC0ncord: Ah that makes sense. I saw similar behavior at one point with `docker` but couldn't reproduce it.

PR #509, opened 2022-06-01
container: Boolean for ecryptfs

jcpunk:
For moving secrets around in various environments, I'm finding ecryptfs to be a workable approach.  This adds a boolean that will permit my containers to access the filesystem if enabled.

### DIFF
--- a/policy/modules/kernel/filesystem.if
+++ b/policy/modules/kernel/filesystem.if
@@ -1931,6 +1931,84 @@ interface(`fs_cifs_domtrans',`
 	domain_auto_transition_pattern($1, cifs_t, $2)
 ')
 
+########################################
+## <summary>
+##      Create, read, write, and delete directories
+##      on an eCryptfs filesystem.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+## <rolecap/>
+#
+interface(`fs_manage_ecryptfs_dirs',`
+        gen_require(`
+                type ecryptfs_t;
+        ')
+
+        allow $1 ecryptfs_t:dir manage_dir_perms;
+')
+
+########################################
+## <summary>
+##      Create, read, write, and delete files
+##      on an eCryptfs filesystem.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+## <rolecap/>
+#
+interface(`fs_manage_ecryptfs_files',`
+        gen_require(`
+                type ecryptfs_t;
+        ')
+
+        manage_files_pattern($1, ecryptfs_t, ecryptfs_t)
+')
+
+########################################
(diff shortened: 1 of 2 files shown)


policy/modules/kernel/filesystem.if
```diff
@@ -1931,6 +1931,106 @@ interface(`fs_cifs_domtrans',`
...
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`fs_read_ecryptfs_symlinks',`
+        gen_require(`
+                type ecryptfs_t;
+        ')
+
+        allow $1 ecryptfs_t:dir list_dir_perms;
```
pebenito, line 2008: Please move this rule out of this interface into a new `fs_list_ecryptfs` interface.

policy/modules/kernel/filesystem.if
```diff
@@ -1931,6 +1931,106 @@ interface(`fs_cifs_domtrans',`
...
+##      </summary>
+## </param>
+## <rolecap/>
+#
+interface(`fs_exec_ecryptfs_files',`
+        gen_require(`
+                type ecryptfs_t;
+        ')
+
+        allow $1 ecryptfs_t:dir list_dir_perms;
```
pebenito, line 2030: Same idea as above, this rule shouldn't be here.

policy/modules/services/container.te
```diff
@@ -44,6 +44,14 @@ gen_tunable(container_use_nfs, false)
...
 gen_tunable(container_use_samba, false)
 
+## <desc>
+##      <p>
+##      Allow containers to use eCryptfs filesystems.
+##      </p>
+## </desc>
+gen_tunable(container_use_ecryptfs, false)
+
+
```
pebenito, line 54: Please move this declaration above `container_use_nfs` so the declarations are in order.

policy/modules/services/container.te
```diff
@@ -279,6 +287,14 @@ tunable_policy(`container_use_samba',`
...
 ')
 
+tunable_policy(`container_use_ecryptfs',`
+        fs_manage_cifs_dirs(container_domain)
+        fs_manage_cifs_files(container_domain)
+        fs_manage_cifs_named_sockets(container_domain)
+        fs_read_cifs_symlinks(container_domain)
+        fs_exec_cifs_files(container_domain)
+')
+
```
pebenito, line 297: Please move this above the `container_use_nfs` tunable block..

pebenito (changes requested):
Minor issues.

policy/modules/kernel/filesystem.if
pebenito: Please
1. Remove the rule from this interface and `fs_exec_ecryptfs_files`.
2. Create a new interface that consists of this rule, called `fs_list_ecryptfs`
3. Call `fs_list_ecryptfs` from container.te.

PR #512, opened 2022-06-20
Add additional MCS constraints.

pebenito:
- Collapse overlapping constraints with same expressions.
- Add additional SysV IPC constraints.
- Add additional socket constraints.


0xC0ncord:
Looks good to me. I was wondering about those socket constraints a while back but never got around to it.

PR #518, opened 2022-08-01
More various fixes, mostly ZFS and systemd-related

0xC0ncord:
Another collection of various fixes. Of major note:
- Add a new type for `/etc/fstab`, `fstab_t`, which is used for units generated from it by `systemd-fstab-generator`. This replaces `etc_t` being used as the type for these units.
- Various fixes for ZFS and its components. Make the `zfs` and `zpool` utilities run as `fsadm_t` and add missing accesses for it.
- Silence a few AVCs generated by `sudo` when trying to read the process state of its parent and child processes.
- Various fixes for NFS.
- Fixes for `udevadm` so that it can actually be used on the commandline.


policy/modules/system/init.te
```diff
@@ -23,12 +23,21 @@ gen_tunable(init_upstart, false)
...
+	## </desc>
+	gen_tunable(init_mounton_non_security, true)
+',`
+	## <desc>
+	## <p>
+	## Enable systemd to mount on all non-security files.
+	## </p>
+	## </desc>
+	gen_tunable(init_mounton_non_security, false)
+')
```
pebenito, line 40: No, we're not having this. Distros can use booleans.conf to override defaults.

policy/modules/system/init.te
```diff
@@ -463,6 +473,9 @@ ifdef(`init_systemd',`
 	init_manage_all_unit_files(init_t)
 	init_read_script_state(init_t)
 
+	# for running dracut-initramfs-restore
+	libs_exec_lib_files(init_t)
```
pebenito, line 477: Seems that the label should be changed to `bin_t` instead.

policy/modules/system/mount.te
```diff
@@ -182,6 +183,8 @@ optional_policy(`
 	corenet_dontaudit_udp_bind_all_reserved_ports(mount_t)
 	corenet_tcp_connect_all_ports(mount_t)
 
+	corecmd_exec_shell(mount_t)
```
pebenito, line 186: What requires this?

policy/modules/system/mount.te
```diff
@@ -54,6 +54,7 @@ manage_dirs_pattern(mount_t, mount_runtime_t, mount_runtime_t)
 manage_files_pattern(mount_t, mount_runtime_t, mount_runtime_t)
 rw_files_pattern(mount_t, mount_runtime_t, mount_runtime_t)
 files_runtime_filetrans(mount_t, mount_runtime_t, dir, "mount")
+files_runtime_filetrans(mount_t, mount_runtime_t, file)
```
pebenito, line 57: I think we can merge this into a single line. The `"mount"` name is unnecessary.

pebenito (changes requested):

policy/modules/kernel/files.fc
```diff
@@ -49,6 +49,7 @@ ifdef(`distro_suse',`
 /etc/\.fstab\.hal\..+	--	gen_context(system_u:object_r:etc_runtime_t,s0)
 /etc/blkid(/.*)?		gen_context(system_u:object_r:etc_runtime_t,s0)
 /etc/cmtab		--	gen_context(system_u:object_r:etc_runtime_t,s0)
+/etc/fstab		--	gen_context(system_u:object_r:fstab_t,s0)
```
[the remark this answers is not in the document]
0xC0ncord: For what it's worth I have been running this PR on top of 5 different machines in enforcing and haven't noticed any breakage after the initial work that went into it.

My initial motivation for introducing this change is to fix the confusion of `etc_t` systemd units. To me this was one of those things that has always given off a, "Hmm, that doesn't seem right" intuition. The added bonus of this is it allows restricting access to the `fstab` file itself, although I agree I don't think that alone has a lot of value.

If you insist I think I'd be OK with dropping this change and instead just editing the comments around `etc_t` systemd units to clarify these are generated by `systemd-fstab-generator` and will have the same type of `/etc/fstab`.

policy/modules/services/rpc.te
```diff
@@ -325,12 +326,14 @@ dev_rw_lvm_control(nfsd_t)
 
 files_getattr_tmp_dirs(nfsd_t)
 files_manage_mounttab(nfsd_t)
+files_search_mnt(nfsd_t)
```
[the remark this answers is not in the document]
0xC0ncord: I added the access for `mnt_t` since it was specific to the AVC I ran into on the system I developed this on, but I agree all mountpoints is probably better.

policy/modules/system/init.te
0xC0ncord: Good point. I'll drop this change.

policy/modules/kernel/files.fc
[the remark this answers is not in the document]
0xC0ncord: After some discussion on IRC about this I want to bring up this info to consider, although I think your point on the lack of much value is still valid here.

In theory a domain with access to edit `etc_t` files and to manage `etc_t` units could create arbitrary systemd units and essentially escalate to `init_t`. Systemd allows adding and starting arbitrary units by enabling a unit file via an absolute path:
```
# vi /etc/test.service
# systemctl enable /etc/test.service
Created symlink /etc/systemd/system/multi-user.target.wants/test.service → /etc/test.service.
# systemctl start test.service
```
You can also take advantage of `systemctl link` to add arbitrary paths to the unit search paths.

However, to my knowledge there aren't any domains in refpolicy currently with both access to edit `etc_t` files and to manage `etc_t` services (other than `sysadm_t` and `unconfined_t`), so as of right now this is effectively a non-issue.

It's also worth mentioning that editing `/etc/fstab` and controlling `etc_t` units also effectively grants access to control arbitrary mounts as long as `init_t` has the required permissions, but this relies on having the same access as mentioned above.

policy/modules/system/fstools.te
```diff
@@ -113,16 +113,23 @@ files_getattr_boot_dirs(fsadm_t)
...
 fs_rw_all_image_files(fsadm_t)
 fs_search_auto_mountpoints(fsadm_t)
 fs_getattr_xattr_fs(fsadm_t)
 fs_rw_ramfs_pipes(fsadm_t)
 fs_rw_tmpfs_files(fsadm_t)
+# for zfs mount/unmount commands
+fs_mount_xattr_fs(fsadm_t)
+fs_unmount_xattr_fs(fsadm_t)
+# for zfs filesystem rootcontext property
+fs_relabelfrom_xattr_fs(fsadm_t)
```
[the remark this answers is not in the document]
0xC0ncord: I think you're right. I wanted to avoid this as much as possible but it seems like the tools have enough needed accesses that creeps into the other fs-related domains that it's probably best to split it out.

I'll drop the changes specifically related to these tools from this series and work on that. It will probably end up as a new module.

policy/modules/system/mount.te
0xC0ncord: I can't remember exactly what I did to produce the AVC for this, so unless I can reproduce it I'll drop it.

policy/modules/kernel/files.fc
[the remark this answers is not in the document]
pebenito: In terms of the systemd units, I think it would probably be better to have a type for units in /etc. Then it's cleaner all around, from managing those units and controlling them. That separates it from fstab too.

0xC0ncord:
Dropping to draft status while pursuing work on making ZFS its own module.

policy/modules/kernel/files.fc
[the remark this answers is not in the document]
0xC0ncord: > I think it would probably be better to have a type for units in /etc.

I think that makes sense and covers all of my original concerns. I'll rework this.

policy/modules/system/mount.te
0xC0ncord: This was needed by ZFS when running zedlets in `/usr/libexec/zfs`, but the new ZFS policy module makes this access obsolete.

policy/modules/kernel/files.fc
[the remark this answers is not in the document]
0xC0ncord: >I think it would probably be better to have a type for units in /etc.

After thinking on it more, I'm not sure this is what we want. A unit file can exist anywhere on the system, and the corresponding service object just inherits the unit file's type. We could put a unit file in, say, `/mnt` and enable and start it, although the policy doesn't allow this if the unit file's type is wrong.

I think creating a type for generic units in `/etc` to include `fstab` seems like a step in the wrong direction. I think the only supported use case for arbitrary units to exist elsewhere on the system is if they are manually labeled `systemd_unit_t`. `fstab` is special in that it is the only file which is not inherently a unit file but which service objects are created from.

0xC0ncord:
Changes related to ZFS were dropped in favor of a dedicated ZFS policy module. Please go ahead with the review! :)

policy/modules/kernel/files.fc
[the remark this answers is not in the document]
0xC0ncord: @pebenito I left these original changes in this PR for now, but I am more than happy to split them into a separate PR/issue for additional discussion if it means getting these other changes reviewed and merged sooner.

policy/modules/kernel/corecommands.fc
```diff
@@ -177,6 +177,7 @@ ifdef(`distro_gentoo',`
 /usr/lib/dhcpcd/dhcpcd-hooks(/.*)?	gen_context(system_u:object_r:bin_t,s0)
 /usr/lib/dhcpcd/dhcpcd-run-hooks --	gen_context(system_u:object_r:bin_t,s0)
 /usr/lib/dovecot/.+			gen_context(system_u:object_r:bin_t,s0)
+/usr/lib/dracut(/.*)?		gen_context(system_u:object_r:bin_t,s0)
```
pebenito, line 180: Nit: the label is not aligned.

policy/modules/services/zfs.fc
```diff
@@ -0,0 +1,14 @@
...
+/usr/bin/zfs			--	gen_context(system_u:object_r:zfs_exec_t,s0)
+/usr/bin/zfs-auto-snapshot	--	gen_context(system_u:object_r:zfs_exec_t,s0)
+/usr/bin/zpool			--	gen_context(system_u:object_r:zfs_exec_t,s0)
+
+/usr/sbin/zed			--	gen_context(system_u:object_r:zed_exec_t,s0)
+/usr/sbin/zfs			--	gen_context(system_u:object_r:zfs_exec_t,s0)
+/usr/sbin/zpool			--	gen_context(system_u:object_r:zfs_exec_t,s0)
+
+/etc/zfs(/.*)?				gen_context(system_u:object_r:zfs_config_t,s0)
+/etc/zfs/zpool\.cache	--	gen_context(system_u:object_r:zfs_zpool_cache_t,s0)
```
pebenito, line 11: Nit: label not aligned.

policy/modules/services/zfs.te
```diff
@@ -0,0 +1,120 @@
...
+
+type zed_t;
+type zed_exec_t;
+init_daemon_domain(zed_t, zed_exec_t)
+role zfs_roles types zed_t;
+
+type zfs_t;
+type zfs_exec_t;
+application_domain(zfs_t, zfs_exec_t)
+init_daemon_domain(zfs_t, zfs_exec_t)
```
pebenito, line 18: I'm very unfamiliar with zfs; how is this used as a daemon?

policy/modules/services/zfs.te
```diff
@@ -0,0 +1,120 @@
...
+
+udev_search_runtime(zed_t)
+
+files_search_etc(zed_t)
+list_dirs_pattern(zed_t, zfs_config_t, zfs_config_t)
+read_files_pattern(zed_t, zfs_config_t, zfs_config_t)
+read_lnk_files_pattern(zed_t, zfs_config_t, zfs_config_t)
+
+files_runtime_filetrans(zed_t, zfs_runtime_t, file)
+manage_files_pattern(zed_t, zfs_runtime_t, zfs_runtime_t)
```
pebenito, line 67: Needs line order style cleanup.

policy/modules/services/zfs.te
```diff
@@ -0,0 +1,120 @@
...
+auth_use_nsswitch(zfs_t)
+
+mount_exec(zfs_t)
+
+userdom_use_user_terminals(zfs_t)
+
+files_search_etc(zfs_t)
+list_dirs_pattern(zfs_t, zfs_config_t, zfs_config_t)
+read_files_pattern(zfs_t, zfs_config_t, zfs_config_t)
+read_lnk_files_pattern(zfs_t, zfs_config_t, zfs_config_t)
```
pebenito, line 113: Need line order style cleanup.

policy/modules/kernel/files.te
```diff
@@ -75,6 +75,19 @@ optional_policy(`
 type etc_runtime_t, configfile;
 files_type(etc_runtime_t)
 
+#
+# fstab_t is the type of /etc/fstab
+#
+type fstab_t, configfile;
```
pebenito, line 81: I'm still not on board with this. I'd prefer to split this out to a separate PR.

pebenito (changes requested):

policy/modules/services/zfs.te
0xC0ncord: Ah, this should just be `init_system_domain()` instead. The `application_domain()` can be removed as well since that is also called by the former.

policy/modules/kernel/files.te
0xC0ncord: OK. I'll split it out into its own PR and also open an issue where I'll re-address my concerns.

0xC0ncord:
Changes related to making `/etc/fstab` a private type are dropped from this PR, pending inclusion in a separate one.

policy/modules/system/init.te
```diff
@@ -541,6 +542,11 @@ ifdef(`init_systemd',`
 		files_mounton_non_security(init_t)
 	')
 
+	optional_policy(`
+		# to run systemd-boot-update
+		bootloader_domtrans(init_t)
+	')
```
pebenito, line 548: I'm torn on having this vs. `init_system_domain()` in bootloader. I'm concerned about potential NNP issues.

0xC0ncord: The only thing that is needed is for systemd to start `systemd-boot-update.service` which just runs `bootctl --no-variables --graceful update`. `bootctl` itself should only be on the root filesystem so I'm not sure if NNP will be an issue. Am I missing something?

pebenito (approved):

PR #519, opened 2022-08-17
Add cloud-init.

pebenito:
This is used by cloud providers to set up VMs during deployment.

https://github.com/canonical/cloud-init

### DIFF
--- a/policy/modules/admin/cloudinit.if
+++ b/policy/modules/admin/cloudinit.if
@@ -1,0 +1,108 @@
+## <summary>Init scripts for cloud VMs</summary>
+
+########################################
+## <summary>
+##	Create cloud-init runtime directory.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cloudinit_create_runtime_dirs',`
+	gen_require(`
+		type cloud_init_runtime_t;
+	')
+
+	files_search_runtime($1)
+	allow $1 cloud_init_runtime_t:dir create_dir_perms;
+')
+
+########################################
+## <summary>
+##	Write cloud-init runtime files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cloudinit_write_runtime_files',`
+	gen_require(`
+		type cloud_init_runtime_t;
+	')
+
+	files_search_runtime($1)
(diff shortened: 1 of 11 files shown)


policy/modules/admin/cloudinit.if
```diff
@@ -0,0 +1,103 @@
...
+	files_search_runtime($1)
+	create_files_pattern($1, cloud_init_runtime_t, cloud_init_runtime_t)
+')
+
+#######################################
+## <summary>
+##	Create files in /run with the type used for
+##	cloud-init runtime files.
+## </summary>
+## <param name="domain">
```
dburgener, line 65: This interface takes 3 arguments but only 2 param fields are documented.

policy/modules/admin/cloudinit.te
```diff
@@ -0,0 +1,108 @@
...
+type cloud_init_state_t;
+files_type(cloud_init_state_t)
+
+########################################
+#
+# Local policy
+#
+
+allow cloud_init_t self:capability { chown dac_override dac_read_search fowner fsetid setgid setuid };
+dontaudit cloud_init_t self:capability { net_admin sys_tty_config };
```
dburgener, line 28: How confident are you that the sys_tty_config access is actually unneeded.  It seems to me like the sort of thing a cloud initialization program might want to initialize, but wouldn't be obvious under a lot of testing scenarios, and isn't *that* concerning security wise.

policy/modules/system/libraries.if
```diff
@@ -233,6 +233,31 @@ interface(`libs_dontaudit_write_lib_dirs',`
...
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`libs_dontaudit_manage_lib_dirs',`
+	gen_require(`
+		type lib_t;
+	')
+
+	dontaudit $1 lib_t:dir manage_dir_perms;
```
dburgener, line 258: I worry that this (and the files version) will cause pain by silencing more innocuous permissions like getattr and read.  I guess most domains have those anyways, so maybe it's not a big deal, but I wonder if silencing the actually disallowed permissions (write etc) would be more appropriate?

policy/modules/admin/cloudinit.te
pebenito: Pretty sure, though not 100%.

policy/modules/system/libraries.if
pebenito: Since `domain` has `list_dir_perms`, I do not have this concern.

PR #526, opened 2022-08-30
fapolicyd: Initial SELinux policy

dsugar100:
Initial policy for fapolicyd running on RHEL9.  Looked a bit at the RHEL9 policy, but here I have setup different domains for the fapolicyd daemon vs. the fapolicyd-cli vs. fagenpolicy (tool to regen the policy)

I'm sure there will be comments and if there are specific rules in question I'm happy to gather denials.  This was a bit of work and lots of denials while making progress.

### DIFF
--- a/policy/modules/admin/fapolicyd.fc
+++ b/policy/modules/admin/fapolicyd.fc
@@ -1,0 +1,15 @@
+/etc/fapolicyd(/.*)?				gen_context(system_u:object_r:fapolicyd_config_t,s0)
+/etc/fapolicyd/compiled\.rules	--	gen_context(system_u:object_r:fapolicyd_compiled_rules_t,mls_systemhigh)
+
+/usr/lib/systemd/system/[^/]*fapolicyd.*   --	gen_context(system_u:object_r:fapolicyd_unit_t,s0)
+
+/usr/sbin/fapolicyd				--	gen_context(system_u:object_r:fapolicyd_exec_t,s0)
+/usr/sbin/fapolicyd-cli			--	gen_context(system_u:object_r:fapolicyc_exec_t,s0)
+/usr/sbin/fagenrules			--	gen_context(system_u:object_r:fagenrules_exec_t,s0)
+
+/var/lib/fapolicyd(/.*)?			gen_context(system_u:object_r:fapolicyd_var_lib_t,s0)
+
+/var/log/fapolicyd-access\.log	--	gen_context(system_u:object_r:fapolicyd_log_t,s0)
+
+/var/run/fapolicyd(/.*)?			gen_context(system_u:object_r:fapolicyd_runtime_t,s0)
+/var/run/fapolicyd\.pid			--	gen_context(system_u:object_r:fapolicyd_runtime_t,s0)
--- a/policy/modules/admin/fapolicyd.if
+++ b/policy/modules/admin/fapolicyd.if
@@ -1,0 +1,157 @@
+## <summary>
+## The fapolicyd software framework controls the execution of applications based
+## on a user-defined policy. This is one of the most efficient ways to prevent
+## running untrusted and possibly malicious applications on the system.
+## </summary>
+
+########################################
+## <summary>
+##	Read fapolicyd config files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fapolicyd_mmap_read_config_files',`
+	gen_require(`
+		type fapolicyd_config_t;
+	')
+
+	mmap_read_files_pattern($1, fapolicyd_config_t, fapolicyd_config_t)
+')
+
+######################################
+## <summary>
+##	Execute fagenrules in the fagenrules_t domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`fapolicyd_domtrans_fagenrules',`
+	gen_require(`
+		type fagenrules_t, fagenrules_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, fagenrules_exec_t, fagenrules_t)
+')
+
+########################################
+## <summary>
+##	Execute fagenrules in the fagenrules domain,
+##	and allow the specified roles the
+##	fagenurles domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`fapolicyd_run_fagenrules',`
+	gen_require(`
+		attribute_role fagenrules_roles;
+	')
+
+	fapolicyd_domtrans_fagenrules($1)
+	roleattribute $2 fagenrules_roles;
+')
+
+
+#####################################
+## <summary>
+##	Execute fapolicyd-cli in the fapolicyc domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`fapolicyd_domtrans_cli',`
+	gen_require(`
+		type fapolicyc_t, fapolicyc_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, fapolicyc_exec_t, fapolicyc_t)
+')
+
+########################################
+## <summary>
+##	Execute fapoliyd-cli in the fapolicyc domain,
+##	and allow the specified roles the
+##	fapolicyc domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`fapolicyd_run_cli',`
+	gen_require(`
+		attribute_role fapolicyc_roles;
+	')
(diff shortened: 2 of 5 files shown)


policy/modules/admin/fapolicyd.te
```diff
@@ -0,0 +1,160 @@
...
+files_map_usr_files(fapolicyd_t)
+files_watch_all_mountpoints(fapolicyd_t)
+
+fs_getattr_xattr_fs(fapolicyd_t)
+
+kernel_getattr_proc(fapolicyd_t)
+kernel_read_kernel_sysctls(fapolicyd_t)
+kernel_read_crypto_sysctls(fapolicyd_t)
+
+logging_log_filetrans(fapolicyd_t, fapolicyd_log_t, file)
```
pebenito, line 81: Seems that `fapolicyd_log_t` is dead, as this is the only access from inside this module.

policy/modules/admin/fapolicyd.te
```diff
@@ -0,0 +1,160 @@
...
+allow fapolicyd_t self:process { setcap setsched };
+
+manage_fifo_files_pattern(fapolicyd_t, fapolicyd_runtime_t, fapolicyd_runtime_t)
+manage_files_pattern(fapolicyd_t, fapolicyd_runtime_t, fapolicyd_runtime_t)
+mmap_manage_files_pattern(fapolicyd_t,fapolicyd_var_lib_t,fapolicyd_var_lib_t)
+
+mmap_read_files_pattern(fapolicyd_t,fapolicyd_config_t,fapolicyd_config_t)
+files_runtime_filetrans(fapolicyd_t, fapolicyd_runtime_t, { file fifo_file })
+
+read_files_pattern(fapolicyd_t, fapolicyd_config_t, fapolicyd_compiled_rules_t)
```
pebenito, line 63: Needs cleanup on line ordering

policy/modules/admin/fapolicyd.te
```diff
@@ -0,0 +1,160 @@
...
+
+fapolicyd_read_config_files(fapolicyd_t)
+
+domain_read_all_entry_files(fapolicyd_t)
```
pebenito, line 67: Should be redundant to the `files_read_all_files()` below, no?

policy/modules/admin/fapolicyd.te
```diff
@@ -0,0 +1,160 @@
...
+corecmd_exec_shell(fagenrules_t)
+
+# fagenpolicy uses 'pidof' to find running fapolicyd
+domain_dontaudit_read_all_domains_state(fagenrules_t)
+
+files_list_runtime(fagenrules_t)
+files_read_etc_files(fagenrules_t)
+
+files_tmp_filetrans(fagenrules_t, fagenrules_tmp_t, { dir file })
+manage_files_pattern(fagenrules_t, fagenrules_tmp_t, fagenrules_tmp_t)
```
pebenito, line 122: Needs line ordering style cleanup.

policy/modules/kernel/files.if
```diff
@@ -1844,6 +1844,24 @@ interface(`files_dontaudit_list_all_mountpoints',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`files_watch_all_mountpoints',`
+	gen_require(`
+		attribute mountpoint;
+	')
+
+	allow $1 mountpoint:dir { watch_mount watch_with_perm };
```
pebenito, line 1862: Need two interfaces. Also "watch" in the interface name would refer to the dir:watch permission, so we need new verbs here.

policy/modules/roles/sysadm.te
```diff
@@ -398,6 +398,11 @@ optional_policy(`
 	fail2ban_admin(sysadm_t, sysadm_r)
 ')
 
+optional_policy(`
+	fapolicyd_admin(sysadm_t, sysadm_r)
+	fapolicyd_run_cli(sysadm_t, sysadm_r)
```
pebenito, line 403: Seems that the run_cli should also be in the admin interface.

pebenito (changes requested):

policy/modules/admin/fapolicyd.fc
```diff
@@ -0,0 +1,13 @@
+/etc/fapolicyd(/.*)?				gen_context(system_u:object_r:fapolicyd_config_t,s0)
+/etc/fapolicyd/compiled\.rules	--	gen_context(system_u:object_r:fapolicyd_compiled_rules_t,s0)
```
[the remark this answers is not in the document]
dsugar100: I setup a different type for compiled.rules because it is compiled by fagenpolicy (which writes as fapolicyd_conf_t but then calls restorecon at the end of the script).  This way the compiled rules are read-only to everyone.
Alternately fagenpoilcy could write to fapolicyd_compiled_rules_t and the restorecon becomes redundant.

policy/modules/admin/fapolicyd.if
```diff
@@ -0,0 +1,117 @@
...
+## <summary>
+##	Read fapolicyd config files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fapolicyd_read_config_files',`
```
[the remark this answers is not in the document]
dsugar100: Yes, it needs mmap read, I will update interface name.

policy/modules/kernel/files.if
dsugar100: I will setup two interfaces and try to make names that are reasonable, but I'm open to naming suggestions.

policy/modules/admin/fapolicyd.if
```diff
@@ -0,0 +1,117 @@
...
+## <summary>
+##	Execute fagenrules in the fagenrules_t domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`fapolicyd_fagenrules_domtrans',`
```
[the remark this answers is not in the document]
dsugar100: Fixed interface name

policy/modules/admin/fapolicyd.if
```diff
@@ -0,0 +1,117 @@
...
+##	administrate an fapolicyd environment.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
```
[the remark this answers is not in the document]
dsugar100: Actually, when including the 'fapolicyd_run_cli' in here (from suggestion below) the role parameter is no longer extra.

policy/modules/admin/fapolicyd.te
dsugar100: Sorry, I'm not sure what you are suggesting here.  There is a log file created.  Maybe the admin interface should grant read access to this type.  Maybe logrotate needs access (maybe the fact it is a log file grants logrotate necessary permissions?).

dsugar100: Or maybe you are suggesting that there is no write access granted, so something is missing.  Let me check into that.

dsugar100: Yup, it looks like it only generates the log file when 'do_stat_report' is enabled in the config file AND is exiting.  But, yes, I will all the rules to write to the log file. I see the denials for the missing rules.

policy/modules/admin/fapolicyd.fc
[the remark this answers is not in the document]
dsugar100: Looks to be too many issues with MLS (beyond fapolicy) will look at that in the future.

policy/modules/admin/fapolicyd.te
dsugar100: I think I got this correct, but please mention anything I missed.

policy/modules/admin/fapolicyd.te
dsugar100: I think I got this correct, again, please let me know if there is anything else.

policy/modules/admin/fapolicyd.if
```diff
@@ -0,0 +1,117 @@
...
+#
+interface(`fapolicyd_admin',`
+	gen_require(`
+		type fapolicyd_log_t, fapolicyd_config_t;
+	')
+	logging_search_logs($1)
+	admin_pattern($1, fapolicyd_log_t)
+
+	files_search_etc($1)
+	admin_pattern($1, fapolicyd_config_t)
```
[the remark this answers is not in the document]
dsugar100: Added

policy/modules/admin/fapolicyd.fc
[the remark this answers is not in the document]
pebenito: >Alternately fagenpoilcy could write to fapolicyd_compiled_rules_t and the restorecon becomes redundant.

If that's the only use of /tmp, that can work. If not, then we stick with what you've already written. Please add a comment in the policy explaining this.
 
> Though, I just noticed the comment before the restorecon:
> 
> ```
> # Restore context on MLS system.
> # /tmp is SystemLow & fapolicyd.rules is SystemHigh
> ```
> 
> So, maybe the context should be "gen_context(system_u:object_r:fapolicyd_compiled_rules_t,mls_systemhigh)"

It doesn't seem like this daemon would be usable on an MLS system. If the file is supposed to be world-readable, then that wouldn't work because it could expose paths/filenames that a process doesn't have access to due to its current level.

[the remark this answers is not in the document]
dsugar100: I wavered back and forth here and gave fagenrules_t write access to fapolicyd_compiled_rules_t as it is copying those files into the directory (with a type transition).  I think that is clearer than how it was (though I may have forgotten to edit comments - I will check that).

pebenito (approved):

PR #527, opened 2022-08-31
Allow systemd as `init_t` to make `root_t` and `syslogd_var_run_t` files

jcpunk:
When using `TemporaryFileSystem=/:ro` and then the suggested
`BindReadOnlyPaths=/dev/log /run/systemd/journal/socket /run/systemd/journal/stdout`
to generate an isolated runtime with logging, selinux needs to permit
`init_t` to create files under the `root_t` and `syslogd_var_run_t` type.
This allows systemd to spin up a mount namespace for service isolation.

Signed-off-by: Pat Riehecky <riehecky@fnal.gov>

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -304,6 +304,8 @@ ifdef(`init_systemd',`
 	domain_setpriority_all_domains(init_t)
 
 	allow init_t init_runtime_t:{ dir file } watch;
+	allow init_t root_t:file create;
+	allow init_t syslogd_var_run_t:file create;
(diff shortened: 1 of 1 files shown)


jcpunk:
I have no idea how to import those required types...

pebenito:
WE should probably handle all `init_mountpoint_type` rather than adding rules for individual mountpoints.

jcpunk:
I'll confess I've no idea how to do that....

pebenito:
Please try #535 . You will need to enable the `init_create_mountpoints` boolean and either add `init_mountpoint(syslogd_var_run_t)` or enable the `init_mounton_non_security` boolean.

PR #528, opened 2022-09-03
sysadm: move systemd_dbus_chat_hostnamed to init_systemd block

yizhao1:


pebenito:
I'd prefer to keep systemd lines in optional blocks if they can be expressed that way. A non-systemd systemd probably wouldn't have the systemd module, so the perms would correctly drop.

yizhao1:
> I'd prefer to keep systemd lines in optional blocks if they can be expressed that way. A non-systemd systemd probably wouldn't have the systemd module, so the perms would correctly drop.

ok, that's fine.

PR #533, opened 2022-09-06
selinuxutil: allow semanage to read fips state

dsugar100:
May  9 15:09:48 localhost audispd: node=localhost type=AVC msg=audit(1652108988.025:431): avc:  denied  { read } for  pid=1872 comm="semanage" name="fips_enabled" dev="proc" ino=14036 scontext=system_u:system_r:semanage_t:s0 tcontext=system_u:object_r:sysctl_crypto_t:s0 tclass=file permissive=1
May  9 15:09:48 localhost audispd: node=localhost type=AVC msg=audit(1652108988.025:431): avc:  denied  { open } for  pid=1872 comm="semanage" path="/proc/sys/crypto/fips_enabled" dev="proc" ino=14036 scontext=system_u:system_r:semanage_t:s0 tcontext=system_u:object_r:sysctl_crypto_t:s0 tclass=file permissive=1

### DIFF
--- a/policy/modules/system/selinuxutil.te
+++ b/policy/modules/system/selinuxutil.te
@@ -492,6 +492,7 @@ allow semanage_t semanage_tmp_t:dir manage_dir_perms;
 allow semanage_t semanage_tmp_t:file { manage_file_perms mmap_exec_file_perms };
 files_tmp_filetrans(semanage_t, semanage_tmp_t, { file dir })
 
+kernel_read_crypto_sysctls(semanage_t)
 kernel_read_system_state(semanage_t)
 kernel_read_kernel_sysctls(semanage_t)
 kernel_dontaudit_getattr_proc(semanage_t)


pebenito:
I've been looking for the common denominator(s) on all these fips_enabled checks. I can't immediately see why semanage would be using any crypto. What was happening?

dsugar100:
This case was probably seen when labeling ports for example "semanage port -a -t my_port_t -p udp 12345"
Looking at strace (attached)
It is loading hashlib.py which seems to be loading _hashlib.so (and other libraries).  At some point loading libssl which is then opening /proc/sys/crypto/fips_enabled.  This will only be on systems that have fips enabled (I think).  Before checking /proc it checks /etc/system-fips.  If that is not there I don't think it will check /proc.

This probably belongs in some common place because many things can use libssl.  
Maybe we should consolidate all the uses of 'kernel_read_crypto_sysctls' into the domain_type interface?  

I'm happy to consolidate this if you would like.  Let me know your thoughts.

[semanage_strace.txt](https://github.com/SELinuxProject/refpolicy/files/9539348/semanage_strace.txt)

pebenito:
> This probably belongs in some common place because many things can use libssl. Maybe we should consolidate all the uses of 'kernel_read_crypto_sysctls' into the domain_type interface?
> 
> I'm happy to consolidate this if you would like. Let me know your thoughts.

Let's consolidate into `domain.te`. This new info sank the notions I had of where the common denominator might be.

dsugar100:
Closing commit - different approach - see https://github.com/SELinuxProject/refpolicy/pull/536

PR #536, opened 2022-09-14
doamain: move kernel_read_crypto_sysctls to a common location

dsugar100:


policy/modules/kernel/domain.if
```diff
@@ -76,6 +76,8 @@ interface(`domain_type',`
 	# start with basic domain
 	domain_base_type($1)
 
+	kernel_read_crypto_sysctls($1)
```
pebenito, line 79: This should be in domain.te an work on the `domain` attribute directly.

pebenito (changes requested):

policy/modules/kernel/domain.if
dsugar100: Yes, I should have realized that initially.

pebenito (approved):

PR #539, opened 2022-09-17
CI: update and build SELint from source

cgzones:
* ci: update dependencies
  - Update checkout action to v3
  - Update python-setup action to v4

  Also print basic output from apt-get to debug potential flaky failures.

* ci: build SELint from source
  Disable newly added check C-008:

        usermanage.te:      581: (C): Identifier samba_domain_controller in expression for conditional block not found in own module, but in module samba (candidate for global declaration or interface) (C-008)
        mplayer.te:         122: (C): Identifier xserver_allow_dri in expression for conditional block not found in own module, but in module xserver (candidate for global declaration or interface) (C-008)
        nscd.te:            125: (C): Identifier samba_domain_controller in expression for conditional block not found in own module, but in module samba (candidate for global declaration or interface) (C-008)
        xguest.te:           44: (C): Identifier user_exec_noexattrfile in expression for conditional block not found in own module, but in module userdomain (candidate for global declaration or interface) (C-008)
        xguest.te:           48: (C): Identifier user_rw_noexattrfile in expression for conditional block not found in own module, but in module userdomain (candidate for global declaration or interface) (C-008)
        userdomain.if:     1278: (C): Identifier usbguard_user_modify_rule_files in expression for conditional block not found in own module, but in module usbguard (candidate for global declaration or interface) (C-008)
        Found the following issue counts:
        C-008: 6

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,12 +23,23 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update -q
+        sudo apt-get install -qy autoconf-archive bison flex libconfuse-dev uthash-dev
 
-        # Install SELint from Debian testing
-        wget -O - https://ftp-master.debian.org/keys/archive-key-10.asc 2>/dev/null | sudo apt-key add -
-        sudo add-apt-repository 'deb http://deb.debian.org/debian/ testing main' -y
-        sudo apt-get install -qy selint
-        selint -V
+    - name: Checkout SELint
+      uses: actions/checkout@v3
+      with:
+        repository: SELinuxProject/selint
+        # support exclusions in interface arguments
+        ref: '41a575e82dea5cd7f60b4fa7aeb84405dba3baba' # "Parse interface taking list of exemptions"
+        path: selint
+
+    - name: Build SELint
+      run: |
+        cd selint/
+        ./autogen.sh
+        ./configure --without-check
+        make -j$(nproc)
+        sudo make install
 
     - name: Create generated policy files
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -3,30 +3,43 @@ name: Build tests
...
-        wget -O - https://ftp-master.debian.org/keys/archive-key-10.asc 2>/dev/null | sudo apt-key add -
-        sudo add-apt-repository 'deb http://deb.debian.org/debian/ testing main' -y
-        sudo apt-get install -qqy selint
-        selint -V
+    - name: Checkout SELint
+      uses: actions/checkout@v3
+      with:
+        repository: SELinuxProject/selint
+        # support exclusions in interface arguments
+        ref: '41a575e82dea5cd7f60b4fa7aeb84405dba3baba' # "Parse interface taking list of exemptions"
```
dburgener, line 33: If you would prefer tying to a tag instead of a ref, v1.3.0 is now tagged and released.

.github/workflows/tests.yml
```diff
@@ -3,30 +3,43 @@ name: Build tests
 on: [push, pull_request]
 
 env:
-  SELINUX_USERSPACE_VERSION: checkpolicy-3.1
+  # 3.4 fails to validate, fixed with 88a703399f3f
+  # ("libsepol: fix validation of user declarations in modules")
+  SELINUX_USERSPACE_VERSION: checkpolicy-3.3
```
pebenito, line 8: Does selint require the new userspace? I want to keep this at refpolicy's minimum supported userspace version.

.github/workflows/tests.yml
pebenito: I'd prefer to use a tag.

pebenito (changes requested):

.github/workflows/tests.yml
```diff
@@ -3,30 +3,42 @@ name: Build tests
...
-        wget -O - https://ftp-master.debian.org/keys/archive-key-10.asc 2>/dev/null | sudo apt-key add -
-        sudo add-apt-repository 'deb http://deb.debian.org/debian/ testing main' -y
-        sudo apt-get install -qqy selint
-        selint -V
+    - name: Checkout SELint
+      uses: actions/checkout@v3
+      with:
+        repository: SELinuxProject/selint
+        # support exclusions in interface arguments
+        ref: 'v1.3.0' # "Parse interface taking list of exemptions"
```
dburgener, line 32: The comment referred to the old ref, and no longer matches the tag.  Given that the tag is somewhat more descriptive than a ref, the comment can probably be dropped.

PR #541, opened 2022-09-19
zfs: various fixes

0xC0ncord:
Fixes for ZFS, including allowing Zed to use sendmail and write LED statuses to enclosure devices and for the `zpool` utility to properly create and destroy ZFS pools.

Also add file contexts for nullmailer in `/var/spool` and allow mailer domains to trigger it.


0xC0ncord:
I am not set on the type for `/var/spool/nullmailer/trigger`. `mqueue_spool_t` may be more optimal.

policy/modules/services/zfs.te
```diff
@@ -112,9 +122,35 @@ mount_exec(zfs_t)
...
+
+	manage_dirs_pattern(zed_t, zed_mail_tmp_t, zed_mail_tmp_t)
+	manage_files_pattern(zed_t, zed_mail_tmp_t, zed_mail_tmp_t)
+	files_tmp_filetrans(zed_t, zed_mail_tmp_t, { dir file })
+
+	mta_sendmail_domtrans(zed_t, zed_mail_t)
+
+	allow zed_mail_t self:capability { dac_override dac_read_search };
+
+	storage_raw_rw_fixed_disk(zed_mail_t)
```
pebenito, line 155: This is very concerning, a networked process with raw disk access.

0xC0ncord: I'll see if there's a workaround or some other way we can avoid this.

pebenito: If not, please make a tunable. Then it will at least make the users aware of the access.

0xC0ncord: That was my next question. :)

0xC0ncord: I found that this access is not needed at all yet for some reason is attempted by the `mail` process that gets spawned. I ended up just `dontaudit`ing it.

pebenito: Perhaps a leaked fd?

PR #545, opened 2022-09-23
radius: fixes for freeradius

yizhao1:
* Add dac_read_search capability to radiusd_t
* Add getcap, ptrace and execmem to radiusd_t process

Fixes:
avc: denied { dac_read_search } for pid=473 comm="radiusd" capability=2 scontext=system_u:system_r:radiusd_t
tcontext=system_u:system_r:radiusd_t tclass=capability permissive=1

avc: denied { getcap } for pid=473 comm="radiusd"
scontext=system_u:system_r:radiusd_t
tcontext=system_u:system_r:radiusd_t tclass=process permissive=1

avc: denied { ptrace } for pid=474 comm="radiusd"
scontext=system_u:system_r:radiusd_t
tcontext=system_u:system_r:radiusd_t tclass=process permissive=1

avc: denied { execmem } for pid=475 comm="radiusd" scontext=system_u:system_r:radiusd_t
tcontext=system_u:system_r:radiusd_t tclass=process permissive=1

### DIFF
--- a/policy/modules/services/radius.te
+++ b/policy/modules/services/radius.te
@@ -32,9 +32,9 @@ files_type(radiusd_var_lib_t)
 # Local policy
 #
 
-allow radiusd_t self:capability { chown dac_override fsetid kill setgid setuid sys_resource sys_tty_config };
+allow radiusd_t self:capability { chown dac_override dac_read_search fsetid kill setgid setuid sys_resource sys_tty_config };
 dontaudit radiusd_t self:capability sys_tty_config;
-allow radiusd_t self:process { getsched setrlimit setsched sigkill signal };
+allow radiusd_t self:process { getcap getsched setrlimit setsched sigkill signal };
 allow radiusd_t self:fifo_file rw_fifo_file_perms;
 allow radiusd_t self:unix_stream_socket { accept listen };
 allow radiusd_t self:tcp_socket { accept listen };


policy/modules/services/radius.te
```diff
@@ -32,9 +32,9 @@ files_type(radiusd_var_lib_t)
 # Local policy
 #
 
-allow radiusd_t self:capability { chown dac_override fsetid kill setgid setuid sys_resource sys_tty_config };
+allow radiusd_t self:capability { chown dac_override dac_read_search fsetid kill setgid setuid sys_resource sys_tty_config };
 dontaudit radiusd_t self:capability sys_tty_config;
-allow radiusd_t self:process { getsched setrlimit setsched sigkill signal };
+allow radiusd_t self:process { getcap getsched setrlimit setsched sigkill signal ptrace execmem };
```
pebenito, line 37: The `execmem` and `ptrace` are a bit surprising. What fails without this access?

yizhao1: I have tested it again. Seems dropping them has no effect on the radiusd daemon.

policy/modules/services/radius.te
```diff
@@ -82,7 +82,7 @@ corenet_tcp_connect_snmp_port(radiusd_t)
 
 corenet_sendrecv_generic_server_packets(radiusd_t)
 corenet_udp_bind_generic_port(radiusd_t)
-corenet_dontaudit_udp_bind_all_ports(radiusd_t)
+corenet_udp_bind_all_unreserved_ports(radiusd_t)
```
pebenito, line 85: Can you clarify how radiusd is selecting which port to bind to? It seems that we would still want the dontaudit to cover cases where it chooses a reserved port.

yizhao1: The default ports of radius are 1812 and 1813 [1]. Add corenet_udp_bind_all_unreserved_ports to make sure it can bind these ports. We can keep the line corenet_dontaudit_udp_bind_all_ports.

[1] https://en.wikipedia.org/wiki/RADIUS

pebenito: In that case, why not have labeling to on those two ports and then only allow binding on them?

yizhao1: Thanks, This is because the port 18120 is enabled in configure file. It works when I bind  this port with semanage port.

PR #546, opened 2022-09-24
Add a policy for OpenSM

0xC0ncord:
Add a policy for [OpenSM](https://github.com/linux-rdma/opensm), a software implementation of an InfiniBand subnet manager.

### DIFF
--- a/policy/modules/services/opensm.if
+++ b/policy/modules/services/opensm.if
@@ -1,0 +1,86 @@
+## <summary>OpenSM is a software implementation of an InfiniBand subnet manager.</summary>
+
+########################################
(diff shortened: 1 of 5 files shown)


policy/modules/services/opensm.if
```diff
@@ -0,0 +1,86 @@
+## <summary>Policy for opensm</summary>
```
pebenito, line 1: Please put a description for opensm. It should be obvious this is a policy for opensm.

pebenito (changes requested):

pebenito (approved):

PR #547, opened 2022-09-24
Various fixes for GlusterFS

0xC0ncord:
Various fixes for GlusterFS. Of note:
- add a type for the Gluster daemon's management ports
- add a type for Gluster bricks
- add a tunable to allow Gluster to automatically manage fcontexts for bricks
- allow `mount` to call `mount.glusterfs` to mount Gluster volumes


policy/modules/services/glusterfs.if
```diff
@@ -1,5 +1,89 @@
...
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`glusterfs_run',`
```
pebenito, line 39: These interfaces need some renaming. From the names themselves, it's not clear when they provide access to the CLI program or the daemon. I think `glusterfs_[verb]` should be operating on glusterd, and `glusterfs_[verb]_cli` would operate on the CLI tool.

policy/modules/services/glusterfs.te
```diff
@@ -27,16 +37,20 @@ files_tmp_file(glusterd_tmp_t)
...
 #
 
-allow glusterd_t self:capability { chown dac_override dac_read_search fowner sys_admin sys_resource };
-allow glusterd_t self:process { setrlimit signal };
+allow glusterd_t self:capability { chown dac_override dac_read_search fowner ipc_lock sys_admin sys_resource };
+allow glusterd_t self:process { getsched setrlimit signal signull };
 allow glusterd_t self:fifo_file rw_fifo_file_perms;
-allow glusterd_t self:tcp_socket { accept listen };
-allow glusterd_t self:unix_stream_socket { accept listen };
+allow glusterd_t self:netlink_audit_socket create_socket_perms;
```
pebenito, line 51: Is this glusterd itself auditing, or is it from running `setfiles`?

policy/modules/services/glusterfs.te
```diff
@@ -58,16 +72,20 @@ manage_files_pattern(glusterd_t, glusterd_runtime_t, glusterd_runtime_t)
 manage_sock_files_pattern(glusterd_t, glusterd_runtime_t, glusterd_runtime_t)
 files_runtime_filetrans(glusterd_t, glusterd_runtime_t, { dir file sock_file })
 
+can_exec(glusterd_t, glusterd_var_lib_t)
```
pebenito, line 75: Is this unavoidable? Writing files it can also execute is an easy arbitrary code execution. Glusterd being on the network makes it a prime target.

policy/modules/system/mount.te
```diff
@@ -66,6 +66,8 @@ kernel_request_load_module(mount_t)
 
 # required for mount.smbfs
 corecmd_exec_bin(mount_t)
+# required for mount.glusterfs
+corecmd_exec_shell(mount_t)
```
pebenito, line 70: Please move this in with the `glusterfs_domtrans`.

pebenito (changes requested):

policy/modules/services/glusterfs.if
0xC0ncord: These interfaces are all for the daemon, but I agree a rename would help to clarify.

policy/modules/services/glusterfs.te
0xC0ncord: I think this is from `setfiles` and I just forgot to move it.

policy/modules/services/glusterfs.te
0xC0ncord: The execute access is needed for hooks that live in `/var/lib/glusterfs/hooks`. To my knowledge these don't get written to so these can just be made a separate type.

policy/modules/services/glusterfs.if
pebenito: I missed that. I'm not clear on why mount.glusterfs would need to run glusterd. To start glusterd if it's not running?

0xC0ncord: It seems like `mount.glusterfs` runs the Gluster CLI which in turn runs `glusterd` at some point. For what reason I'm not sure yet. Investigating...

0xC0ncord: So it seems `mount.glusterfs` will execute `glusterfs` which, on my system, is just a symlink to the gluster daemon binary. In my testing, `glusterfs` was started with these arguments:
```
/usr/sbin/glusterfs --process-name fuse --volfile-server=<server> --volfile-id=/<volume> <mount-path>
```

So yes, the domtrans is needed. Without it, we get:
```
/sbin/mount.glusterfs: line 406: /usr/sbin/glusterfs: Permission denied
Mounting glusterfs on /mnt/temp failed.
```

policy/modules/services/glusterfs.te
0xC0ncord: This access can actually be removed because of the `domtrans()` to `setfiles`.

pebenito (approved):

PR #549, opened 2022-09-29
Dhcpcd fixes

yizhao1:


policy/modules/system/systemd.te
```diff
@@ -1461,6 +1462,7 @@ optional_policy(`
 	dbus_system_bus_client(systemd_resolved_t)
 	dbus_watch_system_bus_runtime_dirs(systemd_resolved_t)
 	dbus_watch_system_bus_runtime_named_sockets(systemd_resolved_t)
+	sysnet_dbus_chat_dhcpc(systemd_resolved_t)
```
pebenito, line 1465: Dhcpc does not appear to have the permissions to be a dbus service. This should probably be a call to `sysnet_dns_name_resolve` in dhcpc instead. That includes the dbus communication to systemd-resolved.

pebenito (changes requested):

policy/modules/system/sysnetwork.te
```diff
@@ -61,11 +61,11 @@ ifdef(`distro_debian',`
 #
 # DHCP client local policy
 #
-allow dhcpc_t self:capability { dac_override fsetid net_admin net_bind_service net_raw setpcap sys_nice sys_resource sys_tty_config };
+allow dhcpc_t self:capability { dac_override fsetid net_admin net_bind_service net_raw setgid setpcap setuid sys_chroot sys_nice sys_resource sys_tty_config };
```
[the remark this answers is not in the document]
yizhao1: These are introduced by dhcpcd privilege separation functionality. It is enabled by default in dhcpcd (https://github.com/NetworkConfiguration/dhcpcd/commit/8a7aa9822d71d4e2755cff3c151ec4f9df3bb83e). Without these capabilities, we will get the following errors when dhcpcd startup:
dhcpcd[410]: ps_dropprivs: chroot: /var/lib/dhcpcd: Operation not permitted
dhcpcd[410]: failed to drop privileges: Operation not permitted
dhcpcd[264]: setrlimit RLIMIT_NOFILE: Permission denied
dhcpcd[264]: setrlimit RLIMIT_NPROC: Permission denied

policy/modules/system/systemd.te
yizhao1: Thanks. It works when using sysnet_dns_name_resolve instead of sysnet_dbus_chat_dhcpc

policy/modules/system/sysnetwork.te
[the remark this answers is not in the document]
pebenito: Thanks for investigating.

policy/modules/system/systemd.te
pebenito: > Thanks. It works when using sysnet_dns_name_resolve instead of sysnet_dbus_chat_dhcpc

I'm confused. On your most recent push you restored the `sysnet_dbus_chat_dhcpc` was this an accident?

pebenito:
I'm confused. On your most recent push you restored the `sysnet_dbus_chat_dhcpc` was this an accident?

yizhao1:
> 

I push a new PR. Can you double check it ? I'm not using sysnet_dbus_chat_dhcpc but systemd_dbus_chat_resolved. Because I found that using systemd_dbus_chat_resolved is enough for this scenario.

PR #551, opened 2022-10-04
Fapolicyd fixes

dsugar100:


policy/modules/admin/fapolicyd.if
```diff
@@ -152,6 +152,8 @@ interface(`fapolicyd_admin',`
 	files_search_runtime($1)
 	admin_pattern($1, fapolicyd_runtime_t)
 
-	fapolicyd_run_fagenrules($1, $2)
+	ifndef(`direct_sysadm_daemon',`
+		fapolicyd_run_fagenrules($1, $2)
+	')
```
pebenito, line 157: Please move this to the end of the interface.

pebenito: Also, wouldn't this prevent the admin from generating rules on DIRECT_INITRC systems?

pebenito (changes requested):
Please add a CI target that will test this build combination so we can check for this (and similar) problems.

policy/modules/admin/fapolicyd.if
dsugar100: I'm not sure.  Is there a better way to solve this issue?  
in /lib/systemd/system/fapolicyd.service has an ExecStartPre that runs /usr/sbin/fagenrules - which is why there is init_script_domain(fagenrules_t, fagenrules_exec_t)
 in fapolicyd.te (and thus the conflict).

pebenito: Ok. I missed that.

pebenito:
> I hope you wanted it in this same pull request.

Yes, thanks.

pebenito (approved):

PR #552, opened 2022-10-09
A few minor fixes

0xC0ncord:
A small batch of minor fixes. Notably:
- add file contexts for systemd units in `/lib/dracut/modules.d` so that the resulting services are `systemd_unit_t` instead of `bin_t`
- add file contexts for `git`'s subcommands. This fixes conflicting type specifications between `/usr/bin/git` and its many hardlinks.
- add a couple file contexts for some programs in `/usr/bin` or `/usr/sbin` as they appear in Gentoo
- minor fixes for `rpc` in support of NFS
- allow `systemd-generator` to perform DNS resolution in order to resolve hostnames in `/etc/fstab` for NFS mounts

A couple of these commits are split out from #511 since they are not strictly related.

### DIFF
--- a/policy/modules/system/mount.te
+++ b/policy/modules/system/mount.te
@@ -42,6 +42,8 @@ application_domain(unconfined_mount_t, mount_exec_t)
 
 # setuid/setgid needed to mount cifs
 allow mount_t self:capability { chown dac_override dac_read_search ipc_lock setgid setuid sys_admin sys_rawio sys_tty_config };
+dontaudit mount_t self:capability { kill net_admin };
+dontaudit mount_t self:process setrlimit;
 
 allow mount_t mount_tmp_t:file manage_file_perms;
 allow mount_t mount_tmp_t:dir manage_dir_perms;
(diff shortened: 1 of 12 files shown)


policy/modules/roles/sysadm.te
```diff
@@ -41,6 +41,9 @@ corenet_ib_manage_subnet_unlabeled_endports(sysadm_t)
 dev_read_kmsg(sysadm_t)
 dev_rw_ipmi_dev(sysadm_t)
 
+# Use nfs tools
+fs_rw_nfsd_fs(sysadm_t)
```
pebenito, line 45: Seems like this should be in `rpc_admin()` instead.

pebenito (changes requested):

policy/modules/roles/sysadm.te
0xC0ncord: I agree that makes more sense.

policy/modules/system/mount.te
```diff
@@ -141,6 +145,8 @@ selinux_getattr_fs(mount_t)
 
 userdom_use_all_users_fds(mount_t)
 
+dbus_dontaudit_write_system_bus_runtime_named_sockets(mount_t)
```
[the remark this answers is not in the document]
0xC0ncord: I added a note for this in the commit where I made this change, but I'll summarize it here:

`mount.nfs` will attempt to start `rpc-statd` if it's not already running. It first tries to start it by using the systemd unit (dbus chat with init and writing to its runtime socket) and will fall back to executing it directly if this fails. I didn't want to allow `mount` to start other units, so these dontaudits are to ignore it.

pebenito (approved):

PR #553, opened 2022-10-09
Add a private type for /etc/fstab

0xC0ncord:
This was split out from #518 for further discussion (please see previous comments [here](https://github.com/SELinuxProject/refpolicy/pull/518#discussion_r935539471)). This PR adds a private type for `/etc/fstab` and reworks the existing interfaces to manage `etc_t` systemd services to manage `fstab_t` services instead.

The primary motivation for this is to make the definition of "`etc_t`" units more clear and more strict. These units are generated by `systemd-fstab-generator` when creating mount units at runtime from it. The services created will inherit the type of `/etc/fstab` which is currently `etc_t`, so starting/stopping/etc these services will be checked against the `etc_t` type.

Making `/etc/fstab` a private type offers additional granularity to the systemd units created from it. Since the units created are mount units, a domain with both write access to `etc_t` files and access to start `etc_t` services can control arbitrary mounts. Currently this is a non-issue though, as the only domains with this access are the sysadm and unconfined users. However, this PR would allow gating this access in a more strict way. A `neverallow` rule could be added, for example.

It's worth noting that systemd does allow starting units from locations that are outside its normal search path, simply by specifying the absolute path to the unit instead of just its name:
```
# vi /etc/test.service
# systemctl enable /etc/test.service
Created symlink /etc/systemd/system/multi-user.target.wants/test.service → /etc/test.service.
# systemctl start test.service
```
The service created above will also be `etc_t`. While units outside of the search path are (probably) an unsupported use case, a private type for `/etc/fstab` would further isolate its units from any that could be created in this way. It's worth mentioning the ability to create arbitrary units with other types is not specific to `etc_t`. A service started from a unit in `/usr/bin` will be `bin_t`, for example. However, rules to manage these units don't exist and so interacting with their services will never be allowed. `/etc/fstab` is special in that it is the only file that is normally just a file, but has services created from it which will inherit its type.


pebenito:
How much testing have you done on this? There must be other domains that have access to `etc_t` for the purpose of accessing fstab. What immediately comes to mind is the mount command.

0xC0ncord:
> How much testing have you done on this? There must be other domains that have access to `etc_t` for the purpose of accessing fstab. What immediately comes to mind is the mount command.

I was running this patch locally on one system for about 2 or 3 months prior to submitting it as part of #518. Amazingly `mount` is one of the things that I did not test against this. Aside from that, I haven't noticed any other breakage but I agree there are likely some other domains that need read access to `fstab` that haven't been accounted for here.

pebenito:
Thanks for looking into this. I think this is a good change, but I'm concerned about the potential for unexpected breakage. I'd rather integrate this change into refpolicy3.

pebenito:
After it has been in refpolicy3 for some time (e.g. during alpha/beta periods) then I'd be happy to revisit this PR.

0xC0ncord:
> After it has been in refpolicy3 for some time (e.g. during alpha/beta periods) then I'd be happy to revisit this PR.

Sounds good. Though in any case, on systemd 252 it seems that the mount units generated from `/etc/fstab` are now `init_runtime_t`, so I think most of my qualms about the separation are "fixed" by just making sure `sysadm` can manage these units. PR with this included soon to follow.

PR #562, opened 2022-12-01
fstools: handle gentoo place for drivedb.h

montjoie:
On a gentoo-hardened+selinux, I got denial from fsadm_t reading var_t. This is due to smartctl trying to read /var/db/smartmontools/drivedb.h

### DIFF
--- a/policy/modules/system/fstools.fc
+++ b/policy/modules/system/fstools.fc
@@ -108,6 +108,10 @@
 /usr/sbin/zstreamdump		--	gen_context(system_u:object_r:fsadm_exec_t,s0)
 /usr/sbin/ztest			--	gen_context(system_u:object_r:fsadm_exec_t,s0)
 
+ifdef(`distro_gentoo',`
+/var/db/smartmontools(/.*)?		gen_context(system_u:object_r:fsadm_db_t,s0)
+')
+
 /var/swap			--	gen_context(system_u:object_r:swapfile_t,s0)
 
 /var/log/fsck(/.*)?		gen_context(system_u:object_r:fsadm_log_t,s0)
--- a/policy/modules/system/fstools.te
+++ b/policy/modules/system/fstools.te
@@ -19,6 +19,11 @@ files_tmp_file(fsadm_tmp_t)
 type fsadm_run_t;
 files_runtime_file(fsadm_run_t)
 
+ifdef(`distro_gentoo',`
+type fsadm_db_t;
+files_type(fsadm_db_t)
+')
+
 type swapfile_t; # customizable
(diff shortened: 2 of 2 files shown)


pebenito:
I'd prefer that there are no `var_t` files.

@perfinion

0xC0ncord:
I ran into this missing access before too. `fsadm_t` doesn't normally have read access to anything I think we could label this file as, so a private type may be necessary. The information in this file isn't sensitive.

pebenito:
> I ran into this missing access before too. `fsadm_t` doesn't normally have read access to anything I think we could label this file as, so a private type may be necessary. The information in this file isn't sensitive.

Are you saying that smartmontools wants to install this file to /var/db/smartmontools? It seems like a weird place for a (I presume) static file.

0xC0ncord:
> Are you saying that smartmontools wants to install this file to /var/db/smartmontools? It seems like a weird place for a (I presume) static file.

On Gentoo at least, this file is updated at runtime with `update-smart-drivedb` or by manually copying `/usr/share/smartmontools/drivedb.h` to `/var/db/smartmontools`. Now that I think of it, we could just label this file as `usr_t`.

0xC0ncord:
As far as I'm aware, this file isn't touched by the package itself because the file supports user modifications, so it is up to the user to update it.

montjoie:
So do you agree to label the directory as fsadm_var_t ?

pebenito:
Hmm, on Fedora, the RPM provides `/usr/share/smartmontools/drivedb.h`, and then `update-smart-drivedb` updates it:

```bash
$ sudo update-smart-drivedb
/usr/share/smartmontools/drivedb.h 7.3/5319 updated to 7.3/5417
```

Does Gentoo have a systemd timer/cron to update this routinely? If it's only a manual thing done by an admin, I'd be ok with it being `usr_t`.

0xC0ncord:
> Does Gentoo have a systemd timer/cron to update this routinely?

There is in fact a cronjob installed by Gentoo's `smartmontools` package which runs monthly by default and it executes `/usr/sbin/update-smart-drivedb -q`. The `-q` argument just suppresses info output.

montjoie:
I have updated to no use of var_t, and so create a dedicated fsadm_var_lib_t

pebenito:
Please put in a `distro_gentoo` block.

0xC0ncord:
@montjoie friendly ping. I'd like to be able to carry this into the next Gentoo SELinux policy release if you are able to get this PR across the finish line. ;)

pebenito (approved):

PR #566, opened 2022-12-07
Some more various fixes

0xC0ncord:
More minor fixes. Of particular note:
- Add missing rules for `hddtemp` to be used as an interactive application by `sysadm`
- Minor fixes for `netutils` to allow `nmap` and `traceroute` to work
- Minor fixes for `container` to allow [MetalLB's](https://metallb.org/) BGP speakers to run
- Allow `systemd-creds` to `setattr` on directories on RAM filesystems (or else many services fail to start on systemd 252)
- Allow domains sending syslog messages to connect to kernel unix stream sockets. This seems to be because of `journald` starting at boot before the policy is loaded. AVCs around this access seem to be way more prominent on systemd 252
- Add a filecon for SASL keytab files and allow `postfix_smtpd_t` to read them
- Various fixes for `libvirtd` and `systemd-machines`

Also, it seems that as of systemd 252, mount units generated from `/etc/fstab` are now properly labeled `init_runtime_t` as well as their service objects. This PR adds the needed access for `sysadm` to manage these. I've left the old `etc_t:service` access for backwards compatibility, but it should be removed in the future.


policy/modules/services/postfix.te
```diff
@@ -208,6 +208,7 @@ allow postfix_master_t postfix_etc_t:file rw_file_perms;
 
 allow postfix_master_t postfix_data_t:dir manage_dir_perms;
 allow postfix_master_t postfix_data_t:file manage_file_perms;
+allow postfix_master_t postfix_data_t:file map;
```
pebenito, line 211: nit: Please merge lines. Making a `mmap_manage_file_perms` would be good too.

policy/modules/system/init.te
```diff
@@ -618,6 +620,10 @@ ifdef(`init_systemd',`
 		unconfined_create_keys(init_t)
 		unconfined_write_keys(init_t)
 	')
+
+	optional_policy(`
+		virt_stream_connect(init_t)
```
pebenito, line 625: This is related to machined?

pebenito (changes requested):

policy/modules/system/init.te
pebenito: From what I can tell, it's related to `virtlogd-admin.socket`, so virtlogd needs a socket activation call instead.

pebenito:
Looks like the `mmap_manage_file_perms` can be used elsewhere:
```
aptcacher.te:        54: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
matrixd.te:          54: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
nsd.te:              47: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
postfix.te:         511: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
mailman.te:         112: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { map manage_file_perms }, would add (none)) (S-010)
mailman.te:         126: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { map manage_file_perms }, would add (none)) (S-010)
mailman.te:         286: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { map manage_file_perms }, would add (none)) (S-010)
mailman.te:         296: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { map manage_file_perms }, would add (none)) (S-010)
mozilla.te:          89: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
pulseaudio.te:       61: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
alsa.te:             71: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
pulseaudio.if:       48: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
apt.if:             194: (S): Suggesting permission macro: mmap_manage_file_perms (replacing { manage_file_perms map }, would add (none)) (S-010)
```

0xC0ncord:
Added another commit to replace those instances with the new perm set. :)

pebenito:
Thanks for the extra change

pebenito (approved):

PR #572, opened 2022-12-15
Fix templates parsing in gentemplates.sh

miroshko:
Template definitions might have a whitespace after the comma, e.g. su_restricted_domain_template
in /policy/modules/admin/su.if

template(\`su_restricted_domain_template', `
  ...
')

gentemplates.sh silently fails to parse it. This works unless 'set -e' is set, in which case the script fails non-silently.

This commit adds support of whitespace after comma, which is a valid syntax.


pebenito:
Thanks for the patch, but you need to add a Signed-off-by to your commit message.

miroshko:
thx @pebenito I added it to the commit message

PR #573, opened 2022-12-16
rsyslog: add label for /var/empty/dev/log

montjoie:
On gentoo, starting rsyslog give this:
allow syslogd_t var_t:dir { add_name remove_name }; allow syslogd_t var_t:sock_file { create setattr unlink };

This is due to the following piece of code in configuration: """
 Create an additional socket for the default chroot location
 (used by net-misc/openssh[hpn], see https://bugs.gentoo.org/490744)
 input(type="imuxsock" Socket="/var/empty/dev/log")
"""

So let's add correct label for this file

### DIFF
--- a/policy/modules/kernel/devices.fc
+++ b/policy/modules/kernel/devices.fc
@@ -236,3 +236,7 @@ ifdef(`distro_redhat',`
 /var/named/chroot/dev/random -c	gen_context(system_u:object_r:random_device_t,s0)
 /var/named/chroot/dev/zero -c	gen_context(system_u:object_r:zero_device_t,s0)
 ')
+
+ifdef(`distro_gentoo',`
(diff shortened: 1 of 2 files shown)


pebenito:
This doesn't seem complete. This will fix the label of the socket on relabel, but what creates `/var/empty/dev`? Perhaps it should be `device_t` and then the socket will be labeled naturally when it is created.

pebenito (changes requested):
This doesn't seem complete. This will fix the label of the socket on relabel, but what creates `/var/empty/dev`? Perhaps it should be `device_t` and then the socket will be labeled naturally when it is created.

montjoie:
Ok I do the same than /var/named/chroot/dev

pebenito (approved):

PR #574, opened 2022-12-27
mount: dbus interface must be optional

montjoie:
On gentoo, when emerging selinux-base-policy, the post install (loading policy) fail due to a missing type. This is due to mount.te using a dbus interface and the dbus module is not present. Fix this by setting the dbus interface as optional;


0xC0ncord:
You need to add a DCO signoff to your commit.

Also you should move this new optional block below with the other ones.

pebenito (approved):

PR #576, opened 2022-12-28
munin: disk-plugin: transition to fsadm

montjoie:
smart_ plugin currently execute smartctl on the disk_munin_plugin_t domain. But lot of rules are still missing for a correct smartctl execution. Instead of duplicating most of all fsadm rules, it is easier to transition to the correct domain.

### DIFF
--- a/policy/modules/services/munin.if
+++ b/policy/modules/services/munin.if
@@ -43,6 +43,23 @@ template(`munin_plugin_template',`
 
 ########################################
 ## <summary>
+##	Permit to read/write Munin TCP sockets
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`munin_rw_tcp_sockets',`
+	gen_require(`
+		type munin_t;
+	')
+	allow $1 munin_t:tcp_socket rw_socket_perms;
+')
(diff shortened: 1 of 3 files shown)


policy/modules/services/munin.if
```diff
@@ -189,3 +189,20 @@ interface(`munin_admin',`
...
+## <summary>
+##	Permit to read/write Munin TCP sockets
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`munin_readwrite_tcp_sockets',`
```
pebenito, line 203: The name should have `rw` instead of `readwrite`.

policy/modules/services/munin.te
```diff
@@ -260,7 +260,7 @@ optional_policy(`
 ')
 
 optional_policy(`
-	fstools_exec(disk_munin_plugin_t)
+	fstools_run(disk_munin_plugin_t, system_r)
```
pebenito, line 263: I think `fstools_domtrans()` is a better fit here. The `run` interfaces for intended for interactive running by users.

pebenito (changes requested):

pebenito (approved):

PR #580, opened 2023-01-04
munin: add fc for munin-node plugin state

montjoie:
Gentoo deploy munin-node plugin state in /var/lib/munin-node


pebenito (changes requested):
These should be in `distro_gentoo` blocks.

pebenito (approved):

PR #581, opened 2023-01-09
RFC: munin: handle plugins location for gentoo

montjoie:
On gentoo, munin plugins are installed in /usr/libexec/munin/plugins instead of /usr/share/munin/plugins.

WARNING: I do not like this way, but all my tries to do cleaner is not working.
See: https://github.com/montjoie/refpolicy/commit/529d8528656ebac89acca81c9ef3c1a5b70afca4

Any help on why (libexec|share) it is not working is welcome.


pebenito:
The regex probably doesn't work due to the file context sorting. This would probably be better served as a path substitution in the Gentoo policy. cc @perfinion @0xC0ncord

0xC0ncord:
This was discussed on IRC last week. I agree I think a path substituion would work best as far as Gentoo policy is concerned. cc @perfinion

PR #585, opened 2023-01-10
selinuxutil: permit load_policy to use portage ptys

montjoie:
Each time portage build and install a new SELinux policy I got the following AVC: allow load_policy_t portage_devpts_t:chr_file { read write };

### DIFF
--- a/policy/modules/system/selinuxutil.te
+++ b/policy/modules/system/selinuxutil.te
@@ -216,6 +216,7 @@ optional_policy(`
 
 optional_policy(`
 	portage_dontaudit_use_fds(load_policy_t)
+	portage_dontaudit_use_inherited_ptys(load_policy_t)
 ')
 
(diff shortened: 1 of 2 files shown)


policy/modules/system/selinuxutil.te
```diff
@@ -223,6 +223,10 @@ optional_policy(`
         unconfined_dontaudit_read_pipes(load_policy_t)
 ')
 
+optional_policy(`
+	portage_use_inherited_ptys(load_policy_t)
```
pebenito, line 227: It would seem that the ptys are not normally inherited, as fds are not inherited (see line 218) so this should probably be a dontaudit instead. It should also be in the above portage block too.

pebenito (changes requested):

pebenito (approved):

PR #588, opened 2023-01-13
Patches for Debian/Testing with /run/motd.d directory

etbe:


policy/modules/services/ssh.te
```diff
@@ -260,7 +260,11 @@ corenet_sendrecv_xserver_server_packets(sshd_t)
 
 ifdef(`distro_debian',`
 	allow sshd_t self:process { getcap setcap };
+	auth_read_pam_motd_files(sshd_t)
 	auth_use_pam_motd_dynamic(sshd_t)
+
+	# for unattended-upgrades notices.  Should we have a domain transition?
+	files_read_var_lib_files(sshd_t)
```
pebenito, line 267: I'll take it for now. Ideally there should be no `var_lib_t` files.

policy/modules/system/authlogin.if
```diff
@@ -118,6 +118,44 @@ interface(`auth_use_pam_motd_dynamic',`
...
+## <summary>
+##	Write files in /run/motd.d directory
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`auth_write_pam_motd_files',`
```
pebenito, line 131: This should be `auth_manage_pam_motd_files()`.

policy/modules/system/locallogin.te
```diff
@@ -154,6 +154,7 @@ ifdef(`init_systemd',`
 ')
 
 ifdef(`distro_debian',`
+	auth_read_pam_motd_files(local_login_t)
```
pebenito, line 157: It seems that it would make more sense to add this to `auth_login_pgm_domain()` so any login can read these, no?

pebenito (changes requested):

pebenito:
> + files_read_var_lib_files(sshd_t) I'll take it for now. Ideally there should be no `var_lib_t` files.
> What do you suggest as an alternative?

`pam_motd_runtime_t` might be ok. I don't have a strong preference at this point, which is why I'd take the `var_lib_t` access now.

PR #592, opened 2023-01-27
fsadm: add domain for update-smart-drivedb

montjoie:
update-smart-drivedb is a fsadm_t like but with access to network, so Since it do network access, and dont access any hardware, let's add its own domain.

### DIFF
--- a/policy/modules/services/smartmon.te
+++ b/policy/modules/services/smartmon.te
@@ -120,3 +120,6 @@ optional_policy(`
 	seutil_sigchld_newrole(fsdaemon_t)
 ')
 
+optional_policy(`
+	fstools_read_fsadm_db_files(fsdaemon_t)
+')
--- a/policy/modules/services/smartmon.te
+++ b/policy/modules/services/smartmon.te
@@ -33,6 +33,10 @@ ifdef(`enable_mls',`
 	init_ranged_daemon_domain(fsdaemon_t, fsdaemon_exec_t, mls_systemhigh)
 ')
 
+type smartmon_update_drivedb_t;
+type smartmon_update_drivedb_exec_t;
+init_system_domain(smartmon_update_drivedb_t, smartmon_update_drivedb_exec_t)
+
 ########################################
 #
 # Local policy
@@ -113,6 +117,10 @@ tunable_policy(`smartmon_3ware',`
 ')
 
 optional_policy(`
+	fstools_read_fsadm_db_files(fsdaemon_t)
+')
+
+optional_policy(`
 	mta_send_mail(fsdaemon_t)
 ')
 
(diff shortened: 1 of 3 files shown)


policy/modules/system/fstools.te
```diff
@@ -176,6 +181,10 @@ ifdef(`distro_debian',`
 
 ifdef(`distro_gentoo',`
```
perfinion, line 182: Up to @pebenito if this should be in the gentoo-specific block. I don't imagine this would be much different on other distros?

policy/modules/system/fstools.te
```diff
@@ -234,3 +244,31 @@ optional_policy(`
 	xen_append_log(fsadm_t)
 	xen_rw_image_files(fsadm_t)
 ')
+
+########################################
+#
+# fsadm_update_drivedb policy
+#
+
+allow fsadm_update_drivedb_t self:fifo_file { getattr read write ioctl };
```
pebenito, line 253: `rw_fifo_file_perms` instead of individual perms.

policy/modules/system/fstools.te
```diff
@@ -234,3 +244,31 @@ optional_policy(`
...
+corecmd_exec_bin(fsadm_update_drivedb_t)
+corecmd_exec_shell(fsadm_update_drivedb_t)
+
+corenet_sendrecv_http_client_packets(fsadm_update_drivedb_t)
+corenet_tcp_connect_http_port(fsadm_update_drivedb_t)
+
+files_read_etc_files(fsadm_update_drivedb_t)
+
+gpg_exec_agent(fsadm_update_drivedb_t)
+gpg_exec(fsadm_update_drivedb_t)
```
pebenito, line 267: This should be optional.

policy/modules/system/fstools.te
```diff
@@ -234,3 +244,31 @@ optional_policy(`
...
+
+miscfiles_read_generic_certs(fsadm_update_drivedb_t)
+miscfiles_read_localization(fsadm_update_drivedb_t)
+
+sysnet_dns_name_resolve(fsadm_update_drivedb_t)
+
+xdg_read_config_files(fsadm_update_drivedb_t)
```
pebenito, line 274: Optional.

pebenito (changes requested):

policy/modules/system/fstools.te
```diff
@@ -10,6 +10,11 @@ type fsadm_exec_t;
 init_system_domain(fsadm_t, fsadm_exec_t)
 role system_r types fsadm_t;
 
+type fsadm_update_drivedb_t;
```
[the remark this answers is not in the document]
montjoie: do you have a name for the new module ? I dont have good idea, smartmon_update ?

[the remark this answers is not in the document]
0xC0ncord: I'm not sure I agree this should be split out into its own module. I think it should just be moved into `smartmon` instead.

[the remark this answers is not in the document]
0xC0ncord: > I'm not sure I agree this should be split out into its own module. I think it should just be moved into `smartmon` instead.

And also just rename it to something like `smartmon_update_drivedb_t`.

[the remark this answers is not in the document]
pebenito: > I'm not sure I agree this should be split out into its own module. I think it should just be moved into `smartmon` instead.

Yes, that makes more sense.

pebenito:
@montjoie there are many unrelated commits here. Please remove them.

policy/modules/services/smartmon.te
```diff
@@ -23,6 +23,10 @@ init_script_file(fsdaemon_initrc_exec_t)
 type fsdaemon_runtime_t alias fsdaemon_var_run_t;
 files_runtime_file(fsdaemon_runtime_t)
 
+type smartmon_update_drivedb_t;
+type smartmon_update_drivedb_exec_t;
+init_system_domain(smartmon_update_drivedb_t, smartmon_update_drivedb_exec_t)
```
pebenito, line 28: This should move down to line 39 to keep the declarations in order..

policy/modules/services/smartmon.te
```diff
@@ -120,3 +124,47 @@ optional_policy(`
 	seutil_sigchld_newrole(fsdaemon_t)
 ')
 
+optional_policy(`
+	fstools_read_fsadm_db_files(fsdaemon_t)
+')
```
pebenito, line 129: This should go above the `mta_send_mail()` block.

policy/modules/system/fstools.if
```diff
@@ -264,3 +264,42 @@ interface(`fstools_manage_swap_files',`
...
+interface(`fstools_manage_fsadm_db_files',`
+	gen_require(`
+		type fsadm_db_t;
+	')
+
+	manage_dirs_pattern($1, fsadm_db_t, fsadm_db_t)
+	manage_files_pattern($1, fsadm_db_t, fsadm_db_t)
+	manage_sock_files_pattern($1, fsadm_db_t, fsadm_db_t)
+	allow $1 fsadm_db_t:dir watch;
+')
```
pebenito, line 305: Move to line 120

policy/modules/system/fstools.if
```diff
@@ -264,3 +264,42 @@ interface(`fstools_manage_swap_files',`
...
+#
+interface(`fstools_manage_fsadm_db_files',`
+	gen_require(`
+		type fsadm_db_t;
+	')
+
+	manage_dirs_pattern($1, fsadm_db_t, fsadm_db_t)
+	manage_files_pattern($1, fsadm_db_t, fsadm_db_t)
+	manage_sock_files_pattern($1, fsadm_db_t, fsadm_db_t)
+	allow $1 fsadm_db_t:dir watch;
```
pebenito, line 304: Please move to a new interface.

pebenito (changes requested):

pebenito (approved):

PR #599, opened 2023-02-27
Set label systemd-oomd

bluca:
Feb 24 19:02:53 localhost audit[1664]: AVC avc:  denied  { write } for  pid=1664 comm="systemd-oomd" path=2F6D656D66643A646174612D6664202864656C6574656429 dev="tmpfs" ino=2051 scontext=system_u:system_r:initrc_t tcontext=system_u:object_r:tmpfs_t tclass=file permissive=1

Needs to manage cgroups and kill processes, so make it init_exec_t


0xC0ncord:
I very strongly disagree with letting `systemd-oomd` run as `init_t`. It would be far better to create a separate domain for it, e.g. `systemd_oomd_t` and give it only the access that it needs instead of the wide range of privileges that `init_t` has.

pebenito:
I'm fine with this as an interim step until we have a systemd_oomd_t domain.

0xC0ncord:
> I'm fine with this as an interim step until we have a systemd_oomd_t domain.

I think that's a fair compromise. I'd be willing to help create the domain but I don't think I will be able to test it fully as I don't use `systemd-oomd`.

pebenito (approved):

PR #602, opened 2023-03-02
Define user namespace object class.

pebenito:
Also add permission to chromium.

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -1072,3 +1072,8 @@ class io_uring
 	override_creds
 	sqpoll
 }
+
+class user_namespace
+{
+	create
+}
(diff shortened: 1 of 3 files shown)


cgzones:
`user_namespace : create` should probably also be granted for `mozilla_t` (firefox) and `init_t` (systemd, systemd-coredump)

pebenito:
> `user_namespace : create` should probably also be granted for `mozilla_t` (firefox) and `init_t` (systemd, systemd-coredump)

How does it come into play for systemd-coredump?

0xC0ncord:
Please also add this to `container_engine_domain` (container).

cgzones:
> How does it come into play for systemd-coredump?

I am not sure; maybe if a process gets dumped which created a user namespace themself.

PR #603, opened 2023-03-07
More various fixes

0xC0ncord:
Another set of various fixes, mostly systemd-centric.


policy/modules/kernel/files.if
```diff
@@ -1711,6 +1711,25 @@ interface(`files_dontaudit_relabel_config_files',`
...
+##	Relabel configuration symlinks.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+##
+#
+interface(`files_relabel_config_lnk_files',`
```
pebenito, line 1725: Please use "symlinks" instead of "lnk_files".

policy/modules/kernel/selinux.if
```diff
@@ -200,6 +200,25 @@ interface(`selinux_dontaudit_getattr_fs',`
...
+##	Get the attributes of the selinuxfs
+##	directory.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`selinux_getattr_dir',`
```
pebenito, line 214: Please use "dirs".

policy/modules/kernel/selinux.if
```diff
@@ -276,6 +295,24 @@ interface(`selinux_dontaudit_read_fs',`
...
+## <summary>
+##	Mount on the selinuxfs directory.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`selinux_mounton_dir',`
```
pebenito, line 308: "dirs" as above.

policy/modules/services/redis.te
```diff
@@ -49,6 +49,8 @@ files_search_var_lib(redis_t)
 manage_dirs_pattern(redis_t, redis_runtime_t, redis_runtime_t)
 manage_files_pattern(redis_t, redis_runtime_t, redis_runtime_t)
 manage_lnk_files_pattern(redis_t, redis_runtime_t, redis_runtime_t)
+manage_sock_files_pattern(redis_t, redis_runtime_t, redis_runtime_t)
+files_runtime_filetrans(redis_t, redis_runtime_t, { dir sock_file })
```
pebenito, line 53: Any reason not have `dir file lnk_file` too?

policy/modules/services/zfs.if
```diff
@@ -104,6 +104,26 @@ interface(`zfs_read_config',`
...
+##	automatic transition to the zpool
+##	cache type.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`zfs_create_zpool_cache',`
```
pebenito, line 119: "filetrans" instead of "create" otherwise it's misleading.

policy/modules/system/fstools.te
```diff
@@ -140,6 +140,9 @@ files_manage_mnt_dirs(fsadm_t)
 # for tune2fs
 files_search_all(fsadm_t)
 
+# for fsck to read utab
+mount_read_runtime_files(fsadm_t)
```
pebenito, line 144: Nit: out of order

policy/modules/system/logging.te
```diff
@@ -564,6 +566,8 @@ ifdef(`init_systemd',`
 
 	systemd_manage_journal_files(syslogd_t)
 	systemd_watch_journal_dirs(syslogd_t)
+	systemd_dontaudit_relabelfrom_journal_files(syslogd_t)
+	systemd_dontaudit_relabelto_journal_files(syslogd_t)
```
pebenito, line 570: I'm not dontaudit is the right choice. I've very recently seen some relabeling on a system that has the journal on tmpfs until the writable storage is mounted. However, I haven't had a chance to fully investigate.

policy/modules/system/systemd.fc
```diff
@@ -50,6 +51,7 @@
 
 # Systemd unit files
 HOME_DIR/\.config/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home_t,s0)
+HOME_DIR/\.config/containers/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home_t,s0)
```
pebenito, line 54: Nit: out of order

policy/modules/system/systemd.if
```diff
@@ -1402,6 +1402,24 @@ interface(`systemd_read_userdb_runtime_files', `
...
+## <summary>
+##  Manage symbolic links under /run/systemd/userdb.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_manage_userdb_runtime_lnk_files', `
```
pebenito, line 1415: "symlinks"

policy/modules/system/udev.te
```diff
@@ -261,7 +261,9 @@ ifdef(`distro_redhat',`
 ifdef(`init_systemd',`
 	files_search_kernel_modules(udev_t)
 
-	fs_read_cgroup_files(udev_t)
+	fs_create_cgroup_dirs(udev_t)
+	fs_create_cgroup_files(udev_t)
+	fs_rw_cgroup_files(udev_t)
```
pebenito, line 266: Can you put a comment about what it's doing with the cgroups?

pebenito (changes requested):
Minor stuff.

policy/modules/services/redis.te
0xC0ncord: I see no harm in adding the access for those objects too. I didn't add them initially as I didn't see any AVCs related to them at the time.

policy/modules/system/logging.te
0xC0ncord: I dontaudited these because I didn't see any misbehavior on my systems if the access was denied, but I agree I think `allow` is better here.

policy/modules/system/systemd.if
```diff
@@ -1757,6 +1775,24 @@ interface(`systemd_watch_journal_dirs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_relabelfrom_journal_files',`
+	gen_require(`
+		type systemd_journal_t;
+	')
+
+	dontaudit $1 systemd_journal_t:file relabelfrom_file_perms;
```
pebenito, line 1793: Still `dontaudit`

pebenito (changes requested):

policy/modules/system/systemd.if
0xC0ncord: oops! Fixed it. Thanks!

pebenito (approved):

PR #607, opened 2023-03-15
Add support for memory pressure notifications protocol

bluca:
See: https://github.com/systemd/systemd/blob/main/docs/MEMORY_PRESSURE.md

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -1171,6 +1171,15 @@ ifdef(`init_systemd',`
 	systemd_start_power_units(initrc_t)
 	systemd_watch_networkd_runtime_dirs(initrc_t)
 
+	# Ensures the memory.pressure cgroup file is labelled differently, so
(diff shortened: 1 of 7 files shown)


policy/modules/system/init.te
```diff
@@ -1171,6 +1171,16 @@ ifdef(`init_systemd',`
 	systemd_start_power_units(initrc_t)
 	systemd_watch_networkd_runtime_dirs(initrc_t)
 
+	# Ensures the memory.pressure cgroup file is labelled differently, so
+	# that processes can manage it without having access to the rest of the
+	# cgroup tree. This is a special file so each open is an independent,
+	# separate instance that cannot affect already opened ones, so it is not
+	# necessary to lock it down on a process-by-process base. This is useful
```
pebenito, line 1178: So if a malicious process opens `/sys/fs/cgroup/system.slice/systemd-resolved.service/memory.pressure`, it can't affect systemd-resolved?

PR #608, opened 2023-03-22
dovecot: add missing permissions

montjoie:
I use dovecot for IMAP hosting and several rules are missing.

### DIFF
--- a/policy/modules/services/dovecot.te
+++ b/policy/modules/services/dovecot.te
@@ -124,8 +124,9 @@ create_files_pattern(dovecot_t, dovecot_var_log_t, dovecot_var_log_t)
 setattr_files_pattern(dovecot_t, dovecot_var_log_t, dovecot_var_log_t)
 logging_log_filetrans(dovecot_t, dovecot_var_log_t, { file dir })
 
+allow dovecot_t dovecot_spool_t:dir watch;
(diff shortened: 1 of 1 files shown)


policy/modules/services/dovecot.te
```diff
@@ -124,6 +124,8 @@ create_files_pattern(dovecot_t, dovecot_var_log_t, dovecot_var_log_t)
 setattr_files_pattern(dovecot_t, dovecot_var_log_t, dovecot_var_log_t)
 logging_log_filetrans(dovecot_t, dovecot_var_log_t, { file dir })
 
+allow dovecot_t dovecot_spool_t:file map;
```
0xC0ncord, line 127: This can be combined with the below `manage_files_pattern()` to `mmap_manage_files_pattern()`.

policy/modules/services/dovecot.te
```diff
@@ -353,8 +355,11 @@ stream_connect_pattern(dovecot_deliver_t, dovecot_runtime_t, dovecot_runtime_t,
 
 can_exec(dovecot_deliver_t, dovecot_deliver_exec_t)
 
+allow dovecot_deliver_t self:process signal;
```
pebenito, line 358: This should be the first line of the block, line 341

pebenito (changes requested):

pebenito (approved):

PR #609, opened 2023-03-25
path marking for vlc(mplayer_t)

freedom1b2830:
path marking for vlc(https://www.videolan.org/vlc/)
*To write these paths, the domain still needs to be granted access to xdg_{data,config,cache}_t


pebenito (changes requested):
Rules look ok, but it doesn't follow the style guide in terms of how the lines are placed, particularly the .fc file change and the first hunk of the .te file change.

pebenito:
```text
mplayer_tmpfs_t
mplayer_xdg_cache_t
```

pebenito (changes requested):
Looks good. Please squash the two commits into one and then I can merge.

pebenito (approved):

PR #612, opened 2023-04-18
container: set default context for local-path-provisioner

jcpunk:
The kubernetes local-path-provisioner uses either
/opt/local-path-provisioner or
/var/local-path-provisioner for its physical volumes

### DIFF
--- a/policy/modules/services/container.fc
+++ b/policy/modules/services/container.fc
@@ -32,6 +32,8 @@ HOME_DIR/\.docker(/.*)?		gen_context(system_u:object_r:container_conf_home_t,s0)
 
 /opt/cni(/.*)?		gen_context(system_u:object_r:container_plugin_t,s0)
 
+/opt/local-path-provisioner(/.*)?		gen_context(system_u:object_r:container_file_t,s0)
+
 /etc/containers(/.*)?		gen_context(system_u:object_r:container_config_t,s0)
 /etc/cni(/.*)?		gen_context(system_u:object_r:container_config_t,s0)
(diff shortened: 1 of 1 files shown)


jcpunk:
Moving this to the container specific selinux policy https://github.com/containers/container-selinux/pull/228

0xC0ncord:
container-selinux is an add-on policy for Fedora and derivatives (RHEL, Rocky, etc.) specifically while refpolicy aims to have wider compatibility. Your PR is still relevant here if you're willing to reopen this.

That being said, the only thing I would change is to split the `(var|opt)` regex into separate lines so that it is more consistent with the refpolicy code style.

jcpunk:
I've reopened this PR and in theory made the changes you suggested.

policy/modules/services/container.fc
```diff
@@ -100,6 +100,9 @@ HOME_DIR/\.docker(/.*)?		gen_context(system_u:object_r:container_conf_home_t,s0)
 /var/lib/etcd(/.*)?             gen_context(system_u:object_r:container_file_t,s0)
 /var/lib/kube-proxy(/.*)?		gen_context(system_u:object_r:container_file_t,s0)
 
+/opt/local-path-provisioner(/.*)?		gen_context(system_u:object_r:container_file_t,s0)
```
pebenito, line 103: Please move up to line 34.

pebenito (changes requested):

policy/modules/services/container.fc
jcpunk: Does that look better?

pebenito: Yes, thanks

PR #613, opened 2023-04-24
portage: permit to build on tmpfs

montjoie:
My portage PORTAGE_TMPDIR directory is on a tmpfs, but emerging openssl fail with: Can not execute files in /var/tmp/portage
This is due to missing rules permiting portage to execute files in tmpfs.
After this error, some relabel/manage rules was still missing, so this commit add them also.

### DIFF
--- a/policy/modules/admin/portage.if
+++ b/policy/modules/admin/portage.if
@@ -116,11 +116,15 @@ interface(`portage_compile_domain',`
 	# SELinux-enabled programs running in the sandbox
 	allow $1 portage_tmp_t:file relabel_file_perms;
 
+	can_exec($1, portage_tmpfs_t)
+	manage_dirs_pattern($1, portage_tmpfs_t, portage_tmpfs_t)
 	manage_files_pattern($1, portage_tmpfs_t, portage_tmpfs_t)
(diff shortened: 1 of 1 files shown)


0xC0ncord:
I don't think I agree with adding this access. How are you creating the `tmpfs` on `/var/tmp`? And how are you creating `/var/tmp/portage`?

On my systems my `/var/tmp` is a `tmpfs` created via `systemd-zram-generator` and I have an additional script started after it which creates `/var/tmp/portage` and sets permissions and the SELinux context accordingly. Even if you aren't using systemd there should be a pretty easy and reliable way to make sure `/var/tmp/portage` is created with the right attributes.

0xC0ncord:
I can't remember if portage automatically creates `/var/tmp/portage` if it doesn't already exist. If it does, then a type transition should be all that is needed.

montjoie:
I mount tmpfs only on /var/tmp/portage.

0xC0ncord:
> I mount tmpfs only on /var/tmp/portage.

You can add `rootcontext=system_u:object_r:portage_tmp_t:s0` to the mount options which will set the context properly.

PR #619, opened 2023-05-25
container: rework capabilities

0xC0ncord:
Rework (primarily) non-namespaced capabilities. These accesses are leftovers from earlier policy versions before the container module was introduced that are most likely too coarse for most container applications.

Put all non-namespaced capability accesses for containers behind tunables, borrowing ideas from container-selinux. For the more privileged capabilities (sysadmin, mknod), add a tunable to control both namespaced and non-namespaced access to these operations.

I'm not hard-set on having a tunable for namespaced and a tunable for non-namespaced capabilities each, so if there is any opposition I'd be happy to change it.

### DIFF
--- a/policy/modules/services/container.te
+++ b/policy/modules/services/container.te
@@ -60,10 +60,27 @@ gen_tunable(container_use_ecryptfs, false)
 
 ## <desc>
 ##	<p>
+##	Allow containers to use all capabilities in a
+##	non-namespaced context for various privileged operations
+##	directly on the host.
+##	</p>
+## </desc>
+gen_tunable(container_use_host_all_caps, false)
+
+## <desc>
+##	<p>
 ##	Allow containers to use huge pages.
 ##	</p>
 ## </desc>
 gen_tunable(container_use_hugetlbfs, false)
+
+## <desc>
+##	<p>
+##	Allow containers to use the mknod syscall, e.g. for
+##	creating special device files.
+##	</p>
+## </desc>
+gen_tunable(container_use_mknod, false)
 
 ## <desc>
 ##	<p>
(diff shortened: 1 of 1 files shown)


policy/modules/services/container.te
```diff
@@ -475,6 +527,36 @@ logging_send_audit_msgs(container_t)
...
+')
+
+tunable_policy(`container_use_userns_mknod',`
+	allow container_t self:cap_userns mknod;
+')
+
+tunable_policy(`container_use_userns_sysadmin',`
+	allow container_t self:cap_userns sys_admin;
+')
+
```
pebenito, line 559: Any reason to only have broken-out `mknod` and `sys_admin`? (I'm not asking for a tunable for each cap)

One thing to consider would be that the host cap could possibly imply the `ns` version. e.g. line 556 could be `container_use_userns_sysadmin | container_use_host_sysadmin`.

0xC0ncord: > Any reason to only have broken-out `mknod` and `sys_admin`? (I'm not asking for a tunable for each cap)

This is just how container-selinux handles it and I thought their implementation of this logic made sense.

> One thing to consider would be that the host cap could possibly imply the `ns` version. e.g. line 556 could be `container_use_userns_sysadmin | container_use_host_sysadmin`.

I think that's a great idea -- I don't think there's much reason/functionality to be gained from allowing a capability on the host but not within the user namespace.

pebenito: > > Any reason to only have broken-out `mknod` and `sys_admin`? (I'm not asking for a tunable for each cap)
> 
> This is just how container-selinux handles it and I thought their implementation of this logic made sense.

Ok, I'm fine with that.

pebenito (changes requested):
See above discussion.

0xC0ncord:
Sorry for the long delay -- I've made the change that we discussed previously.

pebenito (approved):

pebenito:
> Sorry for the long delay -- I've made the change that we discussed previously.

No worries. I figured you forgot since it wasn't marked as "request changes."

PR #623, opened 2023-06-29
Add label and interfaces for kernel PSI files

fajs:
The pressure stall information (PSI) special files in /proc/pressure currently don't have a separate file context, and so default to proc_t. Since users need read/write permissions to those files to use PSI, and handing out blanket permissions to proc_t is strongly discouraged, introduce a new proc_psi_t label, as well as interfaces for it.

### DIFF
--- a/policy/modules/kernel/kernel.if
+++ b/policy/modules/kernel/kernel.if
@@ -1537,6 +1537,51 @@ interface(`kernel_read_network_state_symlinks',`
 
 ########################################
 ## <summary>
+##	Allow caller to receive pressure stall information (PSI).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`kernel_read_psi',`
+	gen_require(`
+		type proc_t, proc_psi_t;
+	')
+
+	read_files_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+	read_lnk_files_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+	list_dirs_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+')
+
+########################################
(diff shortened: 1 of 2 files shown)


policy/modules/kernel/kernel.if
```diff
@@ -1972,6 +1972,51 @@ interface(`kernel_rw_net_sysctls',`
...
+	')
+
+	rw_files_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+	read_lnk_files_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+	list_dirs_pattern($1, { proc_t proc_psi_t }, proc_psi_t)
+
+	# kernel requires writers to have CAP_SYS_RESOURCE
+	allow $1 self:capability sys_resource;
+')
+
```
pebenito, line 2019: nit: please move to line 1537. Otherwise it's ok.

policy/modules/kernel/kernel.te
```diff
@@ -106,6 +106,9 @@ optional_policy(`
 type proc_mdstat_t, proc_type;
 genfscon proc /mdstat gen_context(system_u:object_r:proc_mdstat_t,s0)
 
+type proc_psi_t, proc_type;
+genfscon proc /pressure gen_context(system_u:object_r:proc_psi_t,s0)
```
pebenito, line 110: nit: please move to line 111.

pebenito (changes requested):

policy/modules/kernel/kernel.te
fajs: will do.

policy/modules/kernel/kernel.if
fajs: will do.

fajs:
Thanks for the review! Addressed your comments.

pebenito:
Thanks!

PR #633, opened 2023-08-16
Add interfaces for systemd-hostnamed units

dsugar100:
node=localhost type=USER_AVC msg=audit(1689811749.504:399): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='avc:  denied  { start } for auid=n/a uid=0 gid=81 path="/usr/lib/systemd/system/systemd-hostnamed.service" cmdline="/usr/bin/dbus-broker-launch --scope system --audit" function="bus_unit_method_start_generic" scontext=system_u:system_r:system_dbusd_t:s0 tcontext=system_u:object_r:systemd_unit_t:s0 tclass=service permissive=0 exe="/usr/lib/systemd/systemd" sauid=0 hostname=? addr=?  terminal=?' UID="root" AUID="unset" AUID="root" UID="root" GID="dbus" SAUID="root"

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -1538,6 +1538,44 @@ interface(`systemd_dbus_chat_machined',`
 
 ########################################
 ## <summary>
+##	Allow specified domain to start and stop systemd-hostnamed units
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_startstop_hostnamed',`
+	gen_require(`
+		type systemd_hostnamed_unit_t;
+		class service { start stop };
+	')
+
+	allow $1 systemd_hostnamed_unit_t:service { start stop };
+')
+
+########################################
+## <summary>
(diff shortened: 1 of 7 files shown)


policy/modules/system/systemd.if
```diff
@@ -1536,6 +1536,44 @@ interface(`systemd_dbus_chat_machined',`
 	allow systemd_machined_t $1:dbus send_msg;
 ')
 
+########################################
+## <summary>
+##	Allow specified domain to start systemd-hostnamed units
```
pebenito, line 1541: Missing stop.

pebenito (changes requested):
Minor doc issue.

dsugar100:
I'm kind of also wondering if it would be preferable to just allow dbusd_t to start/stop all systemd units?  Thoughts?

pebenito:
> I'm kind of also wondering if it would be preferable to just allow system_dbusd_t to start/stop all systemd units? Thoughts?

Possibly. Potentially any unit could be dbus activated.

dsugar100:
Replaced by merge request #655

PR #635, opened 2023-08-23
The kernel domain should be able to mounton default and runtime directories

gtrentalancia:
The kernel domain should be able to mounton runtime directories during switch_root, otherwise parts of the boot process might fail on some systems (for example, the udev daemon).
---
     policy/modules/kernel/kernel.te |    1 +
     1 file changed, 1 insertion(+)


gtrentalancia:
Similarly, the kernel domain should be able to mounton default directories during switch_root.

Corresponding suspicious permissions are removed from the init domain, altough this might need further testing on a wider number of systems.

pebenito (approved):

Issue #362, opened 2021-04-17
spamassassin is not able to update its rules from the network


gtrentalancia:
The current spamassassin policy module does not allow the spamassassin package to update its spam detection rules from the network using the "sa-update" script.

Fixed in the following proposed patches:

https://github.com/SELinuxProject/refpolicy/pull/360/commits/2acbe3051b4a901c215ea8b596848a7c424ce149
https://github.com/SELinuxProject/refpolicy/pull/360/commits/d7bd0fcbc72d5a89850ecbf24ac208bf3a03dd02
https://github.com/SELinuxProject/refpolicy/pull/360/commits/714531b923c6ad1b2fdb0dbd57b71deef881518a

gtrentalancia:
I didn't get the "help wanted" notification and I have noticed only now...

Do you want to fix the problem ?

gtrentalancia:
This is the help I can offer to resolve the issue: https://github.com/SELinuxProject/refpolicy/pull/636

I have developed, tested and deployed it locally a couple of years ago and it works fine...

Spam filters do not resolve the spam issue but only attempt to mitigate it, however using outdated or buggy spam filtering rules certainly does not improve things !

I hope it helps !

PR #636, opened 2023-08-28
Let spamassassin update its rules from the network

gtrentalancia:
Spamassassin is currently unable to fetch updates filtering rules from the network.

This pull request aims to fix https://github.com/SELinuxProject/refpolicy/issues/362

### DIFF
--- a/policy/modules/apps/gpg.if
+++ b/policy/modules/apps/gpg.if
@@ -197,6 +197,44 @@ interface(`gpg_entry_type',`
 
 ########################################
 ## <summary>
+##	Execute the gpg_agent in the caller domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`gpg_agent_exec',`
+	gen_require(`
+		type gpg_agent_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	can_exec($1, gpg_agent_exec_t)
+')
+
+######################################
+## <summary>
+##	Make gpg_agent executable files an
+##	entrypoint for the specified domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	The domain for which gpg_agent_exec_t is an entrypoint.
+##	</summary>
+## </param>
+#
+interface(`gpg_agent_entry_type',`
+	gen_require(`
+		type gpg_agent_exec_t;
+	')
+
+	domain_entry_file($1, gpg_agent_exec_t)
+')
+
+########################################
+## <summary>
 ##	Send generic signals to gpg.
 ## </summary>
 ## <param name="domain">
@@ -343,6 +381,26 @@ interface(`gpg_runtime_filetrans',`
 
 ########################################
 ## <summary>
+##	Do not audit attempt to getattr gpg runtime dirs.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`gpg_dontaudit_getattr_gpg_runtime_dirs',`
+	gen_require(`
+		type gpg_runtime_t;
+	')
+
+	files_dontaudit_search_runtime($1)
+
+	dontaudit $1 gpg_runtime_t:dir getattr;
+')
+
+########################################
+## <summary>
 ##	filetrans in gpg_secret_t dirs
 ## </summary>
 ## <param name="domain">
@@ -414,6 +472,8 @@ interface(`gpg_dontaudit_search_user_secrets',`
 		type gpg_secret_t;
 	')
 
+	userdom_dontaudit_search_user_home_dirs($1)
+
 	dontaudit $1 gpg_secret_t:dir search_dir_perms;
 ')
 
@@ -434,4 +494,24 @@ interface(`gpg_list_user_secrets',`
 
 	list_dirs_pattern($1, gpg_secret_t, gpg_secret_t)
 	userdom_search_user_home_dirs($1)
+')
+
+########################################
+## <summary>
+##	Do not audit attempt to search gpg user secrets dirs.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`gpg_dontaudit_search_user_secrets_dirs',`
+	gen_require(`
+		type gpg_secret_t;
+	')
+
+	userdom_dontaudit_search_user_home_dirs($1)
+
+	dontaudit $1 gpg_secret_t:dir search;
 ')
--- a/policy/modules/services/spamassassin.te
+++ b/policy/modules/services/spamassassin.te
@@ -104,11 +104,7 @@ type spamd_update_t;
 type spamd_update_exec_t;
 application_domain(spamd_update_t, spamd_update_exec_t)
 role spamd_update_roles types spamd_update_t;
-
-type spamd_update_gpg_t;
-typealias spamd_update_gpg_t alias spamd_gpg_t;
-domain_type(spamd_update_gpg_t)
-role system_r types spamd_update_gpg_t;
+role system_r types spamd_update_t;
 
 ########################################
 #
@@ -529,7 +525,8 @@ optional_policy(`
 #
 
 allow spamd_update_t self:capability { dac_override dac_read_search };
-allow spamd_update_t self:process signal;
+allow spamd_update_t self:process { setrlimit sigchld signal };
+allow spamd_update_t self:fd use;
 allow spamd_update_t self:fifo_file manage_fifo_file_perms;
 allow spamd_update_t self:unix_stream_socket create_stream_socket_perms;
 
@@ -561,6 +558,7 @@ auth_dontaudit_read_shadow(spamd_update_t)
 miscfiles_read_generic_certs(spamd_update_t)
 miscfiles_read_localization(spamd_update_t)
 
+userdom_use_unpriv_users_fds(spamd_update_t)
 userdom_use_inherited_user_terminals(spamd_update_t)
 userdom_dontaudit_search_user_home_dirs(spamd_update_t)
 userdom_dontaudit_search_user_home_content(spamd_update_t)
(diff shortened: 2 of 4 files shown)


pebenito:
I will need some time to better understand this.

gtrentalancia:
Yes, take your time.

If you remember a couple of years ago, I did propose a similar patch to let fsdaemon (update-smart-drivedb) do the equivalent (update the drives database file from the network)...

gtrentalancia:
It's quite simple, in addition to being equivalent to existing patterns.

A new role is added for a process that runs the spamassassin Bayesian filter updater: because it is an administrative task, the role is only added to sysadm and staff (as far as I remember the former is needed for interactive use, while the latter is needed for cron jobs).

The whole updating process is made modular and divided into two execution distinct domains for increased security and isolation between networking and cryptographic operations: a first domain spamd_update_t is used to control the operation, managing temporary and persistent files and for TCP/IP networking, while a second domain spamd_update_gpg_t (activated by a domain transition upon executing GPG) is used exclusively for cryptographic operations, id est to verify the GPG signature of the data fetched from the network by using the former spamd_update_t domain and controlling process.

I hope my explanation helps understanding the design pattern...

policy/modules/services/spamassassin.te
```diff
@@ -549,10 +555,61 @@ userdom_use_inherited_user_terminals(spamd_update_t)
...
+	gpg_entry_type(spamd_update_gpg_t)
+	gpg_exec(spamd_update_gpg_t)
+	gpg_dontaudit_getattr_gpg_runtime_dirs(spamd_update_gpg_t)
+	gpg_dontaudit_search_user_secrets_dirs(spamd_update_gpg_t)
+	gpg_agent_entry_type(spamd_update_gpg_t)
+	gpg_agent_exec(spamd_update_gpg_t)
+')
+
+optional_policy(`
+	mta_manage_config(spamd_update_gpg_t)
```
pebenito, line 614: I have other issues with this PR, but this is still the big one, as it was last time. I still think this domain should be dropped and the permissions added to `spamd_update_t`.

pebenito (changes requested):

policy/modules/services/spamassassin.te
0xC0ncord: Out of curiosity, why does `spamd_update_t` need to execute `gpg`? Are the spamassassin rules GPG-signed?

0xC0ncord: If that is the case I second just merging this behavior into `spamd_update_t`, analogous to how I implemented this behavior for container image signature verification: https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/services/container.te#L780-L785

gtrentalancia: Yes, the spamassassin rules are signed with GPG key whose fingerprints are embedded in the sa-update script.

gtrentalancia: By having just one domain (spamd_update_t) instead of two (spamd_update_t and spamd_update_gpg_t) the level of security is decreased, because spamd_update_t is network-facing and if it is compromised it can potentially allow a remote attacker to access secret GPG files.

gtrentalancia: So why do you want to drop the safer two domains design and employ just one domain for all operations ?

dsugar100: > By having just one domain (spamd_update_t) instead of two (spamd_update_t and spamd_update_gpg_t for the child GPG verification process) the level of security is decreased, because spamd_update_t is network-facing and if it is compromised it can potentially allow a remote attacker to access secret GPG files.

Are there secret keys on each system or just the public key on each system?  I would hope that the secret keys are kept secret by spamassassin.

gtrentalancia: Secret (or private) crypto keys are stored in the home directory under ~/.gnupg/secring.gpg if a user has previously created them (for encrypting emails or signing/encrypting other kinds of documents and files sent over the network).

The trust database is also critical to confidentiality and it is stored under ~/.gnupg/trustdb.gpg.

gtrentalancia: Yes, of course I know that at the moment there is a safety margin.

However, if things can be changed so that they provide a much larger safety margin at no cost, then we end up with a better design which is safer.

pebenito: > However, if things can be changed so that they provide a much larger safety margin

I don't understand what you mean by safety margin.  The policy is denying access.

> at no cost, then we end up with a better design which is safer.

There is a cost. More types means more policy complexity. More policy complexity means increased difficulty to understand and maintain. You're saying this is a better design, but no one else seems to agree (if you're monitoring this thread but not commenting please thumbs up or thumbs down this post to indicate your opinion).

In functional terms, more policy also means more non-swappable kernel memory used.

gtrentalancia: For safety margin, I mean stay away from sensitive domains.

dontaudit does not guarantee anything other than not ending up in the audit log.

neverallow is the only statement which guarantees avoiding a given permission.

Does it make sense ?

gtrentalancia: Nobody else has commented, so it's up to you, I have just committed the amendments you requested with the warning already made...

policy/modules/apps/gpg.if
```diff
@@ -195,6 +195,97 @@ interface(`gpg_entry_type',`
...
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="target_domain">
+##	<summary>
+##	Domain to transition to.
+##	</summary>
+## </param>
+#
+interface(`gpg_agent_spec_domtrans',`
```
pebenito, line 261: "spec_domtrans" means one using setexeccon instead of type_transition. This interface doesn't do that. Since it's not used, it could be dropped.

policy/modules/apps/gpg.if
```diff
@@ -341,6 +432,24 @@ interface(`gpg_runtime_filetrans',`
 	userdom_search_user_runtime($1)
 ')
 
+########################################
+## <summary>
+##	Do not audit attempt to getattr gpg runtime dirs.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
```
pebenito, line 441: This interface is not allowing access.

policy/modules/apps/gpg.if
```diff
@@ -435,3 +544,22 @@ interface(`gpg_list_user_secrets',`
...
 	userdom_search_user_home_dirs($1)
 ')
+
+########################################
+## <summary>
+##	Do not audit attempt to search gpg user secrets dirs.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
```
pebenito, line 554: This interface is not allowing access.

policy/modules/roles/staff.te
```diff
@@ -176,6 +176,7 @@ ifndef(`distro_redhat',`
 
 	optional_policy(`
 		spamassassin_role(staff, staff_t, staff_application_exec_domain, staff_r)
+		spamassassin_run_spamd_update(staff_t, staff_r)
```
pebenito, line 179: Looks redundant due to the change in `spamassassin_role()`.

policy/modules/roles/sysadm.te
```diff
@@ -1019,6 +1019,7 @@ optional_policy(`
 optional_policy(`
 	spamassassin_admin(sysadm_t, sysadm_r)
 	spamassassin_role(sysadm, sysadm_t, sysadm_application_exec_domain, sysadm_r)
+	spamassassin_run_spamd_update(sysadm_t, sysadm_r)
```
pebenito, line 1022: Looks redundant due to the change in `spamassassin_role()`.

policy/modules/services/spamassassin.te
```diff
@@ -29,12 +39,7 @@ gen_tunable(spamd_enable_home_dirs, false)
 ## </desc>
 gen_tunable(rspamd_spamd, false)
 
-type spamd_update_t;
-type spamd_update_exec_t;
-init_system_domain(spamd_update_t, spamd_update_exec_t)
-
-type spamd_update_tmp_t;
```
pebenito, line 36: I don't see an alias added to `spamd_tmp_t` for compatibility.

policy/modules/services/spamassassin.te
```diff
@@ -203,13 +213,17 @@ stream_connect_pattern(spamc_t, { spamd_runtime_t spamd_tmp_t }, { spamd_runtime
 kernel_read_kernel_sysctls(spamc_t)
 kernel_read_system_state(spamc_t)
 
-corenet_all_recvfrom_netlabel(spamc_t)
-corenet_tcp_sendrecv_generic_if(spamc_t)
-corenet_tcp_sendrecv_generic_node(spamc_t)
-corenet_udp_bind_generic_node(spamc_t)
+tunable_policy(`spamassassin_can_network',`
```
pebenito, line 216: This block needs to move down after the `use_samba_home_dirs` block.

policy/modules/services/spamassassin.te
```diff
@@ -338,28 +351,32 @@ can_exec(spamd_t, { spamd_exec_t spamd_compiled_t })
...
 kernel_read_system_state(spamd_t)
 
-corenet_all_recvfrom_netlabel(spamd_t)
-corenet_tcp_sendrecv_generic_if(spamd_t)
-corenet_udp_sendrecv_generic_if(spamd_t)
-corenet_tcp_sendrecv_generic_node(spamd_t)
-corenet_udp_sendrecv_generic_node(spamd_t)
-corenet_tcp_bind_generic_node(spamd_t)
-corenet_udp_bind_generic_node(spamd_t)
+tunable_policy(`spamassassin_can_network',`
```
pebenito, line 354: Needs to move down to be the first tunable block.

pebenito (changes requested):
Minor issues.

policy/modules/apps/gpg.if
gtrentalancia: It simply prevents auditing. It does not grant or deny any permimssion.

policy/modules/apps/gpg.if
gtrentalancia: I will remove it.

policy/modules/apps/gpg.if
gtrentalancia: What do you mean ?

It simply prevents auditing. It does not grant nor deny any permimssion.

policy/modules/services/spamassassin.te
gtrentalancia: Yes, I'll move the block.

policy/modules/apps/gpg.if
gtrentalancia: I suppose you mean the comment... I'll amend that.

policy/modules/apps/gpg.if
gtrentalancia: I suppose you mean the comment... I'll amend that.

policy/modules/services/spamassassin.te
pebenito: There seems to be two copies of this block.

gtrentalancia: I didn't remove it by mistake, thanks for telling me. It has been fixed now !

pebenito (approved):

PR #638, opened 2023-08-30
Minor fix for the gnome module (gconf daemon)

gtrentalancia:
Update the gnome module so that the gconf daemon can accept and listen connections on Unix sockets bringing its own label.
---
 policy/modules/contrib/gnome.te |    2 ++
 1 file changed, 2 insertions(+)


policy/modules/apps/gnome.te
```diff
@@ -80,6 +80,8 @@ optional_policy(`
 # Conf daemon local Policy
 #
 
+allow gconfd_t self:unix_stream_socket { accept listen };
```
pebenito, line 83: Please use `create_stream_socket_perms` instead.

gtrentalancia: I can't foresee a valid reason to add the create socket permissions.

Please see https://github.com/SELinuxProject/refpolicy/pull/639#discussion_r1316142515 for reference about this issue.

gtrentalancia: I have updated the discussion, please see https://github.com/SELinuxProject/refpolicy/pull/639#discussion_r1316346698

gtrentalancia: I think they are inherited socket file descriptors in this case.

gtrentalancia: It might related to dbus...

gtrentalancia: For example, dbus_all_session_domain(), which would prove it does not be granted the "create" socket permission.

What do you say ?

gtrentalancia: I think the socket is created by dbus, which then launches gconfd and the latter inherits the socket file descriptor.

gtrentalancia: I have just modified the dbus policy in order to add the ability to create sockets:

https://github.com/SELinuxProject/refpolicy/pull/667

This is backed by a source code audit...

Similarly I have searched the gconf source code and found no calls to socket() creation function and this reinforces my previous comment. It must be sockets inherited from the dbus daemon process after it executes the various services, including gconfd.

gtrentalancia: Also see the equivalent dbus change for its session bus:

https://github.com/SELinuxProject/refpolicy/pull/671

pebenito: If dbus creates a socket and gconfd_t inherits it, the socket label would not be `gconfd_t`. The fact is that this domain already has a create permission. If you're not going to do the requested change, please close the PR.

gtrentalancia: The create permission comes from: logging_send_syslog_msg

gtrentalancia: Please give me some more time to look at the source code, I have been struggling to find where it eventually creates the socket, so I am still missing something...

gtrentalancia: Unfortunately I could not verify it directly in the source code. However, I have verified it from the syslog files.

The requested change has now been made.

pebenito (approved):

PR #639, opened 2023-08-30
Minor fix for the openoffice module

gtrentalancia:
Update the openoffice module so that it is able to accept and listen connections on Unix sockets with its own label.
---
 policy/modules/contrib/openoffice.te |    2 +-


policy/modules/apps/openoffice.te
```diff
@@ -52,7 +52,7 @@ files_tmp_file(ooffice_tmp_t)
 allow ooffice_t self:process { execmem getsched signal };
 allow ooffice_t self:shm create_shm_perms;
 allow ooffice_t self:fifo_file rw_fifo_file_perms;
-allow ooffice_t self:unix_stream_socket connectto;
+allow ooffice_t self:unix_stream_socket { accept connectto listen };
```
pebenito, line 55: Probably should make this `{ create_stream_socket_perms connectto }` to be comprehensive.

pebenito (changes requested):

policy/modules/apps/openoffice.te
gtrentalancia: The same pattern is already used elsewhere in a very long list of modules, please see the following ones: spamassassin, openvswitch, fail2ban, bluetooth, squid, policykit, courier, zebra, canna, aisexec, pegasus, setroubleshoot, inn, pacemaker, cups, razor, remotelogin, ppp, dbus, sssd, snmp, asterisk, mysql, cron, virt, mpd, munin, apache, dovecot, avahi, postfix, samba, portslave, ctdb, cyrus, clamav, openvpn, soundserver, corosync, amavis, iscsi, xen, logrotate, rpm, apt, passenger, rssh, gnome, webalizer, evolution, pulseaudio, usernetctl, screen, mozilla and userhelper.

So, I don't think the ability to create sockets should be added without a valid reason, which at the moment I can't foresee.

I think the above mentioned patch hunk should be applied as it is.

pebenito: > The same pattern is already used elsewhere in a very long list of modules

And I think they all warrant being fixed. If a socket has the `self` label, then it was almost certainly created by the domain itself; the only case it isn't is if there is a setsockcreate and the socket is inherited. In most if not all of the above cases, the create access is allowed by some other interface.

$ sesearch -A -s ooffice_t -t ooffice_t -c unix_stream_socket -p create policy.33
allow ooffice_t ooffice_t:unix_stream_socket { accept append bind connect connectto **create** getattr getopt ioctl listen read setattr setopt shutdown write };

gtrentalancia: Which interface do you mean ?

That would raise the question of why the "accept/listen/connectto" are not moved to the place where "create" is granted, which would make much more sense. What do you say ?

pebenito: The listen and accept are only necessary if the stream socket is acting as a server. If it is a client, then you don't need listen/accept and need connectto instead. Since this domain is connecting to itself, it needs all three.

gtrentalancia: Yes indeed

gtrentalancia: And I give you a first answer as for the origins of "create":

sysnet_dns_name_resolve() --> nscd_use() --> allow $1 self:unix_stream_socket create_socket_perms

There might be more, I did not check. That is why is pointless to explicitly add "create": it's redundant and not needed.

If you confirm, it is possible to move the accept/listen/connectto permissions to nscd_use(), that would make sense, however there is a drawback: if there are other interface calls that do a similar job, then we get a bug...

Please let me know how you intend to proceed. If you know where all "create" instances come from then, we can fix the nscd_use() bug.

For example, can you please test this: temporarily remove the call to sysnet_dns_name_resolve() and then check again with sesearch.

gtrentalancia: Also consider that there might be inherited socket file descriptors in additions to those created by other interfaces, so even if you move the "accept/listen/connectto" permissions to the interface(s) where "create" is, then you are not done because if those interfaces are later removed, then a bug related to inherited socket file descriptors might arise.

Does that make sense ?

pebenito: If nscd was the only reason openoffice had unix stream sockets, I would agree. The unix stream socket access in `nscd_use()` is so that the caller can connect to nscd; note the stream_connect_pattern(). It is acting as a client to nscd, so it doesn't need listen and accept. Since the access in question here is openoffice connecting to itself, the purpose of the sockets are different; thus, it needs the permissions explicitly, so if someone took away the `nscd_use()` because they don't have nscd, then this other socket access would continue to work.

gtrentalancia: You haven't answered my original question: which interface does currently allow "create" ?

I would like to get the full picture, considering this is a very widely used pattern...

gtrentalancia: In the meanwhile, I have updated the openoffice module only with create socket permissions, because I looked at its source code to make sure it calls socket(AF_UNIX).

pebenito (approved):

PR #641, opened 2023-08-31
Minor miscellaneous fixes for various policy modules

gtrentalancia:


policy/modules/services/accountsd.te
```diff
@@ -21,6 +21,8 @@ files_type(accountsd_var_lib_t)
 # Local policy
 #
 
+dontaudit accountsd_t self:capability dac_read_search;
```
pebenito, line 24: Since `dac_override` is a superset of `dac_read_search`, I'm not seeing the value of `dontaudit`ing this instead of `allow`ing it.

gtrentalancia: From brief testing, it possibly does not need dac_override either.

I propose a revised commit to test this assumption or otherwise drop my dontaudit commit. Please let me know how you prefer to proceed...

pebenito: Looks like the `dac_override` has been there since this module was added in 2010: https://github.com/SELinuxProject/refpolicy/commit/c4834a02d2eed7efbb4faa0f097d077ebfa7bc18#. It may no longer be needed. What distro are you using?

gtrentalancia: I am afraid, I cannot test all existing distributions.

If you agree, I can simply dontaudit both dac capabilities as a test (eventually reporting back, should I later discover it breaks something), but without taking the responsibility of breaking some other system or otherwise drop that patch completely.

Please let me know how you intend to proceed, as that was not the main point of the patchset...

pebenito: I'm not asking you to test all distros, I'm only asking which one you used that did not require this cap.

gtrentalancia: I have not created the accountsd module, so I've now dropped the dontaudit change, as there is the risk that it breaks some other system, including backwards-compatibility.

gtrentalancia: I am not using a distribution, it's a system that I build

gtrentalancia: So, in some way it's similar to Linux From Scratch, if that makes sense...

pebenito: Ok.

I see you dropped this so I'm resolving this.

pebenito (approved):

PR #644, opened 2023-09-02
Allow rsyslog to drop capabilities

dsugar100:
Aug 28 19:01:43 localhost.localdomain audisp-syslog[1565]: node=localhost type=AVC msg=audit(1693249303.693:415): avc:  denied  { setpcap } for  pid=1722 comm="rsyslogd" capability=8 scontext=system_u:system_r:syslogd_t:s0 tcontext=system_u:system_r:syslogd_t:s0 tclass=capability permissive=0
Aug 28 19:01:43 localhost.localdomain rsyslogd[1722]: libcap-ng used by "/usr/sbin/rsyslogd" failed dropping bounding set in capng_apply

### DIFF
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -570,6 +570,11 @@ ifdef(`distro_gentoo',`
 	term_dontaudit_setattr_unallocated_ttys(syslogd_t)
 ')
 
+ifdef(`distro_redhat',`
+	# EL9 configures with --enable-libcap-ng
+	allow syslogd_t self:capability setpcap;
+')
+
 ifdef(`distro_suse',`
 	# suse creates a /dev/log under /var/lib/stunnel for chrooted stunnel
 	files_var_lib_filetrans(syslogd_t, devlog_t, sock_file)


gtrentalancia:
Since this is a non-standard configuration (see the configure script output below), you might want to add a boolean for tunable policy:

  --enable-libcap-ng      Enable dropping capabilities to only the necessary set [default=no]

pebenito:
> Since this is a non-standard configuration (see the configure script output below), you might want to add a boolean for tunable policy:
> 
> --enable-libcap-ng Enable dropping capabilities to only the necessary set [default=no]

Dropping capabilities is a good thing. The only other thing `setpcap` can do is add caps from its bounding set to its inheritable set, but we're still constraining cap use in SELinux, so it can't add arbitrary caps, even if they're in the bounding set. Is there something else I'm missing?

gtrentalancia:
It bears some dangers, as setpcap is quite powerful and at the same time quite dangerous.

The patch tackles recent rsyslog features. I am not an rsyslog developer, however the first thing I would question is why the official developers decided to not allow such functionality by default !

As a second issue, I would consider that the same policy (syslogd_t domain) is "shared" (used) by multiple system logging daemon implementations (not only rsyslog, but also syslogd, syslog-ng, minilogd, systemd-journald and metalog). Therefore, allowing the setpcap capability for one of them (in this case a new feature offered by rsyslog) means allowing such capability for all the possible system logging daemons that share such policy.

Tihird, please take a closer look at the syslogd.te module (syslogd_t local policy): you will notice that you have already questioned the existence of some of the currently granted capabilities, namely net_admin. Indeed it's dodgy and such question has never been answered: more differences exist between different daemons.

Least but not last, I would consider the possibility of unforeseen syslogd or libcap-ng bugs and the undetected presence of compromised binaries...

Does it make sense ?

pebenito:
Thanks for the explanation, that makes sense.

> The patch tackles recent rsyslog features. I am not an rsyslog developer, however the first thing I would question is why the official developers decided to not allow such functionality by default !

You'd have to task them, but I think your other arguments are more compelling than whether or not rsyslog has this function by default.

gtrentalancia:
As for the net_admin capability in syslogd_t local policy, I would like to take a moment to quote the rsyslog man page in order to possibly answer your question:

[SECURITY THREATS]
"... Activating logging over the inet domain sockets will of course expose a system to risks outside of programs or individuals on the local machine."

gtrentalancia:
This new pull request that I created might help to mitigate the above raised issue:

https://github.com/SELinuxProject/refpolicy/pull/652

dsugar100:
Getting back to this.  I would assume they made it a compile option either because some distros don't have libcap-ng and/or may want to use systemd to set the user/group rather than drop caps.  
In either case there are (at least) three paths forward for this change:
1 - always allow syslogd_t to setpcap
2 - wrap this in a boolean
3 - wrap this in a 'redhat' ifdef
4 - do nothing - let this fail (seems like a bad choice)
5 - something else?

Is there a preference for which way we want to see this go?
I think I would prefer 1 or 3.  I'm not sure it is worth a boolean (but if pebenito says boolean I will make one).

gtrentalancia:
My opinion is that "setpcap" should be preferably enclosed in distribution "ifdef" at least until libcap-ng does not become the default configure option for rsyslog or other popular daemons (at the moment it clearly isn't, see above https://github.com/SELinuxProject/refpolicy/pull/644#issuecomment-1705706136).

dsugar100:
Ok, updated this change to move the pcap capability into a rhel ifdef.  I really don't know which other platforms might want the same type of thing.

gtrentalancia:
It's probably just redhat, please see:

https://bugzilla.redhat.com/show_bug.cgi?id=2127403

and

https://bugzilla.redhat.com/show_bug.cgi?id=2149633

You can probably update the latter or mark it as resolved after your change has been merged.

gtrentalancia:
Debian also ships both rsyslog and libcap-ng, however they are probably not linked together (as yet).

pebenito (approved):

PR #647, opened 2023-09-03
Stricter yet more customizable xserver policy and three security bug fixes

gtrentalancia:
Separate the tunable permissions to write xserver tmpfs files from the tunable permissions to write X server shared memory.

This allows a stricter, yet more customizable policy.

Fix a security bug in one xserver module interface which allows to bypass existing tunable policy logic.

### DIFF
--- a/policy/modules/services/xserver.if
+++ b/policy/modules/services/xserver.if
@@ -125,8 +125,12 @@ template(`xserver_restricted_role',`
 	# Client write xserver shm
 	tunable_policy(`allow_write_xshm',`
 		allow $2 xserver_t:shm rw_shm_perms;
+	')
+
+	tunable_policy(`allow_write_xserver_tmpfs',`
 		allow $2 xserver_tmpfs_t:file rw_file_perms;
 	')
+
 	tunable_policy(`xserver_allow_dri',`
 		dev_rw_dri($2)
 	')
@@ -482,6 +486,9 @@ template(`xserver_user_x_domain_template',`
 	# Client write xserver shm
 	tunable_policy(`allow_write_xshm',`
 		allow $2 xserver_t:shm rw_shm_perms;
+	')
+
+	tunable_policy(`allow_write_xserver_tmpfs',`
 		allow $2 xserver_tmpfs_t:file rw_file_perms;
 	')
 ')
--- a/policy/modules/services/xserver.if
+++ b/policy/modules/services/xserver.if
@@ -127,7 +127,7 @@ template(`xserver_restricted_role',`
 		allow $2 xserver_t:shm rw_shm_perms;
 	')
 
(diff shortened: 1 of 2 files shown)


policy/modules/services/xserver.if
```diff
@@ -125,8 +125,12 @@ template(`xserver_restricted_role',`
 	# Client write xserver shm
 	tunable_policy(`allow_write_xshm',`
 		allow $2 xserver_t:shm rw_shm_perms;
+	')
+
+	tunable_policy(`allow_write_xserver_tmpfs',`
```
pebenito, line 130: It seems that this block and the block added below should be 
```
tunable_policy(`allow_write_xshm | allow_write_xserver_tmpfs',`
```
so that shm access works correctly.

gtrentalancia: They are three blocks in total...

I can introduce the change you request, given that X shared memory access is much more intrusive than X server tmpfs rw access.

gtrentalancia: Logic OR is double-pipe as in C syntax, so I have edited that too...

pebenito (approved):

policy/modules/services/xserver.te
```diff
@@ -33,6 +33,14 @@ gen_require(`
 ## </desc>
 gen_tunable(allow_write_xshm, false)
 
+## <desc>
+## <p>
+## Allows clients to write to the X server tmpfs
+## files.
+## </p>
+## </desc>
+gen_tunable(allow_write_xserver_tmpfs, false)
```
pebenito, line 42: On second thought, as with the other tunable you recently added, I'd like to make this name start with `xserver`. How about `xserver_client_rw_tmpfs`?

gtrentalancia: Then allow_write_xshm also needs to be renamed. I'll do that.

gtrentalancia: Also, the "rw" wording would be misleading, as the read permission is always granted...

pebenito: > Then allow_write_xshm also needs to be renamed. I'll do that.

The problem is that breaks backward compatibility.

> Also, the "rw" wording would be misleading, as the read permission is always granted...

Fair enough.

gtrentalancia:
The security bugs due to illicit bypassing of tunable policy logic for the X server shared memory write access turned out to be wider than expected, but it has now hopefully been tackled thoroughly !

policy/modules/services/xserver.te
pebenito: The `allow_write_xshm` needs to be restored.

gtrentalancia: I have now restored the original naming, so it should be ready to merge.

pebenito (approved):

PR #648, opened 2023-09-04
Fix two long-standing build failures

gtrentalancia:
This patchset fixes two long-standing build failures due to semantic errors in the systemd and dbus modules (interfaces/templates).

### DIFF
--- a/policy/modules/services/dbus.if
+++ b/policy/modules/services/dbus.if
@@ -60,7 +60,7 @@ interface(`dbus_exec',`
 template(`dbus_role_template',`
 	gen_require(`
 		class dbus { send_msg acquire_svc };
-		attribute session_bus_type;
+		type $1_dbusd_t;
 		type system_dbusd_t, dbusd_exec_t;
 		type session_dbusd_tmp_t, session_dbusd_home_t;
 		type session_dbusd_runtime_t;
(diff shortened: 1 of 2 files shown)


pebenito:
What configuration results in this build failure? It would be nice to improve the build tests if possible.

gtrentalancia:
If for configuration you mean the policy type as mls, mcs, standard, then I suppose all of them can possibly trigger such build error.

The problem is two semantic errors (which is why it does not depend on the policy type): probably such kind of semantic error is no longer being picked up by latest versions of the tools, such as updated checkpolicy.

I think it's a good thing that the error is triggered and it is better fixed than left there.

The version I have used to trigger such error is checkpolicy 2.6.

pebenito:
Based on the change, it sounds like a compiler bug in 2.6. Refpolicy requires 3.1+ these days. If you feel it is a bug that checkpolicy no longer deems this a compile error, please raise an issue on the SELinux mail list.

IMO it is correct as it is currently written in main.

gtrentalancia:
Yes, in my opinion it's a bug in recent build tools.

It's never a good thing to break backwards-compatibility.

Also consider a template is an interface: how can it be semantically correct policy, if it declares a type instead of stating that it requires it ? And declaring attributes that aren't needed in the context should trigger lint errors.

I am quite busy and I cannot guarantee that I will be able to find the time to report the issue to the SELinux mailing list...

pebenito:
> Also consider a template is an interface: how can it be semantically correct policy, if it declares a type instead of stating that it requires it ? And declaring attributes that aren't needed in the context should trigger lint errors.

By definition a template can create types. It has always been like that. That's the primary difference between an interface and a template.

PR #652, opened 2023-09-06
Increase general syslog daemon policy security by making network permissions tunable

gtrentalancia:
This patchset introduces a new "logging_syslog_can_network" boolean and corresponding tunable policy to isolate the syslog daemon from the network for increased security in most of its use cases.

See also: https://github.com/SELinuxProject/refpolicy/pull/644#issuecomment-1707302056

### DIFF
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -5,6 +5,14 @@ policy_module(logging)
 # Declarations
 #
 
+## <desc>
+## <p>
+## Allows syslogd internet domain sockets
+## functionality (dangerous).
+## </p>
+## </desc>
+gen_tunable(logging_syslog_can_network, false)
+
 attribute logfile;
 
 type auditctl_t;
@@ -386,8 +394,7 @@ optional_policy(`
 # chown fsetid for syslog-ng
 # sys_admin for the integrated klog of syslog-ng and metalog
 # sys_nice for rsyslog
-# cjp: why net_admin!
-allow syslogd_t self:capability { chown dac_override fsetid net_admin setgid setuid sys_admin sys_nice sys_resource sys_tty_config };
+allow syslogd_t self:capability { chown dac_override fsetid setgid setuid sys_admin sys_nice sys_resource sys_tty_config };
 dontaudit syslogd_t self:capability { sys_ptrace };
 dontaudit syslogd_t self:cap_userns { kill sys_ptrace };
 # setpgid for metalog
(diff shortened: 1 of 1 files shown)


policy/modules/system/logging.te
```diff
@@ -5,6 +5,14 @@ policy_module(logging)
 # Declarations
 #
 
+## <desc>
+## <p>
+## Allows syslogd internet domain sockets
+## functionality (dangerous).
+## </p>
+## </desc>
+gen_tunable(syslog_can_network, false)
```
pebenito, line 14: The preference now is that the module name is the first work of the tunable, like it is with interface names. Please name it something like `logging_syslog_networking`.

policy/modules/system/logging.te
```diff
@@ -457,28 +464,32 @@ kernel_read_ring_buffer(syslogd_t)
...
-corenet_tcp_connect_postgresql_port(syslogd_t)
-corenet_tcp_connect_mysqld_port(syslogd_t)
-
-# syslog-ng can send or receive logs
-corenet_sendrecv_syslogd_client_packets(syslogd_t)
-corenet_sendrecv_syslogd_server_packets(syslogd_t)
-corenet_sendrecv_postgresql_client_packets(syslogd_t)
-corenet_sendrecv_mysqld_client_packets(syslogd_t)
+tunable_policy(`syslog_can_network',`
+	allow syslogd_t self:capability { net_admin net_bind_service };
```
pebenito, line 468: You should be able to drop the explicit `net_bind_service`. The corenet interfaces will add it if the ports that it binds to are < 1024.

policy/modules/system/logging.te
```diff
@@ -457,28 +464,32 @@ kernel_read_ring_buffer(syslogd_t)
...
-corenet_tcp_connect_syslogd_port(syslogd_t)
-corenet_tcp_connect_postgresql_port(syslogd_t)
-corenet_tcp_connect_mysqld_port(syslogd_t)
-
-# syslog-ng can send or receive logs
-corenet_sendrecv_syslogd_client_packets(syslogd_t)
-corenet_sendrecv_syslogd_server_packets(syslogd_t)
-corenet_sendrecv_postgresql_client_packets(syslogd_t)
-corenet_sendrecv_mysqld_client_packets(syslogd_t)
+tunable_policy(`syslog_can_network',`
```
pebenito, line 467: This block should go between the `ifdef` and `optional_policy` blocks.

pebenito (changes requested):
I'm fine with the idea, just minor notes.

policy/modules/system/logging.te
gtrentalancia: Good point, I hope the macro works.

policy/modules/system/logging.te
gtrentalancia: Amended.

pebenito (approved):

PR #653, opened 2023-09-06
Add iio-sensor-proxy.

etbe:
iio-sensor-proxy (Debian package iio-sensor-proxy)
 IIO sensors to D-Bus proxy
 Industrial I/O subsystem is intended to provide support for devices
 that in some sense are analog to digital or digital to analog convertors
 .
 Devices that fall into this category are:
  * ADCs
  * Accelerometers
  * Gyros
  * IMUs
  * Capacitance to Digital Converters (CDCs)
  * Pressure Sensors
  * Color, Light and Proximity Sensors
  * Temperature Sensors
  * Magnetometers
  * DACs
  * DDS (Direct Digital Synthesis)
  * PLLs (Phase Locked Loops)
  * Variable/Programmable Gain Amplifiers (VGA, PGA)
  
  Signed-off-by: Russell Coker <russell@coker.com.au>

### DIFF
--- a/policy/modules/services/iiosensorproxy.te
+++ b/policy/modules/services/iiosensorproxy.te
@@ -1,0 +1,70 @@
+policy_module(iiosensorproxy)
+
+########################################
+#
+# iio-sensor-proxy (Debian package iio-sensor-proxy)
+# IIO sensors to D-Bus proxy
+# Industrial I/O subsystem is intended to provide support for devices
+# that in some sense are analog to digital or digital to analog convertors
+# .
+# Devices that fall into this category are:
+#  * ADCs
+#  * Accelerometers
+#  * Gyros
+#  * IMUs
+#  * Capacitance to Digital Converters (CDCs)
+#  * Pressure Sensors
+#  * Color, Light and Proximity Sensors
+#  * Temperature Sensors
+#  * Magnetometers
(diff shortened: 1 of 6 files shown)


policy/modules/services/iiosensorproxy.te
```diff
@@ -0,0 +1,73 @@
...
+
+kernel_read_system_state(iiosensorproxy_t)
+
+dev_read_iio(iiosensorproxy_t)
+
+# for /sys/bus/iio/devices
+dev_read_sysfs(iiosensorproxy_t)
+
+# for writing to current_trigger
+dev_write_sysfs(iiosensorproxy_t)
```
pebenito, line 52: Can you be a little more explicit on what is being accessed? In the future I'd like to try to add more specific labeling in `/sys`, and this info would be useful.

Otherwise looks good to me.

pebenito (changes requested):

policy/modules/services/iiosensorproxy.te
etbe: Is this good enough?

pebenito: Yes, thanks.

pebenito (approved):

PR #663, opened 2023-09-08
Debian /run/motd.d

etbe:


policy/modules/system/authlogin.if
```diff
@@ -137,6 +137,25 @@ interface(`auth_read_pam_motd_dynamic',`
...
+## <summary>
+##	Read files in /run/motd.d directory
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`auth_read_pam_motd_files',`
```
dsugar100, line 150: I recently added the interface 'auth_read_pam_motd_dynamic' which looks very similar to this - see #626 (which looks to be just above this interface).  Probably should combine.

pebenito (changes requested):
Please add perms to `auth_read_pam_motd_dynamic` instead of adding `auth_read_pam_motd_files`.

policy/modules/system/authlogin.if
etbe: ok. I'll do a new pr for this.

PR #664, opened 2023-09-08
Policy for firmware update daemon

etbe:


policy/modules/system/fwupd.te
```diff
@@ -0,0 +1,176 @@
...
+#
+
+# sys_admin is for "FuPluginUefiCapsule  skipping device that failed coldplug: failed to read fw_class"
+allow fwupd_t self:capability { dac_override dac_read_search linux_immutable sys_rawio sys_admin };
+
+allow fwupd_t self:process { signal getsched setsched };
+allow fwupd_t self:fifo_file rw_fifo_file_perms;
+allow fwupd_t self:fd use;
+
+allow fwupd_t self:netlink_kobject_uevent_socket { create getattr setopt bind read };
```
pebenito, line 47: Please use `create_socket_perms`.

policy/modules/system/fwupd.te
```diff
@@ -0,0 +1,176 @@
...
+
+########################################
+#
+# Declarations
+#
+
+attribute_role fwupd_roles;
+type fwupd_t;
+type fwupd_exec_t;
+init_daemon_domain(fwupd_t, fwupd_exec_t)
```
pebenito, line 17: Why does the command line tool run in this domain if it is a daemon? Which part (daemon/cli) does the firmware install?

pebenito (changes requested):

policy/modules/system/fwupd.te
etbe: ok

PR #665, opened 2023-09-11
init and shutdown fixes

gtrentalancia:
Fix the label and interface needed to manage the random seed file (additional entropy source) used during system boot/reboot.

Modify the init and shutdown policy accordingly and also fix some bugs in the latter module.

### DIFF
--- a/policy/modules/system/init.if
+++ b/policy/modules/system/init.if
@@ -1453,6 +1453,30 @@ interface(`init_relabel_var_lib_dirs',`
 
 ########################################
 ## <summary>
+##	Create, read, write, and delete the
+##	pseudorandom number generator seed
+##	file in /var/lib or /var/run
+##	directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`init_manage_random_seed',`
+	gen_require(`
+		type init_random_seed_t;
+	')
+
+	files_rw_var_lib_dirs($1)
+	files_rw_runtime_dirs($1)
+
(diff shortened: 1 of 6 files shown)


policy/modules/system/init.if
```diff
@@ -1451,6 +1451,27 @@ interface(`init_relabel_var_lib_dirs',`
...
+## </param>
+#
+interface(`init_manage_random_seed',`
+	gen_require(`
+		type init_random_seed_t;
+	')
+
+	files_search_var_lib($1)
+	files_search_runtime($1)
+	manage_files_pattern($1, init_random_seed_t, init_random_seed_t)
```
pebenito, line 1472: Considering the labeling in the `.fc`, this seems excessive since there is no `init_random_seed_t` dirs.

pebenito (changes requested):

policy/modules/system/init.if
gtrentalancia: Good point, I have just amended it as suggested ! Thanks.

policy/modules/system/init.if
```diff
@@ -1451,6 +1451,27 @@ interface(`init_relabel_var_lib_dirs',`
...
+##	</summary>
+## </param>
+#
+interface(`init_manage_random_seed',`
+	gen_require(`
+		type init_random_seed_t;
+	')
+
+	files_search_var_lib($1)
+	files_search_runtime($1)
```
pebenito, line 1471: You'll need to add a `files_rw_runtime_dirs()` so it can also `add_name` and `remove_name`.

pebenito (changes requested):

pebenito (approved):

PR #670, opened 2023-09-12
Daemon to monitor memory pressure and notify applications and change …

etbe:
…kernel

OOM settings.

### DIFF
--- a/policy/modules/services/lowmemorymonitor.te
+++ b/policy/modules/services/lowmemorymonitor.te
@@ -17,7 +17,7 @@ init_daemon_domain(low_mem_mon_t, low_mem_mon_exec_t)
 #
 
 allow low_mem_mon_t self:capability ipc_lock;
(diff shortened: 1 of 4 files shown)


policy/modules/services/lowmemorymonitor.te
```diff
@@ -0,0 +1,36 @@
...
+type low_mem_mon_exec_t;
+init_daemon_domain(low_mem_mon_t, low_mem_mon_exec_t)
+
+########################################
+#
+# Local policy
+#
+
+allow low_mem_mon_t self:capability ipc_lock;
+allow low_mem_mon_t self:unix_dgram_socket { create write };
```
pebenito, line 20: Can this perm set upgrade to `create_socket_perms`? It seems lacking without a `sendto`.

etbe: ok

pebenito (approved):

PR #671, opened 2023-09-12
Dbus also creates Unix domain sockets in session mode but has insecure networking code

gtrentalancia:
Session-bus equivalent of the following pull request:

https://github.com/SELinuxProject/refpolicy/pull/667

Also introduce a new boolean to control insecure TCP/IP networking features.

### DIFF
--- a/policy/modules/services/dbus.te
+++ b/policy/modules/services/dbus.te
@@ -11,6 +11,16 @@ gen_require(`
 
 ## <desc>
 ## <p>
+## Determine whether the dbus server
+## can use the network (insecure
+## except than in the case of the
+## loopback interface).
+## </p>
+## </desc>
+gen_tunable(dbus_can_network, false)
+
+## <desc>
+## <p>
 ## Allow dbus-daemon system bus to access /dev/net/tun
 ## which is needed to pass tun/tap device file descriptors
 ## over D-Bus.  This is needed by openvpn3-linux.
@@ -283,7 +293,6 @@ allow session_bus_type self:file rw_inherited_file_perms;
 allow session_bus_type self:fifo_file rw_fifo_file_perms;
 allow session_bus_type self:dbus { send_msg acquire_svc };
 allow session_bus_type self:unix_stream_socket { accept listen };
-allow session_bus_type self:tcp_socket { accept listen };
 allow session_bus_type self:netlink_selinux_socket create_socket_perms;
 
 allow session_bus_type dbusd_etc_t:dir list_dir_perms;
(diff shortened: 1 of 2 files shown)


pebenito:
If you're saying the policy is insecure because it allows networking on the dbus session bus, I don't agree. If you're saying dbus code itself has insecure networking, do you have a reference?

gtrentalancia:
I meant the latter, as explained in the D-Bus specification which has been quoted in the commit.

gtrentalancia:
Here is the D-Bus Specification reference (as already said, it's also quoted in the patch):

https://dbus.freedesktop.org/doc/dbus-specification.html#transports-tcp-sockets

gtrentalancia:
It uses TCP/IP connections that are not authenticated and not encrypted, so unless Kerberos is used (which is not the default) or unless the connection is over a loopback interface or private network, it's extremely insecure.

Even if it the connection is over a loopback interface or private network, it would still be insecure to a good degree, due to the lack of encryption and authentication, thus exposing it to man-in-the-middle-attacks from within the local system or private network.

dsugar100:
Is this boolean blocking network accessf for loopback AND off box, or just network connections off box?  It looks like it is blocking even over loopback which seems like should be left enabled by default.  And, some of this network access is also blcoked by firewall rules.

0xC0ncord:
> It looks like it is blocking even over loopback which seems like should be left enabled by default. And, some of this network access is also blcoked by firewall rules.

Seems like loopback access should be allowed by default according to the referenced documentation.

> And, some of this network access is also blcoked by firewall rules.

From a defense in depth perspective I think as long as loopback access is still allowed this would be a valuable policy change.

gtrentalancia:
The policy does not distinguish between packets from different interfaces, they are all considered TCP/IP packets.

So, it blocks networking on the loopback interface too.

But nobody would even think about using the loopback interface when Unix domain sockets can be used: the former would just imply insecurity and a lot of overhead, thus affecting the speed enourmously !

The loopback interface should not be allowed by default. All D-Bus needs on a secure system is Unix domain sockets.

policy/modules/services/dbus.te
```diff
@@ -205,7 +215,7 @@ ifdef(`init_systemd', `
 	')
 ')
 
-tunable_policy(`dbus_pass_tuntap_fd',`
+tunable_policy(`dbus_pass_tuntap_fd && dbus_can_network',`
```
pebenito, line 218: Even if dbus has no networking, it may still need to pass a tun device, and this access is required.

policy/modules/services/dbus.te
```diff
@@ -312,13 +321,17 @@ corecmd_read_bin_files(session_bus_type)
...
+	corenet_all_recvfrom_netlabel(session_bus_type)
+	corenet_tcp_sendrecv_generic_if(session_bus_type)
+	corenet_tcp_sendrecv_generic_node(session_bus_type)
+	corenet_tcp_bind_generic_node(session_bus_type)
 
-corenet_sendrecv_all_server_packets(session_bus_type)
-corenet_tcp_bind_reserved_port(session_bus_type)
+	corenet_sendrecv_all_server_packets(session_bus_type)
+	corenet_tcp_bind_reserved_port(session_bus_type)
+')
```
pebenito, line 334: This should move down above the optional block.

pebenito (changes requested):

policy/modules/services/dbus.te
gtrentalancia: Amended

policy/modules/services/dbus.te
gtrentalancia: Amended

pebenito (approved):

PR #672, opened 2023-09-12
Remote X11 TCP/IP functionality is generally insecure: switch it off by default. Strengthen XDM authentication file access.

gtrentalancia:
Disable X11 server TCP/IP networking functionality by default, as it is generally insecure.

### DIFF
--- a/policy/modules/services/xserver.te
+++ b/policy/modules/services/xserver.te
@@ -179,6 +179,9 @@ files_lock_file(xdm_lock_t)
 type xdm_rw_etc_t;
 files_type(xdm_rw_etc_t)
 
+type xdm_auth_t;
+files_type(xdm_auth_t)
+
 type xdm_var_lib_t;
 files_type(xdm_var_lib_t)
 
@@ -356,6 +359,11 @@ manage_lnk_files_pattern(xdm_t, xdm_tmpfs_t, xdm_tmpfs_t)
 manage_fifo_files_pattern(xdm_t, xdm_tmpfs_t, xdm_tmpfs_t)
 manage_sock_files_pattern(xdm_t, xdm_tmpfs_t, xdm_tmpfs_t)
 fs_tmpfs_filetrans(xdm_t, xdm_tmpfs_t, { dir file lnk_file sock_file fifo_file })
+
+# this file should be accessed only by xdm_t (rw) and xserver_t (ro)
+manage_dirs_pattern(xdm_t, xdm_auth_t, xdm_auth_t)
+manage_files_pattern(xdm_t, xdm_auth_t, xdm_auth_t)
+xserver_xdm_auth_filetrans(xdm_t, xdm_auth_t, file)
 
 manage_dirs_pattern(xdm_t, xdm_var_lib_t, xdm_var_lib_t)
 manage_files_pattern(xdm_t, xdm_var_lib_t, xdm_var_lib_t)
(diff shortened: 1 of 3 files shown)


pebenito:
What situation would you want networking on only one of xdm or xserver?

gtrentalancia:
To be honest, I didn't check carefully and went by memory...

I suppose a machine can act as an XDM server for remote clients (provide graphical login services) without providing the actual X server.

Similar a client can use a local XDM and a remote X server.

But you believe this is important I can check more carefully.

gtrentalancia:
[begin of xdm.1 man page]

Xdm  manages a collection of X displays, which may be on the local host or remote servers...

[end of xdm.1 man page]

So, the machine can use a local XDM to connect to a remote X server. In that case the local X server does not need network functionality, while XDM does.

On the other hand the machine can host an X server used by remote clients and no XDM because it does not offer local graphical logins.

gtrentalancia:
I'll try to rephrase in another way, which is probably simpler to understand: an XDM needs networking only if it offers connections to remote X servers, while an X server needs networking only if acts as an X server for remote X clients.

Does it make sense ?

policy/modules/services/xserver.te
```diff
@@ -32,6 +32,22 @@ gen_require(`
...
+## </desc>
+gen_tunable(xserver_can_network, false)
+
+## <desc>
+## <p>
+## Allows the X display manager to use
+## TCP/IP networking functionality (insecure).
+## </p>
+## </desc>
+gen_tunable(xdm_can_network, false)
```
pebenito, line 49: Pleasee rename to `xserver_xdm_can_network`.

policy/modules/services/xserver.te
```diff
@@ -396,17 +410,22 @@ kernel_view_key(xdm_t)
...
-corenet_udp_sendrecv_generic_if(xdm_t)
-corenet_tcp_sendrecv_generic_node(xdm_t)
-corenet_udp_sendrecv_generic_node(xdm_t)
-corenet_tcp_bind_generic_node(xdm_t)
-corenet_udp_bind_generic_node(xdm_t)
-corenet_tcp_connect_all_ports(xdm_t)
-corenet_sendrecv_all_client_packets(xdm_t)
-# xdm tries to bind to biff_port_t
-corenet_dontaudit_tcp_bind_all_ports(xdm_t)
+tunable_policy(`xdm_can_network',`
```
pebenito, line 413: This block needs to move down.

policy/modules/services/xserver.te
```diff
@@ -705,16 +722,24 @@ kernel_write_proc_files(xserver_t)
...
-corenet_tcp_sendrecv_generic_if(xserver_t)
-corenet_udp_sendrecv_generic_if(xserver_t)
-corenet_tcp_sendrecv_generic_node(xserver_t)
-corenet_udp_sendrecv_generic_node(xserver_t)
-corenet_tcp_bind_generic_node(xserver_t)
-corenet_tcp_bind_xserver_port(xserver_t)
-corenet_tcp_connect_all_ports(xserver_t)
-corenet_sendrecv_xserver_server_packets(xserver_t)
-corenet_sendrecv_all_client_packets(xserver_t)
+tunable_policy(`xserver_can_network',`
```
pebenito, line 725: This block also needs to move down.

pebenito (changes requested):

gtrentalancia:
For many workstations (possibly not all of them) in many companies it's probably relatively safe to use over Virtual Private Networks, IPSEC or local networks.

In any case, I have now added a new patch to strengthen XDMCP security for local servers (no networking case) by creating a new file label for the authorization data file and making it exclusively accessible by xserver_t and xdm_t.

I hope it helps.

policy/modules/services/xserver.te
gtrentalancia: Renamed

policy/modules/services/xserver.te
gtrentalancia: I moved it down

policy/modules/services/xserver.te
gtrentalancia: I moved it down

pebenito (approved):

PR #674, opened 2023-09-12
separate label for /etc/security/opasswd

dsugar100:
Seting /etc/security/opasswd to shadow_t has some negative side effects like the fact that pam_unix needs to read that.  Once pam_unix can read shadow_t that changes the behavour of how pam_unix uses unix_update to update the password.  So, this change defines the new type, shadow_history_t, for /etc/secuirty/opasswd.

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -75,6 +75,9 @@ neverallow ~can_relabelto_shadow_passwords shadow_t:file relabelto;
 type shadow_lock_t;
 files_lock_file(shadow_lock_t)
 
+type shadow_history_t;
+files_auth_file(shadow_history_t)
+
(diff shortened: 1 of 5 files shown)


policy/modules/system/authlogin.te
```diff
@@ -391,7 +398,6 @@ files_manage_etc_files(updpwd_t)
 term_dontaudit_use_console(updpwd_t)
 term_dontaudit_use_unallocated_ttys(updpwd_t)
 
-auth_etc_filetrans_shadow(updpwd_t)
```
pebenito, line 394: This makes me nervous. How sure are you this is correct, notwithstanding the desire to get opasswd to the right context? Otherwise, this makes sense.

PR #676, opened 2023-09-13
Allow all users to send syslog messages

dsugar100:
Aug 29 12:53:06 localhost.localdomain audisp-syslog[1550]: node=localhost type=AVC msg=audit(1693313586.678:437): avc:  denied  { write } for  pid=1757 comm="systemctl" name="socket" dev="tmpfs" ino=58 scontext=user_u:user_r:user_t:s0 tcontext=system_u:object_r:devlog_t:s0 tclass=sock_file permissive=1
Aug 29 12:53:06 localhost.localdomain audisp-syslog[1550]: node=localhost type=AVC msg=audit(1693313586.678:437): avc:  denied  { sendto } for  pid=1757 comm="systemctl" path="/run/systemd/journal/socket" scontext=user_u:user_r:user_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=unix_dgram_socket permissive=1
Aug 29 13:10:01 localhost.localdomain audisp-syslog[1545]: node=localhost type=AVC msg=audit(1693314601.860:435): avc:  denied  { write } for  pid=1756 comm="systemctl" name="socket" dev="tmpfs" ino=58 scontext=user_u:user_r:user_t:s0 tcontext=system_u:object_r:devlog_t:s0 tclass=sock_file permissive=1
Aug 29 13:10:01 localhost.localdomain audisp-syslog[1545]: node=localhost type=AVC msg=audit(1693314601.860:435): avc:  denied  { sendto } for  pid=1756 comm="systemctl" path="/run/systemd/journal/socket" scontext=user_u:user_r:user_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=unix_dgram_socket permissive=1

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -139,6 +139,10 @@ template(`userdom_base_user_template',`
 		allow $1_t self:process execstack;
 	')
 
+	tunable_policy(`user_all_users_send_syslog',`
+		logging_send_syslog_msg($1_t)
+	')
+
 	optional_policy(`
 		devicekit_dbus_chat_disk($1_t)
 		devicekit_dbus_chat_power($1_t)
--- a/policy/modules/system/userdomain.te
+++ b/policy/modules/system/userdomain.te
@@ -21,6 +21,13 @@ gen_tunable(allow_user_postgresql_connect, false)
 
 ## <desc>
 ## <p>
+## Allow all users to send syslog messages
+## </p>
+## </desc>
+gen_tunable(user_all_users_send_syslog, true)
+
+## <desc>
+## <p>
 ## Allow regular users direct mouse access
 ## </p>
 ## </desc>


gtrentalancia:
What is the need for this ?

It can be quite dangerous: usually only processes started by administrative users should be able to write syslog !a

Consider the system log is only readable (and writable) by root: so why should an unprivileged user be able to write to it by sending messages to the syslog daemon ??

gtrentalancia:
An unprivileged user can send fake messages to syslog and trick the administrator to perform actions that are detrimental to the local system security and integrity.

dsugar100:
Upon login to a graphical interface systemctl is setting syslog messages (as the user).   This could contain errors about some service that is failing.

gtrentalancia:
The user domain (i.e. unprivileged users) should not be able to send syslog messages for aforesaid security considerations.

gtrentalancia:
Please see: https://github.com/SELinuxProject/refpolicy/pull/680

pebenito:
> Upon login to a graphical interface systemctl is setting syslog messages (as the user). This could contain errors about some service that is failing.

Setting or sending? It shouldn't be able to set parameters; I would think that would be denied by cap_syslog.

> The user domain (i.e. unprivileged users) should not be able to send syslog messages for aforesaid security considerations.

I agree with this in principle but need to understand the situation better.

0xC0ncord:
I'll add I've also seen this behavior but have hesitated to allow the access for the aforementioned security reasons.

gtrentalancia:
It completely defeats the purpose of SELinux !

0xC0ncord:
Would it be feasible/a better idea to give `systemctl` its own domains for the users with this access instead? e.g. `user_systemctl_t`

gtrentalancia:
It would not change anything, as such domain would be reached using a domain transition from user_t upon executing **systemctl**.

gtrentalancia:
It's not like a whiteboard that everybody can write !

Look at the permissions of the files:

-rw-------. 1 root root 6744080 14 sep 17.14 /var/log/syslog

It's read-write only by root, therefore allowing users to write to it by exploiting the syslog daemon would constitute a violation of DAC permissions in the first place !

dburgener:
> It's read-write only by root, therefore allowing users to write to it by exploiting the syslog daemon would constitute a violation of DAC permissions in the first place !

SELinux is enforced after DAC, so a user needs to have root in order to even reach the SELinux check in this case.  SELinux only does additional denying past DAC, not allowing a bypass of DAC.

0xC0ncord:
> It would not change anything, as such domain would be reached using a domain transition from user_t upon executing **systemctl**.

Such a change would allow this functionality to work while keeping privileges separated from the generic user domains. At the very least this could be made conditional via a tunable instead of leaving this functionality of `systemctl` in a broken state.

dsugar100:
> It's not like a whiteboard that everybody can write !
> 
> Look at the permissions of the files:
> 
> -rw-------. 1 root root 6744080 14 sep 17.14 /var/log/syslog
> 
> It's read-write only by root, therefore allowing users to write to it by exploiting the syslog daemon would constitute a violation of DAC permissions in the first place !

The logging_send_syslog_msg inteface isn't allowing write access to /var/log/syslog, it is allowing write to the message queue that sends messages to the logging system (i.e. /dev/log)

gtrentalancia:
It's not a broken state, it's just forbidden dangerous functionality, i.e. security bugs in the application code that are showing up thanks to SELinux !

If the _user_ domain can transition to such _ user_systemctl_t_ domain, then the situation would be equivalent to the _user_ getting an alias and that is completely pointless in terms of security, as it would not prevent _user_ domain for sending syslog messages.

gtrentalancia:
I think applications should have their own log files in the home directories.

So, for example, such kind of logging from _user_ using the graphical interface should end up in the ~/.xsession-errors file.

0xC0ncord:
>It's not a broken state, it's just forbidden dangerous functionality, i.e. security bugs in the application code that are showing up thanks to SELinux !

I agree, but if `systemctl` is sending logs in this way *by design* then I do not believe it is the policy's job to govern how the application should behave.

0xC0ncord:
> Most permissions that it requires are not even enclosed in an "ifdef" statement, so they spread to the policy of all systems, even those not using it.
> 
> For example, I don't use it for this precise reason !

If there are other parts of the policy that are strictly systemd-specific that aren't enclosed in `ifdef`s for systemd, it's my opinion that's a policy bug.

For something like this access I think it could go either way. Yes, this appears to be systemd-specific functionality but a user with such a system may still want to disable this. In that case I think a tunable would be more appropriate.

> Not allowing that permission does not prevent the rest of the system from running, it just prevents spamming the system log with messages (that can be forged by the user) [...]

I don't disagree, but if the policy outright disallows this access it has the effect of disallowing logging for legitimate applications that need it. Whether or not user applications should all just stop using the systemd journal and go back to using regular log files is a matter of opinion, and again I think a tunable is the best compromise.

dburgener:
+1 for a tunable.

This is a tricky question that I see both sides of.  I think different use cases and users will have different thoughts about the tradeoff between "my applications work as designed" and "This is a security concern", so giving refpolicy users the option seems like the right way to go.

I might vote for a default setting of allowing the access, since I think security conscious users who care about this are the ones more likely to investigate, find and enable the setting, but I don't feel strongly either way.

gtrentalancia:
If it was preventing applications from running, then there were plenty of bug reports in every distribution !

So, in my opinion is should be just _dontaudited_.

As already said, _user_ applications should produce logs in the user home directory or use .xsession-errors (_xserver_rw_xsession_log_()).

And even if you decide to use tunable policy, it should be enclosed in an "ifdef" statement because not everybody uses _systemd_.

gtrentalancia:
In a few words, I think the aim of advanced SELinux use is not that of allowing anything that a software package can do by design of its authors, but instead allowing the minimum set of functionality which guarantees most of the features required by the average security savvy user, while not affecting system security or privacy.

0xC0ncord:
> In a few words, I think the aim of advanced SELinux use is not that of allowing anything that a software package can do by design of its author(s), but instead that of allowing the minimum set of functionality which guarantees most of the features required by the average security savvy user, while not affecting system security or privacy.

While I admire the goal of absolute minimum required functionality, I think actively fighting against softwares' design is only going to hinder refpolicy's ability to be widely deployable on many SELinux-enabled systems. Yes, the policy should only allow what is necessary but "what is necessary" can mean something different depending on the user and their needs.

I think a significantly stricter policy that has the side effect of breaking some functionality, even if that functionality can be lived without, it will be a big enough annoyance for some people that makes refpolicy non-approachable for them, especially if it's non-configurable.

gtrentalancia:
SELinux audit (denial) messages are like a **radar**, they just detect threats, they do not fight anything.

But obviously, if such **radar** is switched off by allowing all permissions that pop up in the audit log, then SELinux is no longer able to protect the system as it should and the policy just degrades and becomes weaker and weaker with time.

I agree that tunable policy is a compromise in many scenarios though...

dburgener:
I think we need to make sure to differentiate between the goals of *SELinux* vs the goals of *refpolicy*.

The goal of SELinux is to provide a flexible mechanism to enforce security policies.  This means that users who want a very strict locked down system can get that, and other users who are more concerned with "just make it work" can get that too.The goals of "SELinux" are to allow policy authors to make decisions about things like users sending syslog message.  SELinux is agnostic to your security goals, it just gives you tools to meet them.

The goals of refpolicy are documented here: https://github.com/SELinuxProject/refpolicy/wiki#project-goals.  This excerpt from the first paragraph there is very relevant:

> In reality, different systems have different requirements and purposes and corresponding differences in the meaning of secure. What is a fundamental security flaw on one system might be acceptable, or even the primary functionality, of another. The challenge for a system policy is to support as many of these differing security goals as is practical.

The "Configurability" goal listed there seems to me the most relevant bit for this.  If the policy does allow all users to send syslog messages, that seems in violation of having a "Flexible base policy" that would support users with security goals like Guido's, since it's tricky for him to remove those rules on systems he cares about.  On the other hand Dave is frustrated by missing log messages on his system, and would like to allow them despite the security concerns.  Adding a tunable for this case makes refpolicy flexibly provide for both cases.

gtrentalancia:
Thanks for quoting the project goals !

I am happy either way, as I am using a different policy and just trying to contribute sometime and learn about new software.

Given the stated goals, then tunable policy enclosed by "ifdef" seem the right way to go.

pebenito:
> Given the stated goals, then tunable policy enclosed by "ifdef" seems the right way to go.

I agree.

dsugar100:
Ok, I think I made the changes requested.  Lots of conversation on this one!  
If I got something wrong, please comment.

0xC0ncord:
PR looks good to me!

policy/modules/system/userdomain.if
```diff
@@ -124,6 +124,12 @@ template(`userdom_base_user_template',`
 	# kdeinit wants systemd status
 	init_get_system_status($1_t)
 
+	ifdef(`init_systemd',`
```
pebenito, line 127: I don't think this should be in a systemd block. It has potential use outside of systemd. Otherwise looks good.

dsugar100: done

pebenito (approved):

PR #679, opened 2023-09-13
Improve a previous syslog tunable policy change

gtrentalancia:
Improve a previous syslog tunable policy change.

### DIFF
--- a/policy/modules/system/logging.te
+++ b/policy/modules/system/logging.te
@@ -408,8 +408,6 @@ allow syslogd_t self:unix_dgram_socket create_socket_perms;
 allow syslogd_t self:unix_stream_socket create_stream_socket_perms;
 allow syslogd_t self:unix_dgram_socket sendto;
 allow syslogd_t self:fifo_file rw_fifo_file_perms;
-allow syslogd_t self:udp_socket create_socket_perms;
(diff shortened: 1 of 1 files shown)


dsugar100:
Silly question - is it worth splitting this (and maybe the similar syslog tunable) into two?  outbound network vs. inbound network?  So one could enable policy for audit or syslog to only send syslog/audit off box, but not receive.  Or to receive only but not send off box?  Maybe it will make for too many tunables, but just thinking about how these types of things are typically used.  In or out.

gtrentalancia:
Receiving logs from remote machines is not dangerous for the machine that receives them (server), as the data that transits over the network can only disclose information of remote machines (clients).

It's the responsibility of the client machines to disable networking on their end !

So, I believe splitting in two gives no benefit.

gtrentalancia:
In my opinion, the issue with adding too many booleans is not much related to the increased number of them, but instead related to the policy becoming more difficult to configure, as it would take much more time for system administrators to understand how to properly use them.

dsugar100:
Sounds good, I agree, I'm not sure it is worth it.  Just a thought I had when I saw your merge request.

pebenito:
The explicit purpose of `audisp-remote` is to log to a remote server. I disagree with this change.

pebenito (approved):

PR #680, opened 2023-09-13
Fix a security bug in the system logging permissions

gtrentalancia:
Fix a security bug which allows unprivileged users to send messages (including fake/forged ones) to the syslog in order to trick the administrator into performing actions detrimental to the system security.

Originally introduced here:

   https://github.com/SELinuxProject/refpolicy/commit/847937da7ddd093893b68d5386b81d60def4b263
   Date:   Tue Nov 13 19:31:43 2007 +0000

 policy/modules/system/userdomain.if |    7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)


pebenito:
Is this still relevant now that #676 is merged?

PR #685, opened 2023-09-14
Remove the syslog capability from the unconfined domain

gtrentalancia:
This bug allows the exploitation of unconfined domains to send fake or forged messages that trick the administrator to perform actions detrimental to the system security.

Originally introduced here:

  https://github.com/SELinuxProject/refpolicy/commit/4067a18530579ba33f7a8cee307ae2be83f6b740
  Date:   Mon Jun 9 14:38:45 2014 +0200

 policy/modules/system/unconfined.if |    2 +-

### DIFF
--- a/policy/modules/system/unconfined.if
+++ b/policy/modules/system/unconfined.if
@@ -38,7 +38,7 @@ interface(`unconfined_domain_noaudit',`
 
 	# Use most Linux capabilities
 	allow $1 self:{ capability cap_userns } { chown dac_override dac_read_search fowner fsetid kill setgid setuid setpcap linux_immutable net_bind_service net_broadcast net_admin net_raw ipc_lock ipc_owner sys_rawio sys_chroot sys_ptrace sys_pacct sys_admin sys_boot sys_nice sys_resource sys_time sys_tty_config mknod lease audit_write audit_control setfcap };
-	allow $1 self:{ capability2 cap2_userns } { syslog wake_alarm bpf perfmon };
+	allow $1 self:{ capability2 cap2_userns } { wake_alarm bpf perfmon };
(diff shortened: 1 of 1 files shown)


dburgener:
The purpose of the unconfined domain is to be unconfined... It's a violation of user expectations to restrict nearly any access here.  If users don't want to allow a domain this access, they shouldn't make it unconfined.

gtrentalancia:
Unconfined should allow most permissions, but not those that are critical to the inner security of the system, such as in this case the ability to undermine the integrity of system logs.

dburgener:
> Please consider that such permission has been probably introduced due to a distribution bug, instead of enclosing it in a distribution "ifdef" statement.

I think the permission was introduced because unconfined is supposed to be unconfined, so the authors enumerated all the capabilities.

> Unconfined should allow most permissions, but not those that are critical to the inner security of the system, such as in this case the ability to undermine the integrity of system logs.

Then your gripe should be with the `dev_unconfined()` call below this which allows it to write directly to raw disks and bypass all the access controls.  An malicious process in an unconfined domain doesn't need to trick an admin into doing something malicious, it can just do it itself.

The idea behind "unconfined" is that it's (effectively) not confined by SELinux.

pebenito:
This is not a bug. This is intended as noted by dburgener. The real remedy for your concern would be to remove unconfined domains from the policy.

gtrentalancia:
If you look at the original commit that introduced such permission, it says that it has been introduced to run dmesg.

Now, I can run dmesg on my system (provided that the "user_dmesg" is set to "true") without the need for such patch.

So, the above means that the permission is not needed and only exposes the system logs to integrity violations.

PR #687, opened 2023-09-15
Clean the modular policy packages before installing

gtrentalancia:
Before installing the policy modules packages in a modular policy, clean all installed modules as they are obsolete and they might have been disabled.
    
This leaves a clean, fresh modular policy packages directory after "make install".

 Rules.modular |    6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

### DIFF
--- a/Rules.modular
+++ b/Rules.modular
@@ -45,7 +45,11 @@ base: $(base_pkg)
 
 modules: $(mod_pkgs)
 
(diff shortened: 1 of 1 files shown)


pebenito:
I don't want the break the use case where someone puts their own module in that dir..

PR #689, opened 2023-09-19
debian motd.d directory

etbe:


policy/modules/services/ssh.te
```diff
@@ -261,6 +261,9 @@ corenet_sendrecv_xserver_server_packets(sshd_t)
 ifdef(`distro_debian',`
 	allow sshd_t self:process { getcap setcap };
 	auth_use_pam_motd_dynamic(sshd_t)
+
+	# for unattended-upgrades notices.  Should we have a domain transition?
+	files_read_var_lib_files(sshd_t)
```
pebenito, line 266: I'd prefer to have no `var_lib_t` files if possible.

etbe: Tests indicate this isn't needed any more, I think changes related to apt and unattended-upgrades addressed the issue.

pebenito (approved):

PR #690, opened 2023-09-19
policy for the Reliability Availability servicability daemon

etbe:


policy/modules/kernel/filesystem.if
```diff
@@ -6136,6 +6136,43 @@ interface(`fs_getattr_tracefs_files',`
...
+## </param>
+#
+interface(`fs_write_tracefs_files',`
+	gen_require(`
+		type tracefs_t;
+	')
+
+	allow $1 tracefs_t:dir list_dir_perms;
+	allow $1 tracefs_t:file rw_file_perms;
+')
```
pebenito, line 6156: A bit of inconsistency. Since everything but the interface name is `rw`, I assume the interface name should be `rw` instead.

policy/modules/services/rasdaemon.te
```diff
@@ -0,0 +1,41 @@
+policy_module(rasdaemon, 1.0.0)
```
pebenito, line 1: Please drop version.

pebenito (changes requested):

policy/modules/services/rasdaemon.te
etbe: ok

PR #692, opened 2023-09-19
systemd-locale write to init_t runtime socket

dsugar100:
Sep 13 19:20:47 localhost.localdomain audisp-syslog[1531]: node=localhost type=AVC msg=audit(1694632847.823:567): avc:  denied  { write } for  pid=1807 comm="systemd-localed" name="notify" dev="tmpfs" ino=47 scontext=system_u:system_r:systemd_locale_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=sock_file permissive=1
Sep 13 19:21:21 localhost.localdomain audisp-syslog[1531]: node=localhost type=AVC msg=audit(1694632881.720:696): avc:  denied  { write } for  pid=1807 comm="systemd-localed" name="notify" dev="tmpfs" ino=47 scontext=system_u:system_r:systemd_locale_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=sock_file permissive=1
Sep 13 19:22:05 localhost.localdomain audisp-syslog[1531]: node=localhost type=AVC msg=audit(1694632925.429:22455): avc:  denied  { write } for  pid=2643 comm="systemd-localed" name="notify" dev="tmpfs" ino=47 scontext=system_u:system_r:systemd_locale_t:s0 tcontext=system_u:object_r:init_runtime_t:s0 tclass=sock_file permissive=1

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -776,6 +776,8 @@ kernel_read_kernel_sysctls(systemd_locale_t)
 
 files_read_etc_files(systemd_locale_t)
 
+init_write_runtime_socket(systemd_locale_t)
+
 selinux_use_status_page(systemd_locale_t)
 
 seutil_read_file_contexts(systemd_locale_t)


pebenito:
> I think this access is fine, but I wonder if it would be better to blanket this access across any `init_daemon_domain()` when `ifdef('init_systemd')` as this would cover all cases where a daemon uses `systemd-notify`. I worry that might be too coarse though as not every service supports it.

I would be fine with that if we change the label of the socket so it is different than /run/systemd/private.

dsugar100:
I think you are suggesting a new type for the socket /run/systemd/notify?  
Is this created by systemd-notify?  I don't see permission or a type transition from systemd_notify_t to create it.  
Is it maybe done by systemd (init_t domain)?

pebenito:
Yes, I mean a new type. It's created by pid 1/init_t.

dsugar100:
Closing - implemented with #710

PR #697, opened 2023-09-20
mon.te patches as well as some fstools patches related to it

etbe:


policy/modules/kernel/files.if
```diff
@@ -434,6 +434,24 @@ interface(`files_tmpfs_file',`
...
+## <summary>
+##	dontaudit getattr on tmpfs files
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not have stat on tmpfs files audited
+##	</summary>
+## </param>
+#
+interface(`files_dontaudit_tmpfs_file_getattr',`
```
pebenito, line 447: Please rename to `files_dontaudi_getattr_all_tmpfs_files`.

policy/modules/kernel/storage.if
```diff
@@ -310,6 +310,30 @@ interface(`storage_dev_filetrans_fixed_disk',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="filename" optional="true">
+##	<summary>
+##	Optional filename of the char device to be created
+##	</summary>
+## </param>
+#
+interface(`storage_dev_filetrans_fixed_disk_control',`
```
pebenito, line 329: Please remove `_control` from the name and make the obj class a parameter.

policy/modules/services/mon.te
```diff
@@ -161,9 +169,10 @@ optional_policy(`
 #
 
 # sys_ptrace is for reading /proc/1/maps etc
-allow mon_local_test_t self:capability { sys_ptrace sys_admin };
+allow mon_local_test_t self:capability { dac_override dac_read_search setgid setuid sys_rawio sys_ptrace sys_admin };
```
pebenito, line 172: The `storage_raw_read_fixed_disk` should have the `sys_rawio` with it, so I'd rather keep it there and drop it here, unless there is another reason to need the perm.

policy/modules/system/fstools.te
```diff
@@ -145,6 +150,8 @@ mls_file_write_all_levels(fsadm_t)
 
 selinux_getattr_fs(fsadm_t)
 
+storage_dev_filetrans_fixed_disk_control(fsadm_t, "megaraid_sas_ioctl_node")
```
pebenito, line 153: I'd rather drop the name unless there is a conflicting type_transition that I'm not seeing.

pebenito (approved):

pebenito (changes requested):
accidentally hit approve on above comments instead of request changes.

policy/modules/kernel/files.if
etbe: done

policy/modules/kernel/storage.if
etbe: There was already an interface without the _control. Following the example of dev_filetrans_cardmgr() I made it take both chr_file and blk_file,

policy/modules/services/mon.te
etbe: done

policy/modules/system/fstools.te
etbe: ok

etbe: libsepol.expand_terule_helper: conflicting TE rule for (kernel_t, device_t:chr_file):  old was event_device_t, new is fixed_disk_device_t
libsepol.expand_module: Error during expand

Seems to be a conflicting transition

etbe: dev_filetrans_input_dev(kernel_t)
storage_dev_filetrans_fixed_disk(kernel_t)

The existence of the above 2 lines precludes having storage_dev_filetrans_fixed_disk() work on both chr_file and blk_file. Shall I replace storage_dev_filetrans_fixed_disk with 2 new interfaces and if so what shall I name them?

pebenito: In that case, I propose: try adding `chr_file` to `storage_dev_filetrans_fixed_disk()` and then keep the `"megaraid_sas_ioctl_node"` in the call.

etbe: ok

PR #698, opened 2023-09-20
policy patches for anti-spam daemons

etbe:


policy/modules/services/spamassassin.te
```diff
@@ -419,6 +419,10 @@ tunable_policy(`rspamd_spamd',`
 	allow spamd_t self:process setrlimit;
 	allow spamc_t self:process setrlimit;
 
+	allow spamd_t self:process execmem;
```
dsugar100, line 422: does it really need execmem?

etbe: # ldd /usr/bin/rspamd|grep lua
        libluajit-5.1.so.2 => /lib/x86_64-linux-gnu/libluajit-5.1.so.2 (0x00007f809b3db000)
# dpkg -s rspamd|grep Depend
Depends: adduser, ca-certificates, fonts-glyphicons-halflings, libjs-bootstrap4, libjs-jquery, libjs-requirejs, perl:any, libc6 (>= 2.34), libgcc-s1 (>= 3.4), libglib2.0-0 (>= 2.70.0), libhyperscan5 (>= 5.4.0), libicu72 (>= 72.1~rc-1~), libluajit-5.1-2 (>= 2.0.4+dfsg) | libluajit2-5.1-2 (>= 2.1~), libpcre2-8-0 (>= 10.22), libsodium23 (>= 0.6.1), libsqlite3-0 (>= 3.7.15), libssl3 (>= 3.0.0), libstdc++6 (>= 11), libunwind8, zlib1g (>= 1:1.2.0)
Pre-Depends: init-system-helpers (>= 1.54~)

The Debian rspamd package is linked against luajit. For Debian users there is no realistic option to turn it off. Are there other distributions with different link options for rspam?

0xC0ncord: On Gentoo you can manually switch between interpreters, whether that be LuaJIT or other Lua versions. The system I am personally using rspamd is using Lua 5.4.

I think we just need to put this specific access in its own tunable with a suitable description.

0xC0ncord: > I think we just need to put this specific access in its own tunable with a suitable description.

I'm not sure if refpolicy allows this, but it would be nice if it were defaulted to enabled when `ifdef(distro_debian)`.

etbe: I created a separate tunable for this. I put it outside the main rspam part because it is possible to have SpamAssassin configurations use execmem.

pebenito (approved):

PR #701, opened 2023-09-21
misc small patches for cron policy

etbe:


policy/modules/services/cron.te
```diff
@@ -422,6 +422,7 @@ optional_policy(`
 
 optional_policy(`
 	systemd_dbus_chat_logind(crond_t)
+	systemd_dontaudit_connect_machined(crond_t)
```
pebenito, line 425: I'm curious, why is it trying to connect?

policy/modules/services/cron.te
```diff
@@ -499,6 +503,7 @@ corenet_udp_sendrecv_generic_if(system_cronjob_t)
 corenet_tcp_sendrecv_generic_node(system_cronjob_t)
 corenet_udp_sendrecv_generic_node(system_cronjob_t)
 corenet_udp_bind_generic_node(system_cronjob_t)
+corenet_tcp_connect_tor_port(system_cronjob_t)
```
pebenito, line 506: This seems to be out of place as it seems to be a use case specific to you, or at least, a very rare one.

policy/modules/services/mta.te
```diff
@@ -285,7 +285,12 @@ optional_policy(`
 	userdom_dontaudit_use_user_ptys(system_mail_t)
 
 	optional_policy(`
+ifdef(`hide_broken_symptoms',`
```
pebenito, line 288: Please change this to `distro_debian`. I'd rather not add "broken_symptoms" blocks, but instead elimitate them since they are essentially always enabled.

pebenito (changes requested):

policy/modules/services/cron.te
etbe: ok

policy/modules/services/cron.te
etbe: I don't know either. I'll remove it now and maybe submit in a different PR

policy/modules/services/mta.te
etbe: ok

pebenito (approved):

PR #702, opened 2023-09-22
small postgresql and mysql stuff

etbe:


policy/modules/services/postgresql.te
```diff
@@ -282,7 +285,10 @@ manage_lnk_files_pattern(postgresql_t, postgresql_tmp_t, postgresql_tmp_t)
 manage_fifo_files_pattern(postgresql_t, postgresql_tmp_t, postgresql_tmp_t)
 manage_sock_files_pattern(postgresql_t, postgresql_tmp_t, postgresql_tmp_t)
 files_tmp_filetrans(postgresql_t, postgresql_tmp_t, { dir file sock_file })
-fs_tmpfs_filetrans(postgresql_t, postgresql_tmp_t, { dir file lnk_file sock_file fifo_file })
+fs_tmpfs_filetrans(postgresql_t, postgresql_tmp_t, { dir lnk_file sock_file fifo_file })
```
pebenito, line 288: Can you explain this change? Arguably the entire line (and related rules above) should change to `_tmp_t`.

etbe: A _tmpfs_t type is appropriate if the file is being created specifically as a shared memory file (EG under /dev/shm).

If a tmpfs is just being used in the manner of a regular /tmp then a _tmp_t type seems OK to me, but if you want it all changed to _tmpfs_t then I can do that.

etbe: # ls -lZ /dev/shm
total 1052
-rw-------. 1 postgres postgres system_u:object_r:postgresql_tmpfs_t:s0   26976 Sep 11 08:13 PostgreSQL.1859957116
-rw-------. 1 postgres postgres system_u:object_r:postgresql_tmpfs_t:s0 1048576 Sep 11 08:13 PostgreSQL.4034646664

Looks like it.

pebenito (approved):

PR #703, opened 2023-09-22
small ntp and dns changes

etbe:


policy/modules/services/ntp.te
```diff
@@ -156,6 +156,8 @@ ifdef(`init_systemd',`
 	allow ntpd_t self:capability { fowner setpcap };
 	init_read_state(ntpd_t)
 	init_reload(ntpd_t)
+	init_start_generic_units(ntpd_t)
+	init_stop_generic_units(ntpd_t)
```
pebenito, line 160: What unit(s) is it operating on?

etbe: I don't know. I removed it and I'll do another PR at a later time if I find good evidence for this.

pebenito (approved):

PR #704, opened 2023-09-22
misc small email changes

etbe:


policy/modules/services/courier.te
```diff
@@ -129,20 +132,34 @@ dev_read_rand(courier_pcp_t)
 # POP3/IMAP local policy
 #
 
-allow courier_pop_t self:capability { setgid setuid };
+allow courier_pop_t self:capability { chown dac_read_search fowner setgid setuid };
+dontaudit courier_pop_t self:capability fsetid;
+allow courier_pop_t self:unix_stream_socket connectto;
```
pebenito, line 137: Please add `create_stream_socket_perms` to ensure it can create, listen, accept, etc.

policy/modules/services/postfix.te
```diff
@@ -538,10 +541,14 @@ files_dontaudit_search_var(postfix_map_t)
...
 auth_use_nsswitch(postfix_map_t)
 
+domain_use_interactive_fds(postfix_map_t)
+
 logging_send_syslog_msg(postfix_map_t)
 
 miscfiles_read_localization(postfix_map_t)
 
+userdom_use_user_ptys(postfix_map_t)
+
```
pebenito, line 551: It's been quite some time since I used postfix... how is this used interactively?

policy/modules/services/postfix.te
```diff
@@ -752,6 +760,10 @@ optional_policy(`
 	systemd_use_nss(postfix_showq_t)
 ')
 
+optional_policy(`
+	unconfined_run_to(postfix_showq_t, postfix_showq_exec_t)
```
pebenito, line 764: This interface is meant for downstream use only, so this should change to a call in unconfined.te to a postfix interface.

policy/modules/services/sendmail.te
```diff
@@ -173,6 +173,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	userdom_use_user_ttys(sendmail_t)
```
pebenito, line 176: To confirm, this is only related to posfix use? Also, is it truly limited only to ttys?

pebenito (changes requested):

policy/modules/services/courier.te
etbe: ok

policy/modules/services/postfix.te
etbe: The sysadmin runs postmap to create a hashed or indexed file out of a plain text file.

policy/modules/services/sendmail.te
etbe: It should apply to the sendmail program from every MTA which can be run from the command line and often is.

policy/modules/services/postfix.te
etbe: I removed that, I'll add a Postfix interface in a future PR.

policy/modules/services/postfix.te
pebenito: Why is this needed over having sysadm simply `execute_no_trans` on map and simply write the files?

pebenito: That works for me.

etbe: For now I'll just remove this and I'll add something else for /etc/aliases in a different PR

policy/modules/services/sendmail.te
pebenito: If it's not specific to postfix use, then it shouldn't be in the optional with postfix rules.

`userdom_use_inherited_user_terminals()` would cover ttys and ptys, plus prevent it from opening arbitrary terminals.

etbe: done

pebenito (approved):

PR #705, opened 2023-09-22
allow jabbers to create sock file and allow matrixd to read sysfs

etbe:


policy/modules/services/matrixd.te
```diff
@@ -83,6 +83,7 @@ corenet_udp_bind_generic_node(matrixd_t)
 corenet_udp_bind_generic_port(matrixd_t)
 corenet_udp_bind_reserved_port(matrixd_t)
 
+dev_read_sysfs(matrixd_t)
```
0xC0ncord, line 86: I'm curious, do you have an AVC for this? Was this seen on Synapse?

etbe: I only use synapse so that was where it was. I don't have an AVC. Do you run synapse without seeing it?

policy/modules/services/jabber.te
```diff
@@ -39,6 +39,7 @@ allow jabberd_domain self:tcp_socket { accept listen };
 
 manage_files_pattern(jabberd_domain, jabberd_var_lib_t, jabberd_var_lib_t)
 allow jabberd_domain jabberd_var_lib_t:dir manage_dir_perms;
+allow jabberd_domain jabberd_var_lib_t:sock_file create;
```
pebenito, line 42: Wouldn't the domain also delete the socket on shutdown? It seems like this should be `manage_sock_file_perms` instead.

etbe: done. I had the unlink access in my toadd.te file on my jabber server. ;)

pebenito (approved):

PR #706, opened 2023-09-22
small storage changes

etbe:


policy/modules/kernel/files.if
```diff
@@ -1426,6 +1426,25 @@ interface(`files_unmount_all_file_type_fs',`
...
+## <summary>
+##	watch all directories of file_type
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`files_watch_all_file_type_dir',`
```
pebenito, line 1439: files_watch_all_dirs

policy/modules/services/smartmon.if
```diff
@@ -56,3 +56,24 @@ interface(`smartmon_admin',`
...
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`fsdaemon_read_lib',`
+	gen_require(`
+		type fsdaemon_var_lib_t;
+	')
+
+	allow $1 fsdaemon_var_lib_t:dir search;
```
pebenito, line 76: please use `search_dir_perms`, alternatively switch the implementation to use `read_files_pattern`.

pebenito (changes requested):

policy/modules/kernel/files.if
etbe: done

policy/modules/services/smartmon.if
etbe: done

pebenito (approved):

PR #707, opened 2023-09-22
small network patches

etbe:


policy/modules/admin/netutils.te
```diff
@@ -142,6 +142,7 @@ logging_send_syslog_msg(ping_t)
 miscfiles_read_localization(ping_t)
 
 userdom_use_inherited_user_terminals(ping_t)
+term_use_unallocated_ttys(ping_t)
```
pebenito, line 145: What is the use case for this?

policy/modules/services/networkmanager.fc
```diff
@@ -15,7 +15,7 @@
 /etc/wicd/wireless-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 /etc/wicd/wired-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 
-/usr/lib/NetworkManager/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
+/usr/lib/NetworkManager/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_exec_t,s0)
 /usr/lib/networkmanager/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
 /usr/libexec/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
```
pebenito, line 20: This is labeling the same files but in different locations, so they should all be the same label.

That being said, please elaborate on why the current initrc labeling is incorrect.

policy/modules/system/libraries.if
```diff
@@ -531,3 +531,21 @@ interface(`libs_relabel_shared_libs',`
...
+## <summary>
+##	watch lib dirs
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`libs_watch_shared_libs_dir',`
```
pebenito, line 545: Please make the "dir" plural, i.e. `libs_watch_shared_libs_dirs`

policy/modules/system/sysnetwork.if
```diff
@@ -567,6 +567,24 @@ interface(`sysnet_manage_config',`
...
+## <summary>
+##     Watch a network config dir
+## </summary>
+## <param name="domain">
+##     <summary>
+##     Domain allowed access.
+##     </summary>
+## </param>
+#
+interface(`sysnet_watch_config_dir',`
```
pebenito, line 580: Same plural "dirs" as above.

pebenito (changes requested):

policy/modules/admin/netutils.te
etbe: Booting in recovery mode (boot failed enter root password) and running ping.

policy/modules/system/libraries.if
etbe: done

policy/modules/system/sysnetwork.if
etbe: done

policy/modules/services/networkmanager.fc
etbe: I can't remember why I did it so I reverted that. I'll do more tests and submit at a future time with evidence if needed.

pebenito (approved):

PR #708, opened 2023-09-26
small systemd patches

etbe:


policy/modules/system/init.if
```diff
@@ -3263,6 +3263,24 @@ interface(`init_tcp_recvfrom_all_daemons',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`init_restart_units',`
+	gen_require(`
+		type init_var_run_t;
+	')
+
+	allow $1 init_var_run_t:service { start status stop };
```
pebenito, line 3281: There already are interfaces for transient units (and a new type).

policy/modules/system/systemd.if
```diff
@@ -61,6 +57,7 @@ template(`systemd_role_template',`
 	allow $1_systemd_t $3:process { setsched rlimitinh signal_perms };
 	corecmd_shell_domtrans($1_systemd_t, $3)
 	corecmd_bin_domtrans($1_systemd_t, $3)
+	corecmd_shell_entry_type($1_systemd_t)
```
pebenito, line 60: When is this needed?

policy/modules/system/systemd.if
```diff
@@ -153,6 +150,12 @@ template(`systemd_role_template',`
 	allow $3 systemd_user_runtime_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
 	allow $3 systemd_user_runtime_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
 
+	# for "machinectl shell"
+	allow $1_systemd_t systemd_machined_t:fd use;
+	allow $3 systemd_machined_t:fd use;
+	allow $3 systemd_machined_t:dbus send_msg;
+	allow systemd_machined_t $3:dbus send_msg;
```
pebenito, line 157: This doesn't seem like something we'd want on all users.

pebenito (changes requested):

policy/modules/system/init.if
etbe: ok

policy/modules/system/systemd.if
etbe: Doesn't seem to be needed now so I'll remove it.

policy/modules/system/systemd.if
etbe: Why not? systemd-machined requires authentication before doing things.

pebenito (approved):

PR #710, opened 2023-09-26
Separate label for /run/systemd/notify

dsugar100:
label systemd_runtime_notify_t
Allow daemon domains to write by default

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -126,6 +126,7 @@ template(`systemd_role_template',`
 	systemd_search_user_runtime_unit_dirs($1_systemd_t)
 	systemd_search_user_transient_unit_dirs($1_systemd_t)
 	systemd_read_user_units_files($1_systemd_t)
+	systemd_write_notify_socket($1_systemd_t)
 
 	dbus_system_bus_client($1_systemd_t)
 	dbus_spec_session_bus_client($1, $1_systemd_t)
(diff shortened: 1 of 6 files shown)


dsugar100:
This is intended as an alternative approach for #692 

It will need some review and testing.  There are other domains that use init_write_runtime_socket that may need to be updated to systemd_write_notify_socket like podman, kuberneties, docker, logrotate.

policy/modules/system/systemd.if
```diff
@@ -274,6 +275,25 @@ interface(`systemd_user_unix_stream_activated_socket',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_write_notify_socket',`
+	gen_require(`
+		type systemd_runtime_notify_t;
+	')
+
+	allow $1 systemd_runtime_notify_t:sock_file write;
```
pebenito, line 294: It needs:
* the dir accesses to get to the sock_file
* a unix socket connectto/sendto init_t.

pebenito (changes requested):

policy/modules/system/systemd.if
dsugar100: Ok, I did both of these.  The dir access makes sense.  I'm not sure I understand the unix socket change and I think I did what you are asking.  Of course, if I misunderstood, let me know and I will alter it.

dsugar100: Ok, just the sendto.  Hopefully this is now good.

pebenito (approved):

PR #711, opened 2023-09-29
misc small patches

etbe:

PR #712, opened 2023-09-29
misc kernel and system patches

etbe:


policy/modules/admin/usermanage.te
```diff
@@ -69,7 +69,7 @@ role useradd_roles types useradd_t;
 # Chfn local policy
 #
 
-allow chfn_t self:capability { chown dac_override fsetid setgid setuid sys_resource };
+allow chfn_t self:capability { chown dac_override fsetid setgid setuid sys_ptrace sys_resource };
```
dsugar100, line 72: sys_ptrace doesn't seem correct here.  Why does chfn_t need to ptrace?

etbe: Not sure and I can't reproduce it so I'll remove it.

pebenito (changes requested):
Please revise the PR as I described in #711.

etbe:
I'll do new smaller PRs for ths.

PR #717, opened 2023-10-04
Use interface that already exists.

dsugar100:


policy/modules/system/systemd.if
```diff
@@ -152,8 +152,7 @@ template(`systemd_role_template',`
 	# for "machinectl shell"
 	allow $1_systemd_t systemd_machined_t:fd use;
 	allow $3 systemd_machined_t:fd use;
```
dsugar100, line 154: Should these also be changed to 'systemd_use_inherited_machined_ptys'?

pebenito (approved):

PR #722, opened 2023-10-07
systemd: allow systemd-networkd and sytemd-resolved to write to syste…

yizhao1:
…md-notify socket

Fixes:
avc:  denied  { write } for  pid=277 comm="systemd-resolve" name="notify" dev="tmpfs" ino=31
scontext=system_u:system_r:systemd_resolved_t
tcontext=system_u:object_r:systemd_runtime_notify_t tclass=sock_file permissive=1

avc:  denied  { write } for  pid=324 comm="systemd-network" name="notify" dev="tmpfs" ino=31
scontext=system_u:system_r:systemd_networkd_t
tcontext=system_u:object_r:systemd_runtime_notify_t tclass=sock_file permissive=1

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -195,7 +195,7 @@ init_daemon_domain(systemd_modules_load_t, systemd_modules_load_exec_t)
 
 type systemd_networkd_t;
 type systemd_networkd_exec_t;
-init_system_domain(systemd_networkd_t, systemd_networkd_exec_t)
+init_daemon_domain(systemd_networkd_t, systemd_networkd_exec_t)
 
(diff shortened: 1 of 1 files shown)


0xC0ncord:
Since `systemd-networkd` and `systemd-resolved` are daemons, shouldn't we change them from `init_system_domain()` to `init_daemon_domain()`? This would also fix the issue.

yizhao1:
> Since `systemd-networkd` and `systemd-resolved` are daemons, shouldn't we change them from `init_system_domain()` to `init_daemon_domain()`? This would also fix the issue.

Thanks. It's a better solution.

pebenito (approved):

PR #724, opened 2023-10-08
make resolved work with the new labelling for the notification socket

etbe:


pebenito:
Already fixed in #722

PR #725, opened 2023-10-10
$1_wm_t changes including on-screen keyboard

etbe:


pebenito:
Is this supported by Debian? I'm uneasy adding these since phone window manager seems to be a bit too niche.

etbe:
It is supported by Debian. Most of the changes aren't phone specific. The on-screen keyboard functionality also applies to tablets and 2in1 systems.

PR #728, opened 2023-10-23
Fix modutils kernel module loading file contexts

chrisschnei:
File contexts assigned to e.g. modprobe symlink are currently bin_t. kmod_exec_t should be applied.


pebenito:
Sorry for the delay, I was away for two weeks and I'm catching up.

The intent for symlinks in `bin_t` directories is for them to also be `bin_t`. The symlinks being the same label as their target does not add additional benefit from my perspective.

Is there something in particular you are trying to achieve?

chrisschnei:
Hello @pebenito ,
no problem. Also from my side the symlinks could have bin_t while the kmod binary is labelled as kmod_exec_t.
In this case we can rely on the inherited bin_t label and remove the symlinks (lsmod, rmmod, modprobe, ...). Only the /bin/kmod entry would remain.
The idea of this commit is to either apply kmod_exec_t to the symlinks or remove them to have consistent contexts.

pebenito:
The remaining lsmod, rmmod, etc. entries are to cover old systems prior to when they were merged to kmod.

The important point is to have the executables themselves be labeled `kmod_exec_t`. If you do exec() on a symlink, the symlink is simply read, and then all the important exec() checks are done on the target; in this case kmod. We want the symlinks to remain `bin_t` to keep things simple, since reading `bin_t` symlinks isn't any better or worse security-wise than reading `kmod_exec_t` symlinks. Having kmod_exec_t only adds complexity.

chrisschnei:
Thanks for the information. I will close the PR then.

PR #733, opened 2023-11-18
Secure the Private Keys corresponding to SSL Certificates used by the HTTP daemon

gtrentalancia:
Secure the Private Keys corresponding to SSL Certificates used by the HTTP daemon against an Information Disclosure vulnerability.

The new file contexts are based upon the official Apache HTTP Server recommended locations (see http://www.apache.com/how-to-setup-an-ssl-certificate-on-apache) and might need to be adapted for other distributions.

Modify the webadm role with new tunable policy to support file management of SSL Private Keys.

Fix the certmonger policy module so that it can manage SSL Certificates and corresponding SSL Private Keys and CSR.

This pull request is now being replaced by a revised solution: https://github.com/SELinuxProject/refpolicy/pull/737

### DIFF
--- a/policy/modules/roles/webadm.te
+++ b/policy/modules/roles/webadm.te
@@ -8,6 +8,16 @@ policy_module(webadm)
 ## <desc>
 ##	<p>
 ##	Determine whether webadm can
+##	manage http daemon secret
+##	configuration files such as
+##	private keys for SSL certs.
+##	</p>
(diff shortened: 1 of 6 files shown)


gtrentalancia:
Superseded by a revised pull-request: https://github.com/SELinuxProject/refpolicy/pull/737

Issue #735, opened 2023-11-18
Information Disclosure vulnerability related to SSL Private Keys and CSR used by the HTTP daemon


gtrentalancia:
A serious **Information Disclosure vulnerability** exists in the current Reference Policy file contexts and permissions handling the SSL Private Keys and CSR files used by the HTTP and Mail Transport daemons.

The SSL Private Keys and CSR are highly sensitive secret files and the issue probably arises from confusing them with CA certificates or SSL certificates, which are public and not sensitive secret files.

If any of the very wide number of modules currently allowed to access those files through the miscfiles_read_generic_certs(), miscfiles_manage_generic_cert_files(), apache_{read,manage}_config() is exploited by the means, for example, of rogue binaries or network-borne buffer overflow attacks, then a very serious Information Disclosure incident arises, leading to the leak of secret files locally or over the network.

A leak of the SSL Private Keys and CSR secret files would then allow **unlawful interception of encrypted SSL traffic or the malicious impersonation of the Internet domains** hosted by the HTTP daemon.

Generic distributions, as well as Debian, Ubuntu and Gentoo file locations in the current Reference Policy are all affected, while RedHat file locations are not affected thanks to the original work in https://github.com/SELinuxProject/refpolicy/commit/cc91fed88d4cedaa4488f045a2da1132d9aeddca which apparently has not been followed up by others.

A patch to fix the issue is proposed in https://github.com/SELinuxProject/refpolicy/pull/737.

Further work possibly needed is field-testing on a wider number of HTTP and/or Mail server setups.

gtrentalancia:
Please note that if the Reference Policy previous to merging of https://github.com/SELinuxProject/refpolicy/issues/735 and https://github.com/SELinuxProject/refpolicy/pull/736 is actually being used on running servers, then the two vulnerabilities might need to be reported on public databases such as those maintained CISA (https://www.cisa.gov/report), NIST (https://nvd.nist.gov) or Mitre (https://cve.mitre.org).

Finally if there are instances of the Reference Policy deployed on HTTP, IMAP, SMTP or LDAP servers prior to merging of https://github.com/SELinuxProject/refpolicy/issues/735 and https://github.com/SELinuxProject/refpolicy/pull/736 using TLS/SSL Private Keys at the involved file locations, then those TLS/SSL Certificates and Private Keys would need to be revoked and regenerated, as they might have been leaked and compromised.

gtrentalancia:
[...]

PR #736, opened 2023-11-19
Restrict LDAP server init script permissions on generic certificate files

gtrentalancia:
The LDAP server only needs to read generic certificate files, not manage them.

Modify the init policy to match the comment and the LDAP server actual behavior.

---
 policy/modules/system/init.te |    2 +-

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -914,7 +914,7 @@ logging_read_audit_config(initrc_t)
 
 miscfiles_read_localization(initrc_t)
 # slapd needs to read cert files from its initscript
-miscfiles_manage_generic_cert_files(initrc_t)
+miscfiles_read_generic_certs(initrc_t)
 
 seutil_read_config(initrc_t)
 


pebenito:
I'm reluctant to make this change because it appears the comment itself is wrong. The rule was upgraded from `read` to`manage` in 2087bde, but the comment wasn't dropped. This change will likely break other things.

gtrentalancia:
The change you mentioned https://github.com/SELinuxProject/refpolicy/commit/2087bde93412d9bbbb92fd8316b5e8176451d836 introduced a vulnerability because initrc scripts and/or openldap should not be able to manage certificates: it simply isn't their role.

There is absolutely no reason for allowing such permission.

For example, fake Certification Authorities (CA) certificates could be introduced, leading to a serious compromission of the integrity and confidentiality of encrypted TLS/SSL communications.

The LDAP server should only be allowed to read the CA certificates, not tamper them.

gtrentalancia:
Consider the https://github.com/SELinuxProject/refpolicy/commit/2087bde93412d9bbbb92fd8316b5e8176451d836 change possibly introduced vulnerabilities in the init policy just to keep running the toute-fair unbounded-growth init daemon.

Many secure systems might not want or need such an overcomplex init daemon, for many security reasons, including this one.

pebenito:
> There is absolutely no reason for allowing such permission.

That is a very strong opinion, but unfortunately, I'm not as sure as you are.

> For example, fake Certification Authorities (CA) certificates could be introduced, leading to a serious compromission of the integrity and confidentiality of encrypted TLS/SSL communications.

I understand the risks, but I think there are key assumptions that we have for them:
1. They are necessary for system init/shutdown and do highly privileged operations.
2. PID 1 (or subprocesses that run the init scripts) is high integrity. If they're not, then the security of the system collapses.
3. The script file itself is high integrity. If they're not, then the security of the system collapses.
4. They are short running, making attacking the process (e.g. by sockets or shm) difficult.

With these in mind, I'm comfortable with the access remaining unless there is reasonable evidence to demonstrate it won't break things or was an incorrect change in the first place.

> Consider the [2087bde](https://github.com/SELinuxProject/refpolicy/commit/2087bde93412d9bbbb92fd8316b5e8176451d836) change possibly introduced vulnerabilities in the init policy just to keep running the toute-fair unbounded-growth init daemon.

The reality is that systemd isn't going anywhere anytime soon. In refpolicy, we need to do our best to keep it protected but not intentionally break it.

gtrentalancia:
I'll try to make it simpler.

LDAP has its own certificates which are labeled as _slapd_cert_t_ and it might need to "manage" them in the sense of _ldapmodify_, _ldapadd_, _ldapdelete_ and so on, which constitute licit operations.

The change in https://github.com/SELinuxProject/refpolicy/commit/2087bde93412d9bbbb92fd8316b5e8176451d836 which amends miscfiles_read_generic_certs(initrc_t) with miscfiles_manage_generic_cert_files(initrc_t) is wrong as it introduces a system-wide cross-services **Information Disclosure vulnerability** in that it allows LDAP to modify CA Certificates (and at the moment also SSL/TLS Private Keys and CSR, see https://github.com/SELinuxProject/refpolicy/issues/735 for further reference).

I can give you an example: a malicious init rc script (not necessary the slapd init script) can modify CA Certificates by adding bogus ones and/or tamper with SSL/TLS Private Keys and CSR to allow illicit eavesdropping of encrypted communications (SIGINT operations).

Does it make sense now ?

pebenito:
> The change in [2087bde](https://github.com/SELinuxProject/refpolicy/commit/2087bde93412d9bbbb92fd8316b5e8176451d836) which amends miscfiles_read_generic_certs(initrc_t) with miscfiles_manage_generic_cert_files(initrc_t) is wrong as it introduces a system-wide cross-services **Information Disclosure vulnerability** in that it allows LDAP to modify CA Certificates

You're conflating LDAP with its init scripts, which aren't the same thing.

However, I'm prepared to apply this patch to revalidate the need for this permission. Understand this: if is required for proper systemd operation, the permission will be restored.

pebenito (approved):
Applying to revalidate need for this access.

gtrentalancia:
The systemd patches, in particular the one being mentioned in this vulnerability introduced a lot of weaknesses and vulnerabilities in the init policy without even enclosing them in an _init_systemd_ ifdef block.

Consider that systemd is not needed to run a modern Linux system, that systemd does not comply with the original Unix Laboratories System V Specifications and that the permissions that have been added indicate very suspicious activity.

PR #737, opened 2023-11-19
Secure the Private Keys corresponding to SSL Certificates used by the HTTP daemon

gtrentalancia:
Secure the Private Keys corresponding to SSL Certificates used by the HTTP daemon in order to fix a serious **Information Disclosure vulnerability** caused by the erroneous labeling of TLS Private Keys and CSR.

The new file contexts are based upon the official Apache HTTP Server recommended locations (see http://www.apache.com/how-to-setup-an-ssl-certificate-on-apache), have been extended to Debian and Gentoo locations, but might need to be further customized for other possible file locations which might still be exposed to the vulnerability.

RedHat distributions are not affected by this issue, thanks to https://github.com/SELinuxProject/refpolicy/commit/cc91fed88d4cedaa4488f045a2da1132d9aeddca but apparentlly such change was not taken up by other distributions file contexts specifications.

Fixes: https://github.com/SELinuxProject/refpolicy/issues/735

Replaces: https://github.com/SELinuxProject/refpolicy/pull/733

Also fixes the certmonger module so that it can manage the above mentioned secret files.

### DIFF
--- a/policy/modules/system/miscfiles.fc
+++ b/policy/modules/system/miscfiles.fc
@@ -10,10 +10,13 @@ ifdef(`distro_gentoo',`
 #
 /etc/avahi/etc/localtime --	gen_context(system_u:object_r:locale_t,s0)
 /etc/httpd/alias/[^/]*\.db(\.[^/]*)* -- gen_context(system_u:object_r:cert_t,s0)
+/etc/httpd/conf/ssl(/.*)?	--	gen_context(system_u:object_r:tls_privkey_t,s0)
+/etc/httpd/conf/ssl/.*\.crt	--	gen_context(system_u:object_r:cert_t,s0)
 /etc/localtime		--	gen_context(system_u:object_r:locale_t,s0)
 /etc/pki(/.*)?			gen_context(system_u:object_r:cert_t,s0)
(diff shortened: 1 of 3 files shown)


policy/modules/system/miscfiles.fc
```diff
@@ -10,10 +10,13 @@ ifdef(`distro_gentoo',`
 #
 /etc/avahi/etc/localtime --	gen_context(system_u:object_r:locale_t,s0)
 /etc/httpd/alias/[^/]*\.db(\.[^/]*)* -- gen_context(system_u:object_r:cert_t,s0)
+/etc/httpd/conf/ssl(/.*)?	--	gen_context(system_u:object_r:tls_privkey_t,s0)
```
pebenito, line 13: Maybe this should be a new type, since it's httpd's private key, not a general/generic one.

gtrentalancia: Yes, it could be a new type, but it won't add any gain in terms of security at this time.

Consider, the HTTP daemon is the main consumer of TLS/SSL Certificates and corresponding Private Keys. The second main consumer is the Mail Transport daemon. Then possibly the LDAP daemon.

I think at the moment we should focus on fixing the Information Disclosure vulnerability in the shortest time possible, it can be later improved if needed.

pebenito (approved):

PR #700, opened 2023-09-21
cockpit web management system

etbe:

PR #740, opened 2023-12-11
Add SELinux policy for cockpit

dsugar100:
This was started from #700 and tested on RHEL9 system.
It is a bunch of commits because I wanted to be clear about some of the denials I was seeing and why those rules were added.
I of course will squash before the final merge (if desired).

Please let me know questions and changes desired.

### DIFF
--- a/policy/modules/admin/sudo.if
+++ b/policy/modules/admin/sudo.if
@@ -199,7 +199,7 @@ template(`sudo_role_template',`
 	')
 
 	optional_policy(`
-		dbus_system_bus_client($1_sudo_t)
+		auth_use_pam_systemd($1_sudo_t)
 
 		ifdef(`init_systemd',`
 			init_dbus_chat($1_sudo_t)
--- a/policy/modules/admin/sudo.if
+++ b/policy/modules/admin/sudo.if
@@ -76,6 +76,7 @@ template(`sudo_role_template',`
 
 	# allow getting the process group of the parent process
 	allow $1_sudo_t $2:process getpgid;
+	allow $1_sudo_t $2:unix_stream_socket rw_socket_perms;
 
 	# allow accessing /proc/pid/stat of the calling domain
 	ps_process_pattern($1_sudo_t, $2)
--- a/policy/modules/services/cockpit.fc
+++ b/policy/modules/services/cockpit.fc
@@ -1,0 +1,20 @@
+# cockpit stuff
+
+/etc/cockpit/ws-certs\.d(/.*)?			gen_context(system_u:object_r:cockpit_cert_t,s0)
+
+/run/cockpit(/.*)?						gen_context(system_u:object_r:cockpit_runtime_t,s0)
+/run/cockpit/certificate-helper(/.*)	gen_context(system_u:object_r:cockpit_cert_t,s0)
+
+/usr/lib/systemd/system/cockpit.*		--	gen_context(system_u:object_r:cockpit_unit_t,s0)
+
+/usr/libexec/cockpit-certificate-ensure	--	gen_context(system_u:object_r:cockpit_cert_manage_exec_t,s0)
+/usr/libexec/cockpit-ws		--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-tls	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-wsinstance-factory	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/libexec/cockpit-ssh	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+
+/var/lib/cockpit(/.*)?      gen_context(system_u:object_r:cockpit_state_t,s0)
+
+/var/run/cockpit(/.*)?   gen_context(system_u:object_r:cockpit_runtime_t,s0)
+/var/run/cockpit-ws(/.*)?   gen_context(system_u:object_r:cockpit_runtime_t,s0)
--- a/policy/modules/services/cockpit.if
+++ b/policy/modules/services/cockpit.if
@@ -56,9 +56,16 @@ template(`cockpit_role_template',`
 	files_dontaudit_execute_default_files($2)
 	files_dontaudit_execuite_etc_runtime_files($2)
 	files_dontaudit_exec_runtime($2)
+	files_watch_etc_files($2)
+	files_watch_root_dirs($2)
+	files_watch_var_dirs($2)
+	files_watch_var_lib_dirs($2)
 
 	cockpit_use_ws_fds($2)
 	cockpit_rw_ws_stream_sockets($2)
+
+	init_watch_runtime_dirs($2)
+	init_watch_utmp($2)
 
 	userdom_dontaudit_execute_user_tmpfs_files($2)
 ')
--- a/policy/modules/services/cockpit.if
+++ b/policy/modules/services/cockpit.if
@@ -203,6 +203,24 @@ interface(`cockpit_delete_cert_files',`
 
 ########################################
 ## <summary>
+##	Allow cockpit to send signals to another domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to send to,
+##	</summary>
+## </param>
+#
+interface(`cockpit_send_signal',`
+	gen_require(`
+		type cockpit_session_t;
+	')
+
+	allow cockpit_session_t $1:process signal;
+')
+
+########################################
+## <summary>
 ##	All of the rules required to administrate
 ##	an cockpit environment
 ## </summary>
--- a/policy/modules/services/cockpit.if
+++ b/policy/modules/services/cockpit.if
@@ -51,8 +51,16 @@ template(`cockpit_role_template',`
 
 	allow $2 $1_cockpit_tmpfs_t:file { manage_file_perms execute };
 
+	dev_dontaudit_execute_dev_nodes($2)
+
+	files_dontaudit_execute_default_files($2)
+	files_dontaudit_execuite_etc_runtime_files($2)
+	files_dontaudit_exec_runtime($2)
+
 	cockpit_use_ws_fds($2)
 	cockpit_rw_ws_stream_sockets($2)
+
+	userdom_dontaudit_execute_user_tmpfs_files($2)
 ')
 
 ########################################
--- a/policy/modules/services/cockpit.if
+++ b/policy/modules/services/cockpit.if
@@ -109,6 +109,42 @@ interface(`cockpit_manage_runtime_symlnks',`
 
 ########################################
 ## <summary>
+##	Inherit and use cockpit session file descriptors.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cockpit_use_session_fds',`
+	gen_require(`
+		type cockpit_session_t;
+	')
+
+	allow $1 cockpit_session_t:fd use;
+')
+
+########################################
+## <summary>
+##	Read and write cockpit session unnamed pipes.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cockpit_rw_session_pipes',`
+	gen_require(`
+		type cockpit_session_t;
+	')
+
+	allow $1 cockpit_session_t:fifo_file rw_fifo_file_perms;
+')
+
+########################################
+## <summary>
 ##	Manage the cockpit certificate files.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/services/cockpit.if
+++ b/policy/modules/services/cockpit.if
@@ -1,0 +1,202 @@
+## <summary>Cockpit web management system for Linux</summary>
+##
+## <desc>
+## Cockpit is a web console that enables users to administer Linux servers
+## via a web browser.
+## see https://cockpit-project.org/
+##
+## For linux logins that are allowed access they must be associated with a
+## SELinux user that uses ssh_role_template (sysadm, system).  To be able
+## to alter system settings the must be allowed sudo access.
+## </desc>
+
+########################################
+## <summary>
+##	Transition to the cockpit session domain.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed to transition.
+## </summary>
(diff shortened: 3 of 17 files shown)


policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+## <summary>
+##	Allow specified domain to get status of cockpit service
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cockpit_status',`
```
pebenito, line 102: `cockpit_get_service_status`

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+## </param>
+#
+interface(`cockpit_startstop',`
+	gen_require(`
+		type cockpit_unit_t;
+		class service { start stop };
+	')
+
+	allow $1 cockpit_unit_t:service { start stop };
+')
```
pebenito, line 147: `startstop` and `enabledisable` should have the `status` perm too.

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+## <summary>
+##	Create, read, write, and delete the cockpick runtime symlink files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cockpit_manage_runtime_lnk_files',`
```
pebenito, line 159: `lnk_files` -> `symlinks`

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+	gen_require(`
+		type cockpit_runtime_t;
+	')
+
+	files_search_runtime($1)
+	manage_lnk_files_pattern($1, cockpit_runtime_t, cockpit_runtime_t)
+')
+
+## <summary>
```
pebenito, line 168: Incomplete comment block

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+#
+interface(`cockpit_rw_session_pipes',`
+	gen_require(`
+		type cockpit_session_t;
+	')
+
+	allow $1 cockpit_session_t:fifo_file rw_fifo_file_perms;
+')
+
+## <summary>
```
pebenito, line 203: Incomplete comment block.

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+##	Read and write cockpit web service stream socket
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`cockpit_rw_ws_stream_socket',`
```
pebenito, line 230: Plural "sockets"

policy/modules/services/cockpit.te
```diff
@@ -0,0 +1,266 @@
...
+init_daemon_runtime_file(cockpit_runtime_t, lnk_file, "motd")
+
+optional_policy(`
+	systemd_tmpfilesd_managed(cockpit_runtime_t)
+')
+
+type cockpit_unit_t;
+init_unit_file(cockpit_unit_t)
+
+type cockpit_var_lib_t;
```
pebenito, line 33: Please rename to `cockpit_state_t`.

policy/modules/services/cockpit.te
```diff
@@ -0,0 +1,266 @@
...
+# cockpit-session communicates back with cockpit-ws
+allow cockpit_session_t cockpit_ws_t:unix_stream_socket rw_stream_socket_perms;
+
+# cockpit-tls and cockpit-ws communicate over a Unix socket
+allow cockpit_ws_t self:unix_stream_socket { create_stream_socket_perms connectto };
+
+allow cockpit_ws_t cockpit_cert_t:file unlink;
+
+ifdef(`init_systemd',`
+	systemd_exec_systemctl(cockpit_ws_t)
```
pebenito, line 134: This can be a regular optional.

policy/modules/services/cockpit.te
```diff
@@ -0,0 +1,266 @@
...
+
+dbus_system_bus_client(cockpit_session_t)
+
+miscfiles_read_localization(cockpit_session_t)
+
+ifdef(`init_systemd',`
+	systemd_dbus_chat_logind(cockpit_session_t)
+	systemd_search_all_user_keys(cockpit_session_t)
+	systemd_write_all_user_keys(cockpit_session_t)
+')
```
pebenito, line 207: Can be an optional block.

policy/modules/system/init.if
```diff
@@ -3044,6 +3044,24 @@ interface(`init_manage_utmp',`
...
+## <summary>
+##	Add a watch on init runtime
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`init_watch_runtime',`
```
pebenito, line 3057: `init_watch_runtime_dirs`

policy/modules/system/init.te
```diff
@@ -303,6 +303,7 @@ ifdef(`init_systemd',`
 
 	allow init_t init_runtime_t:sock_file manage_sock_file_perms;
 
+	allow init_t daemon:process noatsecure;
```
pebenito, line 306: This should be a separate patch. Also it wouldn't be here, but rather in `init_domain`, `init_daemon_domain` etc.

policy/modules/system/systemd.if
```diff
@@ -161,6 +162,8 @@ template(`systemd_role_template',`
 
 	can_exec($3, systemd_analyze_exec_t)
 
+	domtrans_pattern($1_systemd_t, systemd_tmpfiles_exec_t, systemd_tmpfiles_t)
```
pebenito, line 165: The systemd user sessions definitely shouldn't be starting a system service.

pebenito (changes requested):
`cockpit.te` also needs some style cleanups.

policy/modules/system/systemd.if
dsugar100: I'm open to suggestions on this one.  I started down the path of setting up a user tmpfiles concept (i.e. setting up $1_systemd_tmpfiles_t), but it quickly got out of hand on permissions.  I can try this approach again unless there is a better alternative?

policy/modules/services/cockpit.te
dsugar100: updated

policy/modules/system/init.if
dsugar100: fixed

policy/modules/system/systemd.if
0xC0ncord: I agree transitioning from the systemd user session to the tmpfiles domain directly is a pretty significant elevation of privileges.

What kinds of things does this operation need to do? I'm not familiar with the innerworkings of cockpit and how this is related.

policy/modules/services/cockpit.if
dsugar100: done - added use of 'cockpit_status'

policy/modules/services/cockpit.if
dsugar100: renamed

policy/modules/services/cockpit.if
dsugar100: renamed

policy/modules/services/cockpit.if
dsugar100: renamed

policy/modules/system/init.te
dsugar100: Moved into init_daemon_domain and into patch #752

policy/modules/services/cockpit.fc
```diff
@@ -0,0 +1,22 @@
...
+/usr/lib/systemd/system/cockpit.*		--	gen_context(system_u:object_r:cockpit_unit_t,s0)
+
+/usr/libexec/cockpit-certificate-ensure	--	gen_context(system_u:object_r:cockpit_cert_manage_exec_t,s0)
+/usr/libexec/cockpit-ws		--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-tls	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-wsinstance-factory	--	gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
+/usr/libexec/cockpit-session	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+/usr/libexec/cockpit-ssh	--	gen_context(system_u:object_r:cockpit_session_exec_t,s0)
+
+/usr/share/cockpit/motd/update-motd    -- gen_context(system_u:object_r:cockpit_ws_exec_t,s0)
```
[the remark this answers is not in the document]
dsugar100: Either because I don't think we wanted bin_t in cockpit.fc and/or because I didn't think we want /usr/share/cockpit/... in corecommands.fc (which upon further inspection might be OK) and/or because as I was working on this to get things running in enforcing I ended up with this and didn't give it much more thought.  

Either way, bin_t seems like a reasonable label.  I will do more testing to ensure it works as expected.

policy/modules/services/cockpit.if
dsugar100: It too me a stupid amount of time to realize it was missing the line of '#'s above 'summary'.

policy/modules/services/cockpit.if
dsugar100: done

dsugar100:
Changed status to 'draft'.   I'm still working through changes.  I'm happy for more feedback, but don't want anyone to think my updates are indications that I'm done.

policy/modules/admin/sudo.if
```diff
@@ -203,6 +204,8 @@ template(`sudo_role_template',`
 
 		ifdef(`init_systemd',`
 			init_dbus_chat($1_sudo_t)
+			systemd_dbus_chat_logind($1_sudo_t)
+			systemd_read_logind_state($1_sudo_t)
 		')
 	')
```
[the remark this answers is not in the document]
dsugar100: I think I have updated as you asked.  I did leave the "ifdef(init_systemd)" as that was there before.  Let me know if I misunderstood or you want additional changes.

policy/modules/services/ntp.te
```diff
@@ -171,6 +171,7 @@ ifdef(`init_systemd',`
 	optional_policy(`
 		chronyd_enabledisable(ntpd_t)
 		chronyd_startstop(ntpd_t)
+		chronyd_status(ntpd_t)
```
dsugar100, line 174: Related to the comment about cockpit_enabledisable and cockpit_startstop granting 'status' access, should this be here, or should I update chronyd_enabledisable and chronyd_startstop to allow 'status' access instead?

policy/modules/services/cockpit.if
```diff
@@ -0,0 +1,350 @@
...
+##	<summary>
+##	Role allowed access
+##	</summary>
+## </param>
+#
+template(`cockpit_role_template',`
+
+	type $1_cockpit_memfd_t;
+	files_runtime_file($1_cockpit_memfd_t)
+	dev_filetrans($2, $1_cockpit_memfd_t, file)
```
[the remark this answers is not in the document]
dsugar100: It maybe that the name is just wrong there.  But, there is access where cockpit-bridge (running as the user) is attempting to create and file and then execute that file.  The default type was user_tmpfs_t (and I didn't want to grant blanket execute permission on that type).  It does try creating these files in several different locations.  It seemed to be part of libffi (foreign-function interface).  

Maybe $1_cockpit_tmpfs_t would be a better type?

policy/modules/system/systemd.if
dsugar100: Ok, after reviewing this, it does seem that $1_systemd_tmpfiles_t is working ok and not an unreasonable amount of policy.  Maybe I was looking at this too soon during the development of the overall cockpit poilcy and was seeing something else that was confusing me.

dsugar100:
Ok, I have made updates.  There is a remaining question about the memfd thing which I will update after feedback and new question about chronyd_status.  Please let me know if there is anything else.

policy/modules/admin/sudo.if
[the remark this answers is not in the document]
pebenito: Ideally everything for enabling systemd would depend on the presence of the systemd module. In cases like this, the `ifdef(init_systemd)` is because there wasn't anything depending on the systemd module. By adding the `systemd_*` calls, now there are dependencies.

policy/modules/services/cockpit.if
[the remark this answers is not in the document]
pebenito: Lets go with the `tmpfs` label. By default memfd_create() will result in new shm segments. shm segments should be `tmpfs` in other places in the policy.

For future reference, here's the set of locations libffi will try to create the write+exec file (at the time I looked for this comment): https://github.com/libffi/libffi/blob/622caabcd25f4e11f752241417d06c9062acdf1f/src/closures.c#L726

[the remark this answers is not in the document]
dsugar100: Ok, updated to $1_cockpit_tmpfs_t

policy/modules/admin/sudo.if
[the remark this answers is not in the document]
dsugar100: Ok, ifdef(init_systemd) also removed.

dsugar100:
Ok, I believe at this point all your comments have been addressed.   Do you want me to resolve conversations that I believe are complete?
Thanks for the time, I know this is a big pull request.

policy/modules/admin/sudo.if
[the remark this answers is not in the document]
pebenito: Sorry, I made a mistake. Please keep the ifdef. I confused this with one of the other changes.

policy/modules/services/ntp.te
pebenito: > should I update chronyd_enabledisable and chronyd_startstop to allow 'status' access instead?

Yes, this.

policy/modules/admin/sudo.if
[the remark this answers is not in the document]
dsugar100: replace ifdef

policy/modules/services/cockpit.te
dsugar100: moved

policy/modules/services/ntp.te
dsugar100: updated

dsugar100:
Ok, I have updated based on the three comments about the ifdef, chronyd_status and reordering.  Let me know if anything else is needed.

pebenito (approved):

Issue #732, opened 2023-11-16
systemd v255 executor helper


cgzones:
With systemd version 255 services are spawned via a new helper binary (*/usr/lib/systemd/systemd-executor*) instead of forked from pid1 (https://github.com/systemd/systemd/blob/v255-rc2/NEWS#L57-L71).

This new executable should run in `init_t` to mimic the previous relation between the init domain and service domains.

Example denials:

```
----
type=PROCTITLE msg=audit(16/11/23 20:39:49.389:192) : proctitle=/usr/lib/systemd/systemd-journald 
type=PATH msg=audit(16/11/23 20:39:49.389:192) : item=1 name=/lib64/ld-linux-x86-64.so.2 inode=398326 dev=fe:01 mode=file,755 ouid=root ogid=root rdev=00:00 obj=system_u:object_r:ld_so_t:s0 nametype=NORMAL cap_fp=none cap_fi=none cap_fe
=0 cap_fver=0 cap_frootid=0 
type=PATH msg=audit(16/11/23 20:39:49.389:192) : item=0 name=/usr/lib/systemd/systemd-journald inode=435892 dev=fe:01 mode=file,755 ouid=root ogid=root rdev=00:00 obj=system_u:object_r:syslogd_exec_t:s0 nametype=NORMAL cap_fp=none cap_f
i=none cap_fe=0 cap_fver=0 cap_frootid=0 
type=CWD msg=audit(16/11/23 20:39:49.389:192) : cwd=/ 
type=EXECVE msg=audit(16/11/23 20:39:49.389:192) : argc=1 a0=/usr/lib/systemd/systemd-journald 
type=BPRM_FCAPS msg=audit(16/11/23 20:39:49.389:192) : fver=0 fp=none fi=none fe=0 old_pp=chown,dac_override,dac_read_search,fowner,fsetid,kill,setgid,setuid,setpcap,linux_immutable,net_bind_service,net_broadcast,net_admin,net_raw,ipc_l
ock,ipc_owner,sys_module,sys_rawio,sys_chroot,sys_ptrace,sys_pacct,sys_admin,sys_boot,sys_nice,sys_resource,sys_time,sys_tty_config,mknod,lease,audit_write,audit_control,setfcap,mac_override,mac_admin,syslog,wake_alarm,block_suspend,aud
[...]

PR #741, opened 2023-12-18
Various fixes

0xC0ncord:
Some various fixes including changes needed to fix new AVCs with systemd 255.

With this set of changes we also implement the proposed fix in https://github.com/SELinuxProject/refpolicy/issues/732.

### DIFF
--- a/policy/modules/kernel/kernel.te
+++ b/policy/modules/kernel/kernel.te
@@ -390,10 +390,16 @@ ifdef(`init_systemd',`
 	')
 
 	optional_policy(`
+		dev_setattr_generic_usb_dev(kernel_t)
+		dev_delete_generic_usb_dev(kernel_t)
+
 		storage_dev_filetrans_fixed_disk(kernel_t, blk_file)
 		storage_setattr_fixed_disk_dev(kernel_t)
 		storage_create_fixed_disk_dev(kernel_t)
 		storage_delete_fixed_disk_dev(kernel_t)
+
+		storage_setattr_scsi_generic_dev(kernel_t)
+		storage_delete_scsi_generic_dev(kernel_t)
 	')
 ')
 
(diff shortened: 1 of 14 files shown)


policy/modules/system/init.if
```diff
@@ -775,6 +775,26 @@ interface(`init_exec',`
...
+##	entrypoint for the calling domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`init_exec_no_trans',`
```
pebenito, line 793: Dupe of interface above it.

policy/modules/kernel/kernel.te
```diff
@@ -337,6 +337,7 @@ mls_process_set_level(kernel_t)
 selinux_getattr_fs(kernel_t)
 selinux_load_policy(kernel_t)
 
+
```
pebenito, line 340: Unnecessary extra line.

pebenito (changes requested):

policy/modules/kernel/kernel.te
```diff
@@ -390,10 +390,14 @@ ifdef(`init_systemd',`
 	')
 
 	optional_policy(`
+		dev_getattr_generic_usb_dev(kernel_t)
```
pebenito, line 393: The `getattr` is also provided by `dev_delete_generic_usb_dev(kernel_t)`; see `delete_chr_file_perms`.

policy/modules/kernel/kernel.te
pebenito: Still showing an extra newline.

policy/modules/system/init.if
pebenito: Still showing the unnecessary interface.

policy/modules/kernel/kernel.te
0xC0ncord: Sorry -- these snuck back in while juggling this PR between multiple workstations.

pebenito (approved):

PR #743, opened 2023-12-20
Dbus fixes

dsugar100:


policy/modules/services/dbus.if
```diff
@@ -76,6 +76,8 @@ template(`dbus_role_template',`
 	domain_entry_file($1_dbusd_t, dbusd_exec_t)
 	ubac_constrained($1_dbusd_t)
 
+	type $1_dbusd_tmpfs_t;
```
pebenito, line 79: Need `files_tmpfs_file()` too.

pebenito (changes requested):

pebenito (approved):

PR #744, opened 2023-12-25
filesystem:Add type contexts and interface for functionfs

quic-kmeng:
When start up adbd by adb initscript, there's a command like: mount -o uid=2000,gid=2000 -t functionfs adb /dev/usb-ffs/adb

will cause below deny because lack of functionfs related contexts.

avc:  denied  { mount } for  pid=346 comm="mount" name="/" dev="functionfs" ino=17700 scontext=system_u:system_r:mount_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=filesystem permissive=1

### DIFF
--- a/policy/modules/kernel/devices.te
+++ b/policy/modules/kernel/devices.te
@@ -360,6 +360,7 @@ files_mountpoint(usbfs_t)
 fs_pseudo_type(usbfs_t)
 genfscon usbfs / gen_context(system_u:object_r:usbfs_t,s0)
 genfscon usbdevfs / gen_context(system_u:object_r:usbfs_t,s0)
+genfscon functionfs / gen_context(system_u:object_r:usbfs_t,s0)
 
 #
 # usb_device_t is the type for /dev/bus/usb/[0-9]+/[0-9]+


policy/modules/kernel/filesystem.te
```diff
@@ -105,6 +105,11 @@ type configfs_t;
 fs_type(configfs_t)
 genfscon configfs / gen_context(system_u:object_r:configfs_t,s0)
 
+type functionfs_t;
+fs_type(functionfs_t)
+files_mountpoint(functionfs_t)
+genfscon functionfs / gen_context(system_u:object_r:functionfs_t,s0)
```
pebenito, line 111: I think we want to use the existing `usbfs_t` instead of creating a new type.

pebenito (changes requested):

policy/modules/kernel/filesystem.te
quic-kmeng: hi pebenito, but our mount point type is functionfs, usbfs_t can't be labeled for this mount type, do you means we need to change mount from functionfs to usbfs?

quic-kmeng:
Hi @pebenito , I changed that genfscon usbfs_t for functionfs mount type, and tested on our project, usbfs_t related rules could meet our changes.

PR #759, opened 2024-02-22
Minimal Debian system updates

cgzones:


policy/modules/system/init.if
```diff
@@ -3752,6 +3752,26 @@ interface(`init_reload_all_units',`
...
+## </param>
+#
+interface(`init_getattr_all_unit_files',`
+	gen_require(`
+		attribute systemdunit;
+	')
+
+	list_dirs_pattern($1, systemdunit, systemdunit)
+	getattr_files_pattern($1, systemdunit, systemdunit)
+	getattr_lnk_files_pattern($1, systemdunit, systemdunit)
```
pebenito, line 3772: It seems like the symlink access here should actuallybe a read instead..

policy/modules/system/sysnetwork.if
```diff
@@ -489,6 +489,7 @@ interface(`sysnet_create_config',`
 	')
 
 	files_search_etc($1)
+	allow $1 net_conf_t:dir create_dir_perms;
```
pebenito, line 492: Perhaps this should be `{ create_dir_perms add_entry_dir_perms }`.

pebenito (commented):
I think this could be merged as-is but have a couple thoughts.

policy/modules/system/init.if
cgzones: done

policy/modules/system/selinuxutil.te
```diff
@@ -249,6 +249,7 @@ read_lnk_files_pattern(newrole_t, default_context_t, default_context_t)
 
 kernel_read_system_state(newrole_t)
 kernel_read_kernel_sysctls(newrole_t)
+kernel_getattr_proc(newrole_t)
```
cgzones, line 252: or should this be dontaudit'ed?

pebenito: I go back and forth on this. Strictly speaking, dontaudit it, if it isn't required.

PR #761, opened 2024-02-23
systemd: allow systemd-logind to inherit local login file descriptors

yizhao1:
Fix reboot timeout error:
$ reboot
Failed to set wall message, ignoring: Failed to activate service 'org.freedesktop.login1': timed out (service_start_timeout=25000ms) Call to Reboot failed: Failed to activate service 'org.freedesktop.login1': timed out (service_start_timeout=25000ms)

avc:  denied  { use } for  pid=287 comm="systemd-logind" path="anon_inode:[pidfd]" dev="anon_inodefs" ino=1044 scontext=system_u:system_r:systemd_logind_t
tcontext=system_u:system_r:local_login_t tclass=fd permissive=0

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -961,6 +961,7 @@ init_stop_system(systemd_logind_t)
 miscfiles_read_localization(systemd_logind_t)
 
 locallogin_read_state(systemd_logind_t)
+locallogin_use_fds(systemd_logind_t)
 
 seutil_libselinux_linked(systemd_logind_t)
 seutil_read_default_contexts(systemd_logind_t)


0xC0ncord:
Looks good to me, but maybe we should have a comment here indicating this is a `pidfd` that is being inherited as the hooks in the kernel for these may change in the future. I don't know if that's still the case though. @pebenito would you be able to advise?

pebenito:
There's also a pidfs coming for pidfds (see #762 )

brauner:
This should allow you to do proper labeling finally.

pebenito:
> Looks good to me, but maybe we should have a comment here indicating this is a `pidfd` that is being inherited as the hooks in the kernel for these may change in the future. I don't know if that's still the case though. @pebenito would you be able to advise?

With pidfs incoming, I think it will be a bit easier, as we can provide the fd access and pidfs access, in an interface that has "pidfd" in the name, for example, `locallogin_use_pidfd(systemd_logind_t`

PR #769, opened 2024-04-04
systemd: allow notify client to stat socket

cgzones:
Caused by the latest openssh version in Debian sid:

    AVC avc:  denied  { getattr } for  pid=13544 comm="sshd" path="/run/systemd/notify" dev="tmpfs" ino=286 scontext=system_u:system_r:sshd_t:s0 tcontext=system_u:object_r:systemd_runtime_notify_t:s0 tclass=sock_file permissive=0

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -322,7 +322,7 @@ interface(`systemd_write_notify_socket',`
 
 	init_list_runtime($1)
 	init_unix_stream_socket_sendto($1)
-	allow $1 systemd_runtime_notify_t:sock_file write;
+	allow $1 systemd_runtime_notify_t:sock_file write_sock_file_perms;
 ')
 
 ######################################


policy/modules/system/systemd.if
```diff
@@ -322,7 +322,7 @@ interface(`systemd_write_notify_socket',`
 
 	init_list_runtime($1)
 	init_unix_stream_socket_sendto($1)
-	allow $1 systemd_runtime_notify_t:sock_file write;
+	allow $1 systemd_runtime_notify_t:sock_file { getattr write };
```
pebenito, line 325: Please use `write_sock_file_perms` instead.

pebenito (changes requested):

policy/modules/system/systemd.if
cgzones: This would include `open`, is that desired?

pebenito: I don't have a concern here.

cgzones: Adjusted.

pebenito (approved):

PR #778, opened 2024-05-06
Various fixes

0xC0ncord:
More various fixes across the board.

### DIFF
--- a/policy/modules/services/asterisk.te
+++ b/policy/modules/services/asterisk.te
@@ -110,6 +110,7 @@ corenet_udp_bind_sip_port(asterisk_t)
 corenet_sendrecv_generic_server_packets(asterisk_t)
 corenet_tcp_bind_generic_port(asterisk_t)
 corenet_udp_bind_generic_port(asterisk_t)
+corenet_udp_bind_all_unreserved_ports(asterisk_t)
 corenet_dontaudit_udp_bind_all_ports(asterisk_t)
 
 corenet_sendrecv_jabber_client_client_packets(asterisk_t)
--- a/policy/modules/services/matrixd.te
+++ b/policy/modules/services/matrixd.te
@@ -20,6 +20,16 @@ gen_tunable(matrix_allow_federation, true)
 ## </desc>
 gen_tunable(matrix_postgresql_connect, false)
 
+## <desc>
+##  <p>
+##  Determine whether Matrixd is allowed to bind all
+##  TCP ports. This is intended for more complex Matrix
+##	server configurations (e.g. Synapse workers) and may
+##	be used in lieu of manually labeling each port.
+##  </p>
+## </desc>
+gen_tunable(matrix_bind_all_unreserved_tcp_ports, false)
+
 type matrixd_t;
 type matrixd_exec_t;
 init_daemon_domain(matrixd_t, matrixd_exec_t)
@@ -117,7 +127,11 @@ tunable_policy(`matrix_postgresql_connect',`
 	postgresql_tcp_connect(matrixd_t)
 ')
 
+tunable_policy(`matrix_bind_all_unreserved_tcp_ports',`
+	corenet_tcp_bind_all_unreserved_ports(matrixd_t)
+')
+
 optional_policy(`
 	apache_search_config(matrixd_t)
 ')
- 
+
--- a/policy/modules/services/podman.te
+++ b/policy/modules/services/podman.te
@@ -39,6 +39,12 @@ userdom_user_application_domain(podman_user_conmon_t, conmon_exec_t)
 
 allow podman_t podman_conmon_t:process setsched;
 
+kernel_rw_vm_overcommit_sysctl(podman_t)
+
+init_use_fds(podman_t)
+init_setattr_stream_sockets(podman_t)
+init_stream_connect(podman_t)
+
 # for --network=host
 selinux_getattr_dirs(podman_t)
 selinux_mounton_dirs(podman_t)
@@ -67,8 +73,10 @@ podman_spec_rangetrans_conmon(podman_t, s0)
 ifdef(`init_systemd',`
 	init_dbus_chat(podman_t)
 	init_setsched(podman_t)
+	init_get_system_status(podman_t)
 	init_start_system(podman_t)
 	init_stop_system(podman_t)
+	init_reload(podman_t)
 
 	# containers get created as systemd transient units
 	init_get_transient_units_status(podman_t)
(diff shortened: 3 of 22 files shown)


policy/modules/services/asterisk.te
```diff
@@ -110,7 +110,7 @@ corenet_udp_bind_sip_port(asterisk_t)
 corenet_sendrecv_generic_server_packets(asterisk_t)
 corenet_tcp_bind_generic_port(asterisk_t)
 corenet_udp_bind_generic_port(asterisk_t)
-corenet_dontaudit_udp_bind_all_ports(asterisk_t)
+corenet_udp_bind_all_unreserved_ports(asterisk_t)
```
pebenito, line 113: I think we want to keep the original line, so if it tries to bind to one of the defined ports, it will remain silent on audits.

policy/modules/services/matrixd.te
```diff
@@ -77,6 +77,7 @@ corecmd_exec_shell(matrixd_t)
 
 corenet_tcp_bind_generic_node(matrixd_t)
 corenet_tcp_bind_http_port(matrixd_t)
+corenet_tcp_bind_all_unreserved_ports(matrixd_t)
```
pebenito, line 80: What's the purpose for this?

policy/modules/services/podman.te
```diff
@@ -39,6 +39,12 @@ userdom_user_application_domain(podman_user_conmon_t, conmon_exec_t)
 
 allow podman_t podman_conmon_t:process setsched;
 
+kernel_rw_vm_overcommit_sysctl(podman_t)
+
+init_use_fds(podman_t)
+init_setattr_stream_sockets(podman_t)
+init_write_runtime_socket(podman_t)
```
pebenito, line 46: Seems like it would be related to `connectto`/`sendto`.  It would be best to use that interface instead.

policy/modules/system/authlogin.te
```diff
@@ -128,6 +128,8 @@ dev_read_urand(chkpwd_t)
 dev_search_sysfs(chkpwd_t)
 
 files_read_etc_files(chkpwd_t)
+# to read /etc/machine-id
+files_read_etc_runtime_files(chkpwd_t)
```
pebenito, line 132: This is presumably related to pam_systemd; if so, we should add access in the related interface or call it.

policy/modules/system/init.te
```diff
@@ -630,6 +630,11 @@ ifdef(`init_systemd',`
 		fs_rw_rpc_named_pipes(initrc_t)
 	')
 
+	optional_policy(`
+		# use sshd pidfds
+		ssh_use_fds(init_t)
```
pebenito, line 635: Please change the interface to `ssh_use_sshd_pidfds` so it's clear that it's sshd and pidfds.

policy/modules/system/sysnetwork.te
```diff
@@ -380,6 +380,10 @@ ifdef(`distro_ubuntu',`
 	')
 ')
 
+optional_policy(`
+	container_read_runtime_files(ifconfig_t)
```
pebenito, line 384: How is this used? It seems unwise to allow ifconfig to read low integrity files.

policy/modules/system/userdomain.if
```diff
@@ -648,6 +649,9 @@ template(`userdom_common_user_template',`
 
 	allow $1_t unpriv_userdomain:fd use;
 
+	# for Kanidm, which manages user home dirs with a symlink and a UUID for each user
+	allow $1_t user_home_dir_t:lnk_file read_lnk_file_perms;
```
pebenito, line 653: This should be added to `userdom_ro_home_role()` and `userdom_manage_home_role()` instead. Home dirs might have a symlink for many reasons.

pebenito (changes requested):

policy/modules/services/matrixd.te
0xC0ncord: Allow me to double-check, but I recall I needed this access or else images and other media would fail to send over federation.

policy/modules/system/init.te
0xC0ncord: Sure.

I know this is an ongoing discussion, but are there still plans to address `pidfd`s different from the `fd` class? I want to know how to handle these pidfds in the policy going forward and whether we intend to just leave the access as `fd` for now until support lands in the kernel, then we can change these to `pidfd` or however that change needs to happen.

policy/modules/system/sysnetwork.te
0xC0ncord: Allow me to double-check that this is still required and provide the AVC, but I found this was needed by netavark which will be invoked by podman to set up the container network. At some point during container startup it looked like the container's network data was being written to `/run`, and netavark would call `ip` to load it.

policy/modules/system/init.te
pebenito: That's why I'd like to have `pidfd` in the interface title. There's the pidfs behavior that still needs to be resolved. I think it's unlikely pidfds will be handled as a separate class, but having `pidfd` interfaces will preserve the flexibility once all of the dust settles.

0xC0ncord: Makes sense. Thanks for the clarification. :)

policy/modules/services/matrixd.te
0xC0ncord: OK this turned out to be because I am using Synapse workers in my setup, and some workers need to bind to a different HTTP port and may (optionally) expose Prometheus metrics on other ports. I'll instead make this a tunable for the `matrixd` policy.

policy/modules/system/sysnetwork.te
0xC0ncord: The original AVC produced when I hit this is as follows:
```
type=PROCTITLE msg=audit(1679678391.757:560457): proctitle=69700061
type=SYSCALL msg=audit(1679678391.757:560457): arch=c000003e syscall=257 success=no exit=-13 a0=ffffff9c a1=74dfcdafcd10 a2=0 a3=0 items=0 ppid=3165222 pid=3568246 auid=1000 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=pts3 ses=30 comm="ip" exe="/usr/bin/ip" subj=staff_u:sysadm_r:ifconfig_t:s0 key=(null)
type=AVC msg=audit(1679678391.757:560457): avc:  denied  { read } for  pid=3568246 comm="ip" name="netns-6146bd2e-66b7-c06b-89d8-f992038d10bd" dev="tmpfs" ino=11056 scontext=staff_u:sysadm_r:ifconfig_t:s0 tcontext=system_u:object_r:container_runtime_t:s0 tclass=file permissive=0
```
However I have not been able to reproduce it, even after doing a `podman system reset` and restarting all containers on the host I saw this.

This might have instead been from trying to run `ip` directly and it trying to display information from other containers, but I wasn't able to reproduce this with that behavior either.

I'll drop this for now and we can revisit if it manifests again.

pebenito (approved):

PR #779, opened 2024-06-04
Fixes for dhcpcd and newrole

yizhao1:


policy/modules/system/authlogin.if
```diff
@@ -938,6 +938,24 @@ interface(`auth_setattr_faillog_files',`
...
+## <summary>
+##	Search faillock dirs (/run/faillock).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`auth_search_faillog_dirs',`
```
pebenito, line 951: Please remove the `_dirs` from the name, since `search` implies dirs already. Also move the interface above `auth_append_faillog` (line 847)

policy/modules/system/sysnetwork.te
```diff
@@ -280,6 +281,11 @@ optional_policy(`
 	xen_append_log(dhcpc_t)
 ')
 
+optional_policy(`
+	udev_search_runtime(dhcpc_t)
+	udev_read_runtime_files(dhcpc_t)
+')
```
pebenito, line 287: The `udev_search_runtime()` is a subset of `udev_read_runtime_files()`, so it isn't necessary.

Also this block should go above the userdom optional block, on line 269.

pebenito (changes requested):

pebenito (approved):

PR #783, opened 2024-06-16
C-005 Cleanup (Reorder perms and classes)

freedom1b2830:
I decided to start clearing the warnings, even though they are not significant

Signed-off-by: freedom1b2830 <freedom1b2830@gmail.com>

### DIFF
--- a/doc/example.te
+++ b/doc/example.te
@@ -22,7 +22,7 @@ files_tmp_file(myapp_tmp_t)
 # Myapp local policy
 #
 
-allow myapp_t myapp_log_t:file { read_file_perms append_file_perms };
+allow myapp_t myapp_log_t:file { append_file_perms read_file_perms };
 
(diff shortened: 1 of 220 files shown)


pebenito:
Did you use sediff to verify there were no changes in the final policy?

freedom1b2830:
I didn't know about checking with sediff.
launched it with old and new policy.33, no output.
What arguments should I run sediff with?

pebenito:
> I didn't know about checking with sediff. launched it with old and new policy.33, no output. What arguments should I run sediff with?

No output means there are no changes.

pebenito:
It took some time, but I manually reviewed all the changes and it look good.

I've added sediff to the GitHub checks; would you please rebase your change to `main` so it will run the new checks? I'd like extra confirmation since the checks run in multiple combinations of the build options.

PR #786, opened 2024-06-27
Initial policy for haproxy

0xC0ncord:
Add an initial policy for haproxy, a TCP/HTTP reverse proxy for high availability environments.

### DIFF
--- a/policy/modules/services/haproxy.te
+++ b/policy/modules/services/haproxy.te
@@ -1,0 +1,121 @@
+policy_module(haproxy)
+
+########################################
+#
+# Declarations
+#
+
+## <desc>
+##	<p>
+##	Determine whether haproxy can bind to
+##	all TCP ports.
+##	</p>
+## </desc>
+gen_tunable(haproxy_bind_all_tcp_ports, false)
+
+## <desc>
+##	<p>
+##	Determine whether haproxy can bind to
+##	kubernetes ports (typically 6443/tcp).
+##	</p>
+## </desc>
+gen_tunable(haproxy_bind_kubernetes_port, false)
+
+## <desc>
+##	<p>
+##	Determine whether haproxy can connect to
+##	all TCP ports.
+##	</p>
+## </desc>
+gen_tunable(haproxy_connect_all_tcp_ports, false)
+
+## <desc>
+##	<p>
+##	Determine whether haproxy can connect to
+##	kubernetes ports (typically 6443/tcp).
+##	</p>
+## </desc>
+gen_tunable(haproxy_connect_kubernetes_port, false)
+
+type haproxy_t;
+type haproxy_exec_t;
+init_daemon_domain(haproxy_t, haproxy_exec_t)
+
+type haproxy_conf_t;
+files_config_file(haproxy_conf_t)
+
+type haproxy_initrc_exec_t;
+init_script_file(haproxy_initrc_exec_t)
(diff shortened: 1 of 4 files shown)


policy/modules/services/haproxy.te
```diff
@@ -0,0 +1,121 @@
...
+type haproxy_tmpfs_t;
+files_tmpfs_file(haproxy_tmpfs_t)
+
+########################################
+#
+# haproxy local policy
+#
+
+allow haproxy_t self:process { getsched setrlimit signal };
+allow haproxy_t self:capability { kill net_admin net_bind_service setuid setgid };
```
pebenito, line 65: The corenet bind interfaces should be providing `net_bind_service` as needed, so I'd prefer not to have it as an explicit allow. If it's not happening, then there is a bug in corenet generation scripts.

policy/modules/services/haproxy.te
```diff
@@ -0,0 +1,121 @@
...
+dev_dontaudit_read_sysfs(haproxy_t)
+
+kernel_read_kernel_sysctls(haproxy_t)
+kernel_read_state(haproxy_t)
+kernel_read_system_state(haproxy_t)
+
+auth_use_nsswitch(haproxy_t)
+
+miscfiles_read_generic_certs(haproxy_t)
+
```
pebenito, line 86: Extra newline

policy/modules/services/haproxy.te
```diff
@@ -0,0 +1,121 @@
...
+manage_sock_files_pattern(haproxy_t, haproxy_runtime_t, haproxy_runtime_t)
+files_runtime_filetrans(haproxy_t, haproxy_runtime_t, { dir file sock_file })
+
+mmap_manage_files_pattern(haproxy_t, haproxy_tmpfs_t, haproxy_tmpfs_t)
+fs_tmpfs_filetrans(haproxy_t, haproxy_tmpfs_t, file)
+
+create_dirs_pattern(haproxy_t, haproxy_log_t, haproxy_log_t)
+create_files_pattern(haproxy_t, haproxy_log_t, haproxy_log_t)
+append_files_pattern(haproxy_t, haproxy_log_t, haproxy_log_t)
+logging_log_filetrans(haproxy_t, haproxy_log_t, { dir file })
```
pebenito, line 105: Needs to move to line 70 and put log access block after conf access.

testing/sechecker.ini
```diff
@@ -254,6 +254,7 @@ exempt_source = arpwatch_t
                 drbd_t
                 fcoemon_t
                 firewalld_t
+                haproxy_t
```
pebenito, line 257: Please put a comment for why/how haproxy needs net_admin.

pebenito (changes requested):

testing/sechecker.ini
0xC0ncord: I found that haproxy is still functioning in my setup if I `dontaudit` the access, so this won't be needed anymore. :)

policy/modules/services/haproxy.te
0xC0ncord: Neat. Frankly I never knew these interfaces implied that access. I tested without `net_bind_service` explicitly given and it still has that access, so the interfaces seem to be working as intended.

pebenito (approved):

PR #787, opened 2024-06-27
Various fixes

0xC0ncord:
More various fixes, of note:

- add a couple missing PIDFD access for init on local login and sshd
- fixes for postfix alongside dovecot and SASL
- add some missing read accesses for node_exporter
- allow sudo to send a SIGWINCH to su when performing `sudo su`
- allow systemd-logind to read cgroup state of sudo
- allow systemd user managers to read container configs, e.g. if the user has set up user containers via the systemd generator
- allow containers to execute their own tmpfs files
- fixes for asterisk


0xC0ncord:
For the systemd user manager containers access, I have so far only added the template and the access needed for the systemd generator to fire, but I haven't tested actually running a container just yet. Any permissions still needed will go there.

policy/modules/services/sasl.if
```diff
@@ -38,6 +38,24 @@ interface(`sasl_read_keytab',`
...
+## <summary>
+##	Memory map SASL keytab files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`sasl_map_keytab',`
```
pebenito, line 51: I'd prefer to get away from interfaces that only provide `map`, so in this case having a `sasl_mmap_read_keytab` would be better.

policy/modules/services/fail2ban.te
```diff
@@ -42,6 +42,8 @@ allow fail2ban_t self:fifo_file rw_fifo_file_perms;
 allow fail2ban_t self:unix_stream_socket { accept connectto listen };
 allow fail2ban_t self:tcp_socket { accept listen };
 
+corenet_tcp_bind_generic_node(fail2ban_t)
```
pebenito, line 45: Please move down with other corenet calls. Also, I don't see any corenet bind calls for the port(s), so there may still be missing permissions.

policy/modules/system/systemd.te
```diff
@@ -1031,6 +1031,12 @@ userdom_relabelto_user_runtime_dirs(systemd_logind_t)
 userdom_setattr_user_ttys(systemd_logind_t)
 userdom_use_user_ttys(systemd_logind_t)
 
+ifdef(`init_systemd',`
```
pebenito, line 1034: Please move the block after the distro one below.

policy/modules/system/systemd.if
```diff
@@ -519,6 +519,29 @@ template(`systemd_user_manager_dbus_chat',`
...
+##	 all the permissions necessary to run user
+##	 containers.
+## </summary>
+## <param name="prefix">
+##	<summary>
+##	Prefix for the user domain.
+##	</summary>
+## </param>
+#
+template(`systemd_user_use_containers',`
```
pebenito, line 534: I'm struggling to see why this is a systemd interfaces. Perhaps this should be held back until it has more access(?)

policy/modules/services/container.te
```diff
@@ -317,6 +317,7 @@ allow container_domain container_ro_file_t:sock_file read_sock_file_perms;
 fs_tmpfs_filetrans(container_domain, container_tmpfs_t, { dir file fifo_file lnk_file sock_file })
 manage_dirs_pattern(container_domain, container_tmpfs_t, container_tmpfs_t)
 mmap_manage_files_pattern(container_domain, container_tmpfs_t, container_tmpfs_t)
+exec_files_pattern(container_domain, container_tmpfs_t, container_tmpfs_t)
```
pebenito, line 320: Is the full `execute_no_trans` needed or only `execute`?

pebenito (changes requested):

policy/modules/services/fail2ban.te
0xC0ncord: Good point about possibly more missing permissions. I went looking through my audit logs to see what exactly triggered this, but couldn't find it. I think I'd prefer to drop this for now in that case.

policy/modules/system/systemd.if
0xC0ncord: I put this template in the systemd module since the access will be granted to systemd user managers on a per-role basis, e.g. `staff_systemd_t` can create containers but `user_systemd_t` cannot.

However, I think I'll drop this for now and revisit it once it's more complete as the implementation may change.

policy/modules/services/container.te
0xC0ncord: Yes, only `execute` is needed. Changing this accordingly.

pebenito: In that case, you want `mmap_exec_files_pattern` instead of `can_exec`.

pebenito (approved):

PR #793, opened 2024-07-05
sshd: label sshd-session as sshd_exec_t

0xC0ncord:
OpenSSH 9.8 splits out much of the session code from the main sshd binary into a new sshd-session binary. Allow the sshd server to execute this binary by labeling it as sshd_exec_t.

### DIFF
--- a/policy/modules/services/ssh.fc
+++ b/policy/modules/services/ssh.fc
@@ -8,6 +8,7 @@ HOME_DIR/\.ssh(/.*)?			gen_context(system_u:object_r:ssh_home_t,s0)
 /usr/bin/ssh-keygen		--	gen_context(system_u:object_r:ssh_keygen_exec_t,s0)
 /usr/bin/sshd			--	gen_context(system_u:object_r:sshd_exec_t,s0)
 
+/usr/lib/misc/sshd-session	--	gen_context(system_u:object_r:sshd_exec_t,s0)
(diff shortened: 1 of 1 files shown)


pebenito:
I'm not clear what specifically what the session parts are. If the login/password checking and SELinux login context lookup+setexeccon() are separated from the networking code, it could be worth making a more significant policy revision.

0xC0ncord:
From the [changelog](https://www.openssh.com/txt/release-9.8):

* [sshd(8)](https://man.openbsd.org/sshd.8): the server has been split into a listener binary, [sshd(8)](https://man.openbsd.org/sshd.8),
   and a per-session binary "sshd-session". This allows for a much
   smaller listener binary, as it no longer needs to support the SSH
   protocol. As part of this work, support for disabling privilege
   separation (which previously required code changes to disable) and
   disabling re-execution of [sshd(8)](https://man.openbsd.org/sshd.8) has been removed. Further
   separation of sshd-session into additional, minimal binaries is
   planned for the future.

In short, historically the sshd binary would fork a copy of itself when a user logs into a system. This still happens, but the forked binary is now a separate one which lacks any privileged code.

> If the login/password checking and SELinux login context lookup+setexeccon() are separated from the networking code, it could be worth making a more significant policy revision.

I haven't looked into all the exact behavior that `sshd-session` does yet, but if this is the case I agree. However, I strongly recommend we merge this for now and we can make a Github issue to track a possible larger policy revision. Without this change, ssh login is completely broken with OpenSSH 9.8 on refpolicy.

pebenito (approved):

PR #795, opened 2024-07-12
locallogin: remove duplicate interface

yizhao1:
The contents of interface locallogin_use_fds and interface locallogin_use_pidfds are the same, remove the latter.


0xC0ncord:
This is intentional because `pidfd` usage in the policy may change in the future. We want to keep `pidfd`s separate from `fd`s when that happens.

yizhao1:
Thanks for your explanation.

PR #798, opened 2024-07-18
Expand cgroup pressure file coverage

matt-sheets:
This commit will expand unique labeling for cgroup's memory.pressure file to also include io.pressure and cpu.pressure.  As such, the label name has been updated to cgroup_pressure_t to reflect this.

File transition has also been expanded to include udev_t and systemd_user_session_type.


pebenito:
I will need confirmation that the labeling holds up on a supported distro or a resolution to the issues discussed in this thread https://lore.kernel.org/selinux/d3c9593d-fc73-4016-b057-57706e94d170@ieee.org/. Otherwise, I'm reluctant to expand broken policy.

matt-sheets:
It looks like the issue that pebentio pointed out is still and issue.  I tested this on FC workstation 40 and only the pressure files related to my user session got correctly labeled.

We do want this change, but I agree that it should only be merged once the existing memory.pressure labeling has been resolved.

PR #801, opened 2024-08-09
Various fixes

0xC0ncord:
A few various fixes, notably:
- let `podman auto-update` restart containers spawned by quadlet
- allow haproxy to be run interactively
- allow kubelet to create subPath directories on not yet relabeled persistent volumes
- allow containers to read generic certs


policy/modules/services/container.te
```diff
@@ -405,6 +412,10 @@ tunable_policy(`container_read_public_content',`
 	miscfiles_watch_public_dirs(container_domain)
 ')
 
+tunable_policy(`container_read_generic_certs',`
+	miscfiles_read_generic_certs(container_domain)
```
pebenito, line 416: I would be ok this being unconditional. Is there something I'm missing?

0xC0ncord: I think it is probably OK to allow it, but I had a look at what `container-selinux` is doing and they are also using a tunable there.

I think this is a bit of a fringe use case as a container doesn't normally need to access certs on the host unless you explicitly bind-mount them inside the container.

pebenito: Certs are public anyway. Since these are the generic ones, I really think this should be unconditional.

pebenito (approved):

PR #802, opened 2024-08-09
Add policy for KubeVirt

0xC0ncord:
Add policy rules required for KubeVirt and Multus.

Add a new container type `container_kvm_t` which does not have the extra rules a generic `container_t` would have, but rather just enough to be able to run a virtual machine. This also means it needs RW access to `/dev/kvm`. A type with the same name also exists in `container-selinux` which to my knowledge is similar in functionality. However, typical KubeVirt installations with `container-selinux` (e.g. on OpenShift) will use an external type provided by a CIL policy module (`virt_launcher.process`). To use this policy with KubeVirt we need to apply the following config to the KubeVirt installation on the cluster:
```yaml
apiVersion: kubevirt.io/v1
kind: KubeVirt
metadata:
  name: kubevirt
  namespace: kubevirt
spec:
  configuration:
    developerConfiguration:
      featureGates:
      - DisableCustomSELinuxPolicy
    selinuxLauncherType: container_kvm_t
```
The `DisableCustomSELinuxPolicy` tells KubeVirt not to install its custom SELinux policy (which will be denied under refpolicy) and `selinuxLauncherType` tells KubeVirt to use the `container_kvm_t` type for VM pods.

Add various rules required for Multus to function, which is a container networking plugin for running other container network plugins. This is useful for KubeVirt where a the KVM container may need to join the VM to multiple networks on the host, e.g. to a bridge or masqueraded so that it can access networks outside of the kubernetes internal pod network.

Add a few rules required for data volume (DV) manipulation on `rook-ceph`. KubeVirt provides a few ways to upload VM disk images for use by VMs in the cluster. In this case a pod is created with the volume mounted as a block device within the container which will be written to as the VM disk image is uploaded.

Note that we also add a `container_tmp_t` type here which is used for very short-lived files written by Multus on the host `/tmp`. Despite the name, `spc_t` is the only domain with this access currently.


testing/sechecker.ini
```diff
@@ -241,6 +241,7 @@ exempt_source = arpwatch_t
                 chronyd_t           # Conditional access (chronyd_hwtimestamp)
                 condor_startd_t
                 container_engine_t
+                container_kvm_t
```
pebenito, line 244: Please add a comment to explain this.

pebenito:
I can't remember if this will work, but did you try making `virt_launcher.process` an alias of `container_kvm_t`?

0xC0ncord:
> I can't remember if this will work, but did you try making `virt_launcher.process` an alias of `container_kvm_t`?

Trying this, it seems not unfortunately.

selint gives this error:
```
policy/modules/services/container.te:185: (F): syntax error, unexpected PERIOD, expecting SEMICOLON (F-001)
  1[8](https://github.com/0xC0ncord/refpolicy/actions/runs/10389184265/job/28766614430#step:7:9)5 | typealias container_kvm_t alias virt_launcher.process;
      |                                              ^
policy/modules/services/container.te:185: (F): Error: Invalid statement (F-001)
  185 | typealias container_kvm_t alias virt_launcher.process;
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Error: Failed to parse files
```

Ignoring the above, trying to build the policy results in:
```
Compiling refpolicy container.mod module
m4 -E -E -D enable_ubac=true -D mls_num_sens=16 -D mls_num_cats=1024 -D mcs_num_cats=1024 -s support/divert.m4 policy/support/file_patterns.spt policy/support/ipc_patterns.spt policy/support/loadable_module.spt policy/support/misc_macros.spt policy/support/misc_patterns.spt policy/support/mls_mcs_macros.spt policy/support/obj_perm_sets.spt support/undivert.m4 tmp/generated_definitions.conf tmp/all_interfaces.conf policy/modules/services/container.te support/fatal_error.m4 > tmp/container.tmp
/usr/bin/checkmodule -m tmp/container.tmp -o tmp/container.mod
policy/modules/services/container.te:185:ERROR 'type alias identifier virt_launcher.process may not contain periods' at token ';' on line 4689:
# try maintaining compatibility with kubevirt's external policy
typealias container_kvm_t alias virt_launcher.process;
```

pebenito (approved):

PR #812, opened 2024-10-01
Make mta optional in container policy

dsugar100:
Don't need to have mta policy loaded when loading container policy.


pebenito (approved):

pebenito:
@dsugar100 please upload your gpg public key to your github profile since you're signing commits.

PR #816, opened 2024-10-04
validate-appconfig.py: Add further validations.

pebenito:
New validations:
* default_type
* failsafe_context
* default_contexts
* seuser default_contexts

Additionally add GitHub annotations to logging.


pebenito:
@perfinion or @0xC0ncord can you try this out? I imagine binary distros wouldn't have a problem with (optionally) having setools during build, but as Gentoo is source...

PR #818, opened 2024-10-08
add netlink_xperm policy capability and nlmsg permission definitions

stephensmalley:
Add the netlink_xperm policy capability and nlmsg permission definitions. The policy capability is commented-out/disabled by default. Enabling it will require the next release version of libsepol (3.8+) and the next release version of the kernel (6.13+). When the policy capability is enabled, the kernel switches from checking the current nlmsg_read/write/... permissions to always checking a single nlmsg permission with the nlmsg_type value as an extended permission.

nlmsg permission is not added to the *_netlink_socket_perms macros to avoid unwittingly allowing all netlink message types when/if this policy capability is enabled.

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -600,6 +600,7 @@ inherits socket
 {
 	nlmsg_read
 	nlmsg_write
+	nlmsg
 }
 
 class obsolete_netlink_firewall_socket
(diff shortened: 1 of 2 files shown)


pebenito:
SETools build failure is unrelated. Merging.

Issue #819, opened 2024-10-08
Error in validate-appconfig.py


dsugar100:
I'm seeing the following error when running validate-appconfig.py.  I'm on a RHEL9 system.

Validating clip file contexts.                                                                                                                        
/sbin/setfiles -q -c tmp/policy.bin tmp/all_mods.fc
Validating clip appconfig.                                                                                                                            
python3 -bb -t -t -E -W error support/validate-appconfig.py -c /usr/bin/chkcon -l -s -v -x tmp/appconfig tmp/policy.bin
Traceback (most recent call last):                                         
  File "/builddir/build/BUILD/clip-selinux-policy-9.1.0.5/support/validate-appconfig.py", line 42, in <module>                                            class ContextValidator:                                                                                                                             
File "/builddir/build/BUILD/clip-selinux-policy-9.1.0.5/support/validate-appconfig.py", line 46, in ContextValidator                                    def __init__(self, /, policy_path: str | None = None, *,                                                                                          
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'                                                                                   
make: *** [Rules.modular:228: validate] Error 1                                                                                                       
error: Bad exit status from /var/tmp/rpm-tmp.BOvrtp (%install)                                                                                        
    Bad exit status from /var/tmp/rpm-tmp.BOvrtp (%install)
                                                                           
Ideas?

dsugar100:
RHEL9 has python 3.9.  Testing on Fedora which as python3.11 seems to work ok.

dsugar100:
Ok, looks like the 'str | None' syntax was added in 3.10.  I will see if I can update to work with 3.9.

pebenito:
I think this tool will instead have to be disabled if you're building on Python 3.9; see #816. It pulls in SETools, which requires 3.10.

PR #820, opened 2024-10-08
Changes to support python 3.9 (RHEL9)

dsugar100:
To try to address issue #819


pebenito:
I think this tool will instead have to be disabled if you're building on Python 3.9; see #816. It pulls in SETools, which requires 3.10.

dsugar100:
In #816 it seems to imply that SETools is optional.  If that is the case, it would be nice to still have the rest of this stuff work, even if the SETools isn't available.

pebenito (approved):

PR #823, opened 2024-10-18
files context : few fixes for merged-usr distro_gentoo

PPN-SD:


policy/modules/services/apache.fc
```diff
@@ -49,6 +49,13 @@ HOME_DIR/((www)|(web)|(public_html))(/.*)?/logs(/.*)?		gen_context(system_u:obje
 /usr/bin/httpd(\.worker)?				--	gen_context(system_u:object_r:httpd_exec_t,s0)
 /usr/bin/lighttpd					--	gen_context(system_u:object_r:httpd_exec_t,s0)
 /usr/bin/mongrel_rails					--	gen_context(system_u:object_r:httpd_exec_t,s0)
+
+ifdef(`distro_gentoo',`
+/usr/bin/nginx						--	gen_context(system_u:object_r:httpd_exec_t,s0)
+/usr/bin/php.*-fpm					--	gen_context(system_u:object_r:httpd_exec_t,s0)
+/usr/bin/php-fpm[^/]+					--	gen_context(system_u:object_r:httpd_exec_t,s0)
+')
+
```
pebenito, line 58: The nginx one doesn't need to be in the block; I'm unsure if the php ones should be inside. If the ifdef remains, it should move down after the other `/usr/bin` lines, at line 63.

policy/modules/services/rpc.fc
```diff
@@ -35,3 +35,7 @@
 /run/rpc\.statd(/.*)?	gen_context(system_u:object_r:rpcd_runtime_t,s0)
 /run/rpc\.statd\.pid	--	gen_context(system_u:object_r:rpcd_runtime_t,s0)
 /run/sm-notify\.pid	--	gen_context(system_u:object_r:rpcd_runtime_t,s0)
+
+ifdef(`distro_gentoo',`
+/usr/bin/blkmapd	--	gen_context(system_u:object_r:blkmapd_exec_t,s0)
```
pebenito, line 40: This need not be in an ifdef.

pebenito (changes requested):
@perfinion , @0xC0ncord any other thoughts, in a Gentoo-specific context?

policy/modules/services/apache.fc
pebenito: The general guideline is: if the full path is standard, then it doesn't need an `ifdef`. Refpolicy generally considers `/bin`, `/sbin`, `/usr/bin`, and `/usr/sbin` as equivalent and `nginx` is standard too, so I don't think it should be in an ifdef.

I'm not familiar with php, so I don't know if that file naming is standard or Gentoo-specific to handle multiple versions of php (or for some other reason).

PPN-SD: Ok, thanks for the explanation. So it’s not related to the progress of usrmerge in each distro.
This file-naming seems common pratice : [openSUSE](https://opensuse.pkgs.org/15.6/opensuse-oss-x86_64/php8-fpm-8.2.10-150600.1.6.x86_64.rpm.html)/[Fedora](https://rpmfind.net/linux/RPM/remi/enterprise/9/modular/x86_64/php-fpm-8.3.12-1.el9.remi.x86_64.html)/[Debian](https://packages.debian.org/bookworm/amd64/php8.2-fpm/filelist)

0xC0ncord:
Sorry for the delay.

These changes look good for refpolicy, but I'll have to remember to adjust this for nginx as Gentoo is currently shipping a policy for it that is separate from the apache/httpd policy.

PPN-SD:
> Sorry for the delay.
> 
> These changes look good for refpolicy, but I'll have to remember to adjust this for nginx in Gentoo as we are currently shipping a policy for it that is separate from the apache/httpd policy.

Done, before the memory goes away :-)
https://github.com/gentoo/hardened-refpolicy/pull/4

pebenito:
> Sorry for the delay.
> 
> These changes look good for refpolicy, but I'll have to remember to adjust this for nginx in Gentoo as we are currently shipping a policy for it that is separate from the apache/httpd policy.

How different is this? Apache is pretty heavyweight, so a more trim policy for nginx might be upstreamable.

pebenito (approved):

PR #825, opened 2024-10-22
systemd: Fix systemd_write_notify_socket().

pebenito:
The notify socket is SOCK_DGRAM. See sd_notify(3) NOTES section for details.


pebenito:
Same as #822 but signed commit.

PR #827, opened 2024-10-24
Update mcs.if to instruct caller to "use" mcs_constrained and not "remove"

JaysonHurst:
Fixed deprecated statement in mcs interfaces that instructed caller to 'remove' mcs_constrained()'.


JaysonHurst:
> "remove" isn't necessarily incorrect. We've narrowed the refpolicy use case for MCS down to mainly containers an VMs. Outside of those domains, they would want to remove `mcs_constrained()`.

If it makes sense, then I am fine with it.

PR #828, opened 2024-10-30
Misc updates

cgzones:


policy/modules/kernel/files.if
```diff
@@ -3848,14 +3848,31 @@ interface(`files_dontaudit_read_etc_runtime_files',`
...
+##	in /etc that are dynamically
+##	created on boot, such as mtab.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`files_dontaudit_execuite_etc_runtime_files',`
```
pebenito, line 3871: It should be "`exec`" instead of "`execute`".

policy/modules/system/systemd.te
```diff
@@ -581,11 +581,13 @@ init_read_script_files(systemd_generator_t)
 init_getattr_all_unit_files(systemd_generator_t)
 init_getattr_all_script_files(systemd_generator_t)
 
+# vmware_vsock
+kernel_request_load_module(systemd_generator_t)
```
pebenito, line 585: Nit: please move to the end of the `kernel_*` lines (after `kernel_dontaudit_search_unlabeled`)

policy/modules/system/userdomain.if
```diff
@@ -2532,6 +2532,7 @@ interface(`userdom_manage_user_home_content_files',`
 	')
 
 	manage_files_pattern($1, user_home_t, user_home_t)
+	allow $1 user_home_t:file map;
```
pebenito, line 2535: There is a `mmap_manage_files_pattern()` now.

pebenito (changes requested):
A few minor nits.

Side note: I'd take a patch to add a codespell check in the lint github action, if you're interested.

policy/modules/kernel/files.if
cgzones: changed

policy/modules/system/systemd.te
cgzones: moved

policy/modules/system/userdomain.if
cgzones: used macro

pebenito (approved):

PR #832, opened 2024-11-10
ci: add module storage test 

cgzones:
Add test to verify modular policies can be successfully installed from a module storage.

Also drop obsolete setfiles option.


pebenito:
Looks like we were working on the same thing simultaneously (see #833) but I'd prefer to have the complexity in the CI scripts rather than in the Makefile.

PR #839, opened 2024-11-28
systemd-homed: various fixups

WavyEbuilder:
Commit messages for each commit should explain more, let me know if you have any more questions.

Thanks!

### DIFF
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -109,6 +109,7 @@ HOME_ROOT/(.+)\.home	--	gen_context(system_u:object_r:systemd_homed_storage_t,s0
 /run/systemd/ask-password(/.*)?	gen_context(system_u:object_r:systemd_passwd_runtime_t,s0)
 /run/systemd/ask-password-block(/.*)?	gen_context(system_u:object_r:systemd_passwd_runtime_t,s0)
 /run/systemd/home(/.*)?         gen_context(system_u:object_r:systemd_homed_runtime_t,s0)
+/run/systemd/user-home-mount	-d	gen_context(system_u:object_r:systemd_homed_runtime_work_dir_t,s0)
 /run/systemd/network(/.*)?  gen_context(system_u:object_r:systemd_networkd_runtime_t,s0)
 /run/systemd/notify		-s	gen_context(system_u:object_r:systemd_runtime_notify_t,s0)
 /run/systemd/resolve(/.*)?  gen_context(system_u:object_r:systemd_resolved_runtime_t,s0)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -61,6 +61,9 @@ HOME_DIR/\.config/containers/systemd(/.*)?		gen_context(system_u:object_r:system
 HOME_DIR/\.config/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home_t,s0)
 HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data_home_t,s0)
 
+# homed files
+HOME_ROOT/(.+)\.home	--	gen_context(system_u:object_r:systemd_homed_storage_t,s0)
+
 /usr/lib/systemd/user(/.*)?		gen_context(system_u:object_r:systemd_user_unit_t,s0)
 
 /usr/lib/systemd/system/[^/]*halt.*	--	gen_context(system_u:object_r:power_unit_t,s0)
--- a/policy/modules/system/systemd.fc
+++ b/policy/modules/system/systemd.fc
@@ -62,6 +62,8 @@ HOME_DIR/\.config/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home
 HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data_home_t,s0)
 
 # homed files
+HOME_DIR/\.identity	--	gen_context(system_u:object_r:systemd_homed_record_t,s0)
+HOME_DIR/\.identity-blob(/.*)?	gen_context(system_u:object_r:systemd_homed_record_t,s0)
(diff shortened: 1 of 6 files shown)


policy/modules/system/systemd.if
```diff
@@ -1136,6 +1136,25 @@ interface(`systemd_dbus_chat_homed',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_stream_connect_homed',`
+	gen_require(`
+		type systemd_homed_t;
+	')
+
+	allow $1 systemd_homed_t:unix_stream_socket connectto;
```
pebenito, line 1155: The interface summary says this is a named socket, so there should be `dir` and `sock_file` access too, preferably using `stream_connect_pattern()`.

policy/modules/system/systemd.fc
```diff
@@ -61,6 +61,11 @@ HOME_DIR/\.config/containers/systemd(/.*)?		gen_context(system_u:object_r:system
 HOME_DIR/\.config/systemd(/.*)?		gen_context(system_u:object_r:systemd_conf_home_t,s0)
 HOME_DIR/\.local/share/systemd(/.*)?		gen_context(system_u:object_r:systemd_data_home_t,s0)
 
+# homed files
+HOME_DIR/\.identity	--	gen_context(system_u:object_r:systemd_homed_record_t,s0)
+HOME_DIR/\.identity-blob(/.*)?	gen_context(system_u:object_r:systemd_homed_record_t,s0)
```
pebenito, line 66: Are these intended/required to be inaccessible to the user?

policy/modules/system/systemd.te
```diff
@@ -725,14 +735,21 @@ allow systemd_homework_t systemd_homed_runtime_t:dir manage_dir_perms;
...
+allow systemd_homework_t systemd_homed_runtime_work_dir_t:dir mounton;
 
 allow systemd_homework_t systemd_homed_storage_t:file manage_file_perms;
 files_home_filetrans(systemd_homework_t, systemd_homed_storage_t, file)
 
 allow systemd_homework_t systemd_homed_tmpfs_t:file rw_inherited_file_perms;
 
+# setup luks backed home directories in /run/cryptsetup
+lvm_manage_runtime_files(systemd_homework_t)
+lvm_manage_runtime_dirs(systemd_homework_t)
```
pebenito, line 751: This should be made optional.

policy/modules/system/systemd.te
```diff
@@ -769,6 +786,10 @@ systemd_log_parse_environment(systemd_homework_t)
 
 udev_read_runtime_files(systemd_homework_t)
 
+read_files_pattern(systemd_homework_t, systemd_homed_runtime_work_dir_t, systemd_homed_record_t)
+create_files_pattern(systemd_homework_t, systemd_homed_runtime_work_dir_t, systemd_homed_record_t)
+delete_files_pattern(systemd_homework_t, systemd_homed_runtime_work_dir_t, systemd_homed_record_t)
```
pebenito, line 791: This block should be up just above the machine-id hunk you added.

pebenito (changes requested):
Also some lint issues in the github actions.

policy/modules/system/systemd.te
WavyEbuilder: Got it, will do

policy/modules/system/systemd.te
WavyEbuilder: Will do

policy/modules/system/systemd.if
WavyEbuilder: Will do

policy/modules/system/systemd.fc
WavyEbuilder: I did actually consider that. Looking at accountsservice and the various other things that are working with homed, I think the preferred way of getting account info is from the system userdb stuff. It has sensitive data (like the hashed password for that user) which is why I'm hesitant to make it readable by a user domain, but it is just for their own user. However, I can't see any reason for that information to be read or written to by the user, infact the records are signed so a user cannot write to it anyway. I'm happy to make it readable by the user if you wish however

policy/modules/system/systemd.te
WavyEbuilder: Done

policy/modules/system/systemd.if
WavyEbuilder: Done

policy/modules/system/systemd.if
```diff
@@ -1152,7 +1152,7 @@ interface(`systemd_stream_connect_homed',`
 		type systemd_homed_t;
 	')
 
-	allow $1 systemd_homed_t:unix_stream_socket connectto;
+	stream_connect_pattern($1, systemd_userdbd_runtime_t, systemd_userdbd_runtime_t, systemd_homed_t)
```
pebenito, line 1155: Not all the types are in the `gen_require`.

policy/modules/system/systemd.te
```diff
@@ -747,8 +751,10 @@ files_home_filetrans(systemd_homework_t, systemd_homed_storage_t, file)
 allow systemd_homework_t systemd_homed_tmpfs_t:file rw_inherited_file_perms;
 
 # setup luks backed home directories in /run/cryptsetup
-lvm_manage_runtime_files(systemd_homework_t)
-lvm_manage_runtime_dirs(systemd_homework_t)
+optional_policy(`
+	lvm_manage_runtime_files(systemd_homework_t)
+	lvm_manage_runtime_dirs(systemd_homework_t)
+')
```
pebenito, line 757: This should also move to the bottom of the `systemd_homework_t` block.

pebenito (changes requested):

policy/modules/system/systemd.fc
pebenito: No, that's fine. I'm simply making sure I understand the intent.

pebenito (approved):

PR #842, opened 2024-12-12
systemd: various changes

WavyEbuilder:
systemd stores some early boot log files here, such as the userspace measurement event log for TPMs. The logfiles stored here aren't sensitive enough to warrant their own type, so let's just reuse systemd_log_t.


WavyEbuilder:
I've added a few more commits to this PR instead of opening loads of PRs for separate commits, so I've changed the title. I hope that's okay.

pebenito (approved):

PR #853, opened 2025-01-16
Add the necessary policies to run the tpm2 tools

uudiin:


policy/modules/services/tpm2.te
```diff
@@ -50,10 +52,18 @@ files_read_etc_files(tpm2_t)
...
+miscfiles_getattr_localization(tpm2_t)
 
 selinux_getattr_fs(tpm2_t)
 selinux_search_fs(tpm2_t)
 
+fs_getattr_efivarfs(tpm2_t)
+
+userdom_use_user_ptys(tpm2_t)
+
+ssh_use_sshd_pidfds(tpm2_t)
```
dsugar100, line 65: Use of ssh should probably be in an optional block.

policy/modules/system/userdomain.if
```diff
@@ -937,6 +937,10 @@ template(`userdom_common_user_template',`
 		virt_home_filetrans_svirt_home($1_t, dir, "qemu")
 		virt_home_filetrans_virt_home($1_t, dir, "VirtualMachines")
 	')
+
+	optional_policy(`
+		tpm2_run($1_t, $1_r)
```
pebenito, line 942: Why does this need to be run by unprivileged user domains?

uudiin: I think TPM provides general capabilities, not that tools only privileged users can use. Of course, this is not necessary. If it is not necessary, I will remove this line.

pebenito: I'd prefer to remove this since there is no specific need.

uudiin: Got it, updated already.

pebenito (approved):

PR #855, opened 2025-01-17
fixdep dbus

PPN-SD:
auth_use_pam_systemd requires dbus  :  

> /var/lib/selinux/targeted/tmp/modules/400/authlogin/cil:133 =
> (typeattributeset cil_gen_require dbusd_system_bus_client)

See : https://bugs.gentoo.org/941785

### DIFF
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -142,7 +142,6 @@ term_dontaudit_use_all_ptys(chkpwd_t)
 
 auth_read_shadow_history(chkpwd_t)
 auth_use_nsswitch(chkpwd_t)
-auth_use_pam_systemd(chkpwd_t)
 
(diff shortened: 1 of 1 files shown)


policy/modules/system/authlogin.te
```diff
@@ -142,7 +142,10 @@ term_dontaudit_use_all_ptys(chkpwd_t)
 
 auth_read_shadow_history(chkpwd_t)
 auth_use_nsswitch(chkpwd_t)
-auth_use_pam_systemd(chkpwd_t)
+
+ifdef(`init_systemd',` 
+	auth_use_pam_systemd(chkpwd_t)
+')
```
pebenito, line 148: This should move down after the `distro_ubuntu` block.

pebenito (changes requested):

pebenito (approved):

PR #859, opened 2025-02-06
Allow fapolicyd to watch /run/netns

dsugar100:
/run/netns is a tmpfs mountpoint

node=localhost type=AVC msg=audit(1738868630.348:1695): avc:  denied  { watch_sb watch_with_perm } for  pid=967 comm="fapolicyd" path="/run/netns" dev="tmpfs" ino=1954

Signed-of-by: Dave Sugar <dsugar100@gmail.com>

### DIFF
--- a/policy/modules/admin/fapolicyd.te
+++ b/policy/modules/admin/fapolicyd.te
@@ -84,6 +84,9 @@ fs_watch_all_fs(fapolicyd_t)
 logging_log_filetrans(fapolicyd_t, fapolicyd_log_t, file)
 logging_send_syslog_msg(fapolicyd_t)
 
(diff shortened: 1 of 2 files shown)


dsugar100:
I'm not sure the interface name is really the best, while it is a directory, maybe we should be specific that it is a mountpoint?  Thoughts?

pebenito:
> I'm not sure the interface name is really the best, while it is a directory, maybe we should be specific that it is a mountpoint? Thoughts?

The naming seems ok to me, since it expresses the intent; however, it should probably be 2 interfaces.

dsugar100:
Ok, please let me know if you want any changes.

pebenito (approved):

PR #865, opened 2025-02-18
Permit init_t (systemd) to start a detached screen/tmux session

jcpunk:
This policy tweak permits running a screen/tmux session via a system service/unit file.

Given the interactive nature of the process it might be worth putting behind a boolean...

### DIFF
--- a/policy/modules/apps/screen.if
+++ b/policy/modules/apps/screen.if
@@ -138,3 +138,20 @@ interface(`screen_dontaudit_getattr_sock_file',`
 
 	dontaudit $1 screen_runtime_t:sock_file getattr;
 ')
+
+########################################
+## <summary>
+##      Permit running screen
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+interface(`screen_exec',`
+	gen_require(`
+		type screen_exec_t;
+	')
+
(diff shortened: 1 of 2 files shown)


pebenito:
You're saying this is so in a unit you can do `ExecStart=/usr/bin/screen ....`?

jcpunk:
Exactly!  In particular: https://src.fedoraproject.org/rpms/tmux/blob/rawhide/f/tmux%40.service

pebenito:
You would need to create a `screen_exec()` interface in the `screen.if` and then call it from an optional in `init.te`.

jcpunk:
Does this look closer to what you'd expect?

policy/modules/apps/screen.te
```diff
@@ -22,6 +22,13 @@ type screen_runtime_t;
 files_runtime_file(screen_runtime_t)
 ubac_constrained(screen_runtime_t)
 
+## <desc>
+##      <p>
+##      permit init to start a detached screen session
+##      </p>
+## </desc>
+gen_tunable(init_launch_screen, false)
```
pebenito, line 30: I don't think adding a tunable is necessary.

pebenito (changes requested):

policy/modules/apps/screen.te
jcpunk: dropped

pebenito (approved):

PR #867, opened 2025-03-02
Add setcap to knotd / add knotc_initrc_domtrans

PPN-SD:
allow capabilities as it's the default behavior now

allow knotc cmd in init script: knotc conf-check (by default)

See https://github.com/CZ-NIC/knot/blob/master/distro/common/knot.service

### DIFF
--- a/policy/modules/services/knot.if
+++ b/policy/modules/services/knot.if
@@ -47,6 +47,25 @@ interface(`knot_run_client',`
 
 ########################################
 ## <summary>
+##      Execute knotc in knot init
+##      scripts in the initrc domain.
+## </summary>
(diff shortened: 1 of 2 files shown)


pebenito (approved):

PPN-SD:
@pebenito Sorry, it appears that I was misguided by a misinterpretation. This situation does not require domtrans, but rather this traditional solution: 

```
term_dontaudit_use_console(knotc_t)

init_use_fds(knotc_t)
init_use_script_ptys(knotc_t)
```

knotc is only used for a checkconfig here.

Do you agree with this rather than the interface created previously ?

PR #875, opened 2025-03-17
use init_use_script_ptys for knotc in initscript

PPN-SD:
See https://github.com/SELinuxProject/refpolicy/pull/867#issuecomment-2696704152

### DIFF
--- a/policy/modules/services/knot.te
+++ b/policy/modules/services/knot.te
@@ -110,6 +110,9 @@ manage_lnk_files_pattern(knotc_t, knot_var_lib_t, knot_var_lib_t)
 
 domain_use_interactive_fds(knotc_t)
 
+init_dontaudit_use_fds(knotc_t)
+init_use_script_ptys(knotc_t)
+
 files_read_etc_files(knotc_t)
 files_search_runtime(knotc_t)
 files_search_var_lib(knotc_t)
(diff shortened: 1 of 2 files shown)


pebenito:
I don't fully understand what you're saying. If there is no domtrans, then why would there be new permissions for knotc?

PPN-SD:
Sry for the confusion. There's already some domtrans for knotc.
From the beginning, I just need to deal with this warning (that I should have given from the start) :
`type=AVC msg=audit(1742241229.710:686): avc:  denied  { read write } for  pid=17740 comm="knotc" path="/dev/pts/5" dev="devpts" ino=8 scontext=system_u:system_r:knotc_t tcontext=system_u:object_r:initrc_devpts_t tclass=chr_file permissive=0`

And my previous commit didn't fix that at all ...

policy/modules/services/knot.te
```diff
@@ -119,3 +119,8 @@ fs_getattr_tmpfs(knotc_t)
 miscfiles_read_localization(knotc_t)
 
 userdom_use_user_ptys(knotc_t)
+
+init_dontaudit_use_fds(knotc_t)
+term_dontaudit_use_console(knotc_t)
+
+init_use_script_ptys(knotc_t)
```
pebenito, line 126: The init ones should be together at line 112 and term at 120.

pebenito (changes requested):
Rule-wise it looks good, but needs a style fix.

policy/modules/services/knot.te
PPN-SD: Thanks. Done.

pebenito (approved):

PR #880, opened 2025-03-19
Allow to specify module version

### LINKED ISSUE #877: Allow to specify module version in devel Makefile
Hi, would it be feasible to introduce a new option for the devel Makefile to specify module version?
The usecase is to build a selinux module on a newer platform that targets older versions with a lower module version.
Here is an example for the change:
https://github.com/SELinuxProject/refpolicy/compare/main...McCzarny:refpolicy:main

McCzarny:
resolves #877

This change introduces an equivalent of  `OUTPUT_POLICY` for policy modules. I was trying to stick to the existing parameter as the reference.
In `Rules.monolothic` there is a check with the kernel policy version using `policyvers`. As far as I can see there is no need for similar logic for the modules.

I've used the max module version displayed by `checkmodule -V` using the current version. Please, let me know if I should use versions defined in the header:
https://github.com/SELinuxProject/selinux/blob/main/libsepol/include/sepol/policydb/policydb.h

Example usage:
```
~ make -f ./Makefile OUTPUT_MODULE=18 mymodule.pp
[...]

~ sedismod module.pp 
Reading policy...
libsepol.policydb_index_others: security:  0 users, 1 roles, 2 types, 0 bools
libsepol.policydb_index_others: security:  1 classes, 0 rules, 0 cond rules
libsepol.policydb_index_others: security:  0 users, 1 roles, 2 types, 0 bools
libsepol.policydb_index_others: security:  1 classes, 0 rules, 0 cond rules
Binary policy module file loaded.
Module name: module
Module version: 1.0
Policy version: 18
```

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -4041,3 +4041,6 @@ Sven Vermeulen (27):
       Allow initrc_t to read stunnel configuration
       Introduce exec-check interfaces for passwd binaries and useradd binaries
       chfn_t reads in file context information and executes nscd
+
+Maciej Czarnecki (1):
+      Add support for output module policy version definition
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,10 @@ ifneq ($(OUTPUT_POLICY),)
 	override CHECKPOLICY += -c $(OUTPUT_POLICY)
 endif
 
+ifneq ($(OUTPUT_MODULE),)
+	override CHECKMODULE += -c $(OUTPUT_MODULE)
+endif
+
 ifneq "$(CUSTOM_BUILDOPT)" ""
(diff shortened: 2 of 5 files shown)


Changelog
```diff
@@ -4041,3 +4041,6 @@ Sven Vermeulen (27):
       Allow initrc_t to read stunnel configuration
       Introduce exec-check interfaces for passwd binaries and useradd binaries
       chfn_t reads in file context information and executes nscd
+
+Maciej Czarnecki (1):
```
21sw-clayton, line 4045: Each changelog release is in reverse chronological order.

It is likely automatically generated via Git

See the [last commit](https://github.com/SELinuxProject/refpolicy/commit/2d59cfe0120a4e21013baf8e99a3ac2f06ed193c)

PR #882, opened 2025-03-21
ssh: fix typo in port forwarding tunable

aerusso:
Extracted from the debian packaging, this was originally authored by Russell Coker.

All credit goes to the original author.

### DIFF
--- a/policy/modules/apps/xscreensaver.te
+++ b/policy/modules/apps/xscreensaver.te
@@ -10,7 +10,7 @@ policy_module(xscreensaver)
 ##	Grant the xscreensaver domains read access to generic user content
 ##	</p>
 ## </desc>
-gen_tunable(`xscreensaver_read_generic_user_content', true)
+gen_tunable(xscreensaver_read_generic_user_content, true)
 
 attribute_role xscreensaver_roles;
 attribute_role xscreensaver_helper_roles;
--- a/policy/modules/services/apache.if
+++ b/policy/modules/services/apache.if
@@ -32,7 +32,7 @@ template(`apache_content_template',`
 	##	be labeled public_content_rw_t.
 	##	</p>
 	## </desc>
-	gen_tunable(`allow_httpd_$1_script_anon_write', false)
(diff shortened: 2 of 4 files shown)


pebenito:
None of the other `gen_tunable()` uses have quoting like this. Additionally, the sediff done by github actions shows no change created by this patch. Is there an issue this resolves?

aerusso:
Sorry, this patch was a bit of "cargo cult" programming.  I ran this grep command `grep 'gen_tunable([^)]*\$' -r .`, and saw this output:

```
./policy/modules/system/userdomain.if:	gen_tunable(`$1_read_generic_user_content', true)
./policy/modules/system/userdomain.if:	gen_tunable(`$1_read_all_user_content', false)
./policy/modules/system/userdomain.if:	gen_tunable(`$1_manage_generic_user_content', false)
./policy/modules/system/userdomain.if:	gen_tunable(`$1_manage_all_user_content', false)
./policy/modules/services/ssh.if:	gen_tunable($1_port_forwarding, false)
./policy/modules/services/apache.if:	gen_tunable(`allow_httpd_$1_script_anon_write', false)
```

I assumed this meant that m4 needed backticks around the argument if there was a variable substitution.  (I'm learning m4 by reading refpolicy.)  I just now went ahead and recompiled with and without this, and I see the variable expansion occurring correctly in both cases (so I'm seeing the same sediff result as you, I believe)

Anyway, there's one other use of backticks in a `gen_tunable` besides the above:

```
./policy/modules/services/i18n_input.te:gen_tunable(`i18n_input_read_generic_user_content', true)
```

I'm guessing these backticks have been placed in error?  Or is it perhaps a GNU extension to m4 that the backticks are not required when doing variable substitution?  In any case, these are inconsistent, and should probably be cleaned up.

I can write the patch making these uniform, but I would like to make sure this isn't intentional.

pebenito:
Dropping the extra quotes should be fine.

pebenito (approved):

PR #890, opened 2025-04-09
systemd: allow reading /dev/cpu/0/msr

dgdegraaf:
To detect confidential virtualization on AMD processors, systemd as of v254 needs to perform an MSR read.


pebenito (approved):

pebenito:
Thanks!

PR #901, opened 2025-05-06
files: add files_delete_var_chr_files interface

21sw-clayton:


policy/modules/kernel/files.if
```diff
@@ -5921,6 +5921,25 @@ interface(`files_search_var',`
...
+##	</summary>
+## </param>
+#
+interface(`files_delete_var_chr_files',`
+	gen_require(`
+		type var_t;
+	')
+
+	allow $1 var_t:chr_file delete_chr_file_perms;
+')
```
pebenito, line 5941: This should move to line 6180. Additionally, you should use `delete_chr_files_pattern()` so appropriate directory access is allowed too.

pebenito (changes requested):

policy/modules/kernel/files.if
21sw-clayton: I wasn't sure where to put it, but I've now located the [Style Guide](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide).

policy/modules/kernel/files.if
```diff
@@ -6159,6 +6159,25 @@ interface(`files_read_var_symlinks',`
...
+##	</summary>
+## </param>
+#
+interface(`files_delete_var_chr_files',`
+	gen_require(`
+		type var_t;
+	')
+
+	delete_chr_files_pattern($1, var_t, var_t)
+')
```
pebenito, line 6179: I think I made some confusion with the line numbers (orig. file vs modified file). This should move down one more interface, just above the filetrans one, on line 6180(orig)/6199(modified), so it's not in the middle of the symlink interfaces.

Sorry for the confusion.

21sw-clayton: Thanks for the direction!

I'm trying to understand this using the [style guide (if)](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide#modulenameif-files):

Is `filetrans` considered a _transform_ interface?

pebenito: No, transform interfaces are ones like `files_type()`, which could fit the description "[type] is a ...".

pebenito (approved):

PR #902, opened 2025-05-10
Rework shadow transitions and access

aerusso:
shadow access is tightly controlled, with separate types for the shadow files and the locks.  This patch distinguishes the two by enumerating the backup filenames and lock file names in their associated file transition rules.

Prior to this, the overbroad file transition rules would cause various shadow-manipulating tools to create lock files with the incorrect shadow_t label.

### DIFF
--- a/policy/modules/system/authlogin.if
+++ b/policy/modules/system/authlogin.if
@@ -737,7 +737,7 @@ interface(`auth_manage_shadow',`
 ## </param>
 ## <param name="name" optional="true">
 ##	<summary>
-##	The name of the object being created.
+##	The name of the object being created. (Deprecated)
 ##	</summary>
 ## </param>
 #
@@ -746,7 +746,16 @@ interface(`auth_etc_filetrans_shadow',`
 		type shadow_t;
 	')
 
-	files_etc_filetrans($1, shadow_t, file, $2)
+	ifelse(`$2',`',`
+		files_etc_filetrans($1, shadow_t, file, "shadow")
+		files_etc_filetrans($1, shadow_t, file, "shadow-")
+		files_etc_filetrans($1, shadow_t, file, "shadow.upwd-write")
+		files_etc_filetrans($1, shadow_t, file, "gshadow")
+		files_etc_filetrans($1, shadow_t, file, "gshadow-")
+	',`
+		refpolicywarn(`$0($*) second parameter is deprecated.')
+		files_etc_filetrans($1, shadow_t, file, $2)
+	')
 ')
 
 ########################################
@@ -865,7 +874,10 @@ interface(`auth_rw_shadow_lock',`
 		type shadow_lock_t;
 	')
 
-	rw_files_pattern($1, shadow_lock_t, shadow_lock_t)
+	allow $1 shadow_lock_t:file rw_file_perms;
+	files_etc_filetrans($1, shadow_lock_t, file, ".pwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "passwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "group.lock")
 ')
 
 ########################################
--- a/policy/modules/admin/dpkg.te
+++ b/policy/modules/admin/dpkg.te
@@ -295,7 +295,7 @@ term_use_all_terms(dpkg_script_t)
 
 files_manage_non_auth_files(dpkg_script_t)
 
-auth_etc_filetrans_shadow(dpkg_script_t, "shadow.upwd-write")
+auth_etc_filetrans_shadow(dpkg_script_t)
 auth_manage_shadow(dpkg_script_t)
 
 init_all_labeled_script_domtrans(dpkg_script_t)
--- a/policy/modules/system/authlogin.te
+++ b/policy/modules/system/authlogin.te
@@ -415,6 +415,7 @@ term_dontaudit_use_console(updpwd_t)
 term_dontaudit_use_unallocated_ttys(updpwd_t)
 
 auth_manage_shadow(updpwd_t)
+auth_etc_filetrans_shadow(updpwd_t)
 auth_use_nsswitch(updpwd_t)
 
 logging_send_syslog_msg(updpwd_t)


aerusso:
I adjusted the rw_files_pattern macro to allow searching /etc (etc_t), and I think that's causing the lint failure.  Should I just have it grant the useless dir search permission for shadow_lock_t, and expect that the etc_t search permission is granted by some other rule?

policy/modules/system/authlogin.if
```diff
@@ -859,9 +861,13 @@ interface(`auth_relabel_shadow',`
 interface(`auth_rw_shadow_lock',`
 	gen_require(`
 		type shadow_lock_t;
+		type etc_t;
 	')
 
-	rw_files_pattern($1, shadow_lock_t, shadow_lock_t)
+	files_etc_filetrans($1, shadow_lock_t, file, ".pwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "passwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "group.lock")
```
dsugar100, line 869: Maybe these uses of 'files_etc_filetrans' belong in a new interface instead (something line auth_filetrans_shadow_lock).  These give permission to transition on file creation, but not to actually create the file.

dsugar100:
With these changes have to tested changing a users' password (twice - due to files created after the first change)?  And keep in mind that things behave differently in enforcing vs. permissive.

policy/modules/system/authlogin.if
pebenito: I agree.

policy/modules/system/authlogin.if
```diff
@@ -719,6 +719,8 @@ interface(`auth_manage_shadow',`
 	auth_manage_shadow_history($1)
 	auth_rw_shadow_lock($1)
 	allow $1 shadow_t:file manage_file_perms;
+	auth_etc_filetrans_shadow($1, "shadow-")
+	auth_etc_filetrans_shadow($1, "gshadow-")
```
pebenito, line 723: Please update the existing `auth_etc_filetrans_shadow` instead.

policy/modules/system/authlogin.if
```diff
@@ -859,9 +861,13 @@ interface(`auth_relabel_shadow',`
...
 	gen_require(`
 		type shadow_lock_t;
+		type etc_t;
 	')
 
-	rw_files_pattern($1, shadow_lock_t, shadow_lock_t)
+	files_etc_filetrans($1, shadow_lock_t, file, ".pwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "passwd.lock")
+	files_etc_filetrans($1, shadow_lock_t, file, "group.lock")
+	rw_files_pattern($1, etc_t, shadow_lock_t)
```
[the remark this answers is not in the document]
aerusso: I separated this change out into another patch and moved the filetrans logic to another macro, so this issue can be handled separately.

policy/modules/system/authlogin.if
aerusso: Done, and I did something similar for the other file transitions.  Let me know if this is what you were thinking of/if it's what you want.

policy/modules/system/authlogin.if
aerusso: I handled this slightly differently, but hopefully it addresses your concern.

aerusso:
I've also included some dpkg-specific changes, but (despite running Debian) have not tested them.  This is in the final patch, and is motivated by a a read of the update-passwd.c source file.

policy/modules/system/authlogin.if
pebenito: After further consideration, I think we should keep the filetranses in the original interfaces and use `ifelse`, like `systemd_tmpfilesd_managed`:

```
ifelse(`$2',`',`
	files_etc_filetrans($1, shadow_lock_t, file, ".pwd.lock")
	[...]
',`
	refpolicywarn(`$0($*) second parameter is deprecated.')
	files_etc_filetrans($1, shadow_t, file, $2)
')
```

pebenito (changes requested):
Please see existing open comment.

policy/modules/system/authlogin.if
```diff
@@ -860,9 +877,18 @@ interface(`auth_relabel_shadow',`
...
-	rw_files_pattern($1, shadow_lock_t, shadow_lock_t)
+	allow $1 shadow_lock_t:file rw_file_perms;
+	ifelse(`$2',`',`
+		files_etc_filetrans($1, shadow_lock_t, file, ".pwd.lock")
+		files_etc_filetrans($1, shadow_lock_t, file, "passwd.lock")
+		files_etc_filetrans($1, shadow_lock_t, file, "group.lock")
+	',`
+		refpolicywarn(`$0($*) second parameter is deprecated.')
+		files_etc_filetrans($1, shadow_t, file, $2)
+	')
```
pebenito: This is the right idea, but I meant to put it in `auth_etc_filetrans_shadow()`, making the changes in the other files unnecessary.

aerusso: Sorry, I'm trying to understand this request.  `auth_etc_filetrans_shadow` seems to me to be a macro designed to cause files created in `/etc` to be in the type `shadow_t`, optionally restricting that behavior to files of a certain name.

Are you suggesting re-purposing that macro to only act on, e.g., {`shadow`, `gshadow`, `shadow-`, `gshadow-`}, so that the existing `auth_etc_filetrans_shadow(foo_t)` lines in other modules do the right thing without being modified?

The lines you are commenting on here are in `auth_relabel_shadow`, so I'm worried I'm missing something.

pebenito: > Are you suggesting re-purposing that macro to only act on, e.g., {`shadow`, `gshadow`, `shadow-`, `gshadow-`}, so that the existing `auth_etc_filetrans_shadow(foo_t)` lines in other modules do the right thing without being modified?

Yes, this is what I'm requesting.

policy/modules/system/authlogin.if
```diff
@@ -732,18 +732,22 @@ interface(`auth_manage_shadow',`
 ##	Domain allowed access.
 ##	</summary>
 ## </param>
-## <param name="name" optional="true">
-##	<summary>
-##	The name of the object being created.
-##	</summary>
-## </param>
```
pebenito, line 742: Please keep this block, but add `(Deprecated)` to the summary, since the interface still responds to the parameter.

Otherwise looks good to me.

pebenito (changes requested):

aerusso:
@pebenito I apparently forgot to push these changes after making them.  I've rebased, and applied your last comment.

pebenito (approved):

pebenito (changes requested):
Please squash to one commit, otherwise this is good to go.

pebenito (approved):

PR #903, opened 2025-05-10
WIP: apache: tag nginx httpd_runtime files

aerusso:
nginx has runtime files that should be appropriately labeled

### DIFF
--- a/policy/modules/admin/logrotate.te
+++ b/policy/modules/admin/logrotate.te
@@ -150,7 +150,8 @@ optional_policy(`
 
 optional_policy(`
 	apache_read_config(logrotate_t)
-	apache_domtrans(logrotate_t)
+	apache_ctl_domtrans(logrotate_t)
+	apache_ctl_domtrans_from_httpd(logrotate_t)
 	apache_signull(logrotate_t)
 ')
 
--- a/policy/modules/services/apache.te
+++ b/policy/modules/services/apache.te
@@ -269,9 +269,18 @@ attribute httpd_script_domains;
 attribute_role httpd_helper_roles;
 roleattribute system_r httpd_helper_roles;
 
+attribute_role httpd_ctl_roles;
+roleattribute system_r httpd_ctl_roles;
+
 type httpd_t;
 type httpd_exec_t;
 init_daemon_domain(httpd_t, httpd_exec_t)
+
+type httpd_ctl_t;
+type httpd_ctl_exec_t;
+application_domain(httpd_ctl_t, httpd_ctl_exec_t)
+allow httpd_ctl_t httpd_exec_t:file entrypoint;
+role httpd_ctl_roles types httpd_ctl_t;
 
 type httpd_cache_t;
 files_type(httpd_cache_t)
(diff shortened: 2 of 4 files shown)


aerusso:
I'm not sure of the best way to handle the nginx transition to allow `-t` to work correctly.  This approach overloads http_helper_roles to allow directly starting nginx in the httpd_t domain.  If that is too much authority, maybe we can create a weaker role (perhaps reuse httpd_helper_t?) that http_exec_t can domtrans to, and is powerful enough to read the configuration files, restart the server, etc.

aerusso:
apachectl is already marked at httpd_exec_t, so this problem is not specific to nginx.  I can make a new domain, like httpd_ctl_t, with associated httpd_ctl_roles.  Would that be the preferred way to do this?

pebenito:
> apachectl is already marked at httpd_exec_t, so this problem is not specific to nginx. I can make a new domain, like httpd_ctl_t, with associated httpd_ctl_roles. Would that be the preferred way to do this?

This sounds like the best way forward.

aerusso:
I went forward and created a `httpd_ctl_t` domain.  It's a little frustrating, because `nginx -t` really wants to check a lot of things out.  It actually needs to be able to bind to various ports.

Indeed, it even wants to open various files (for example, the error log that is, at least on my machine, owned by www-data, and root is not explicitly granted access to it).  Therefore, it winds up using dac_override.  I suppose we could try to interact with various distributions and upstream to make sure that root is given read access to that file, and then we could remove dac_override from both httpd_t and httpd_ctl_t, but for right now this means that httpd_ctl_t needs quite a bit of permission.

policy/modules/admin/logrotate.te
```diff
@@ -150,7 +150,8 @@ optional_policy(`
 
 optional_policy(`
 	apache_read_config(logrotate_t)
-	apache_domtrans(logrotate_t)
+	apache_ctl_domtrans(logrotate_t)
+	apache_ctl_domtrans_from_httpd(logrotate_t)
```
pebenito, line 154: This should be a non-`from` interface call from the apache module. The `from` interfaces are intended for users to use without having to modify the policy sources.

policy/modules/services/apache.te
```diff
@@ -903,6 +912,157 @@ optional_policy(`
...
+
+setattr_dirs_pattern(httpd_ctl_t, httpd_runtime_t, httpd_runtime_t)
+manage_dirs_pattern(httpd_ctl_t, httpd_runtime_t, httpd_runtime_t)
+manage_files_pattern(httpd_ctl_t, httpd_runtime_t, httpd_runtime_t)
+manage_sock_files_pattern(httpd_ctl_t, httpd_runtime_t, httpd_runtime_t)
+files_runtime_filetrans(httpd_ctl_t, httpd_runtime_t, { file sock_file dir })
+
+manage_dirs_pattern(httpd_ctl_t, squirrelmail_spool_t, squirrelmail_spool_t)
+manage_files_pattern(httpd_ctl_t, squirrelmail_spool_t, squirrelmail_spool_t)
+manage_lnk_files_pattern(httpd_ctl_t, squirrelmail_spool_t, squirrelmail_spool_t)
```
pebenito, line 1001: I think this plus especially the port binding makes me reconsider my position...

Overall, I'm beginning to feel this seems too much like `httpd_t` to make it worthwhile. I'm not definitely saying no, but I'm leaning that direction.

Another option would be to move nginx into a new domain.

aerusso: So I've been thinking about how to handle this change.  Would it make sense to create an attribute `httpd_domains` that would be common to both nginx and apache?  If we opted to do that, should I create a new httpd module for the common parts, and an nginx module for the new domain?

pebenito: That is a possibility. I'd have to see the full extent of rules that use it.

pebenito:
@aerusso is this another missed push or are you still working on this?

PR #907, opened 2025-05-15
xserver: usability fixes for the Qt library

gtrentalancia:
Usability fixes for the Qt library used by some graphical applications.

### DIFF
--- a/policy/modules/services/xserver.if
+++ b/policy/modules/services/xserver.if
@@ -170,7 +170,7 @@ template(`xserver_restricted_role',`
 #
 template(`xserver_role',`
 	gen_require(`
-		type iceauth_home_t, xserver_t, xserver_tmp_t, xserver_tmpfs_t, xauth_home_t;
+		type iceauth_home_t, xserver_t, xserver_tmpfs_t, xauth_home_t;
 		type user_fonts_t, user_fonts_cache_t, user_fonts_config_t;
 		type mesa_shader_cache_t, xdm_t;
 	')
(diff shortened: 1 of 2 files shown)


policy/modules/services/xserver.if
```diff
@@ -187,9 +187,6 @@ template(`xserver_role',`
 		allow $2 xserver_tmpfs_t:file rw_file_perms;
 	')
 
-	# XCB Event Queue: used by the Qt library for example
-	allow $2 xserver_tmp_t:file rw_file_perms;
```
pebenito, line 191: You've confirmed this is no longer required for Qt 6?

policy/modules/services/xserver.te
```diff
@@ -59,10 +59,10 @@ gen_tunable(allow_write_xshm, false)
 ## <desc>
 ## <p>
 ## Allows clients to write to the X server tmpfs
-## files.
+## files (needed by the Qt library).
 ## </p>
 ## </desc>
-gen_tunable(xserver_client_writes_xserver_tmpfs, false)
+gen_tunable(xserver_client_writes_xserver_tmpfs, true)
```
pebenito, line 65: Please do not change this default.

pebenito (changes requested):

policy/modules/services/xserver.if
gtrentalancia: Neither of xserver_tmp_t and xserver_tmpfs_t are needed in the latest Qt version 6.

But I have only carried out testing with the qtdoc examples git://code.qt.io/qt/qtdoc.git because all real world applications that I use are still linked to Qt5.

policy/modules/services/xserver.te
gtrentalancia: Yes, considering https://github.com/SELinuxProject/refpolicy/pull/907#discussion_r2093692605, I agree and will amend both the default value and comment.

Issue #797, opened 2024-07-15
Possible revisions to sshd policy due to component separation


0xC0ncord:
As mentioned in https://github.com/SELinuxProject/refpolicy/pull/793, OpenSSH is reworking the `sshd` binary by splitting some of its functionality into separate components. This started with OpenSSH 9.8 by splitting the SSH protocol and listening functionality into `ssh-session` and `sshd` respectively. According to the changelog, there are plans to further separate `sshd` into separate components.

We should watch these coming changes and consider reworking the policy for `sshd` to cover these components individually.

0xC0ncord:
Something additional to consider: how should the policy continue to cover `ssh` daemons which continue to have monolithic behavior (e.g. `dropbear` or older OpenSSH versions)?

pebenito:
> Something additional to consider: how should the policy continue to cover `ssh` daemons which continue to have monolithic behavior (e.g. `dropbear` or older OpenSSH versions)?

Put the permissions that are removed from sshd_t into a conditional `sshd_unified`.

```
optional_policy(`sshd_unified',`
  <perms moved to session domain>
',`
  <domtrans to session domain>
```

xCEVre:
new sshd-auth file https://www.openssh.com/txt/release-10.0

* sshd(8): this release removes the code responsible for the
   user authentication phase of the protocol from the per-
   connection sshd-session binary to a new sshd-auth binary.
   Splitting this code into a separate binary ensures that the
   crucial pre-authentication attack surface has an entirely
   disjoint address space from the code used for the rest of the
   connection. It also yields a small runtime memory saving as the
   authentication code will be unloaded after the authentication
   phase completes. This change should be largely invisible to
   users, though some log messages may now come from "sshd-auth"
   instead of "sshd-session". Downstream distributors of OpenSSH
   will need to package the sshd-auth binary.

xCEVre:
type transitions:
checkpolicy -MbC /sys/fs/selinux/policy -o - |grep typetransition|grep -E "(ssh.*ssh|systemd_t.*systemd_executor_t|kernel.*systemd)"
```
(typetransition kernel_t systemd_exec_t process systemd_t)
(typetransition systemd_t systemd_executor_exec_t process systemd_executor_t)
(typetransition systemd_executor_t usr_bin_sshd_exec_t process openssh_sshd_t)
(typetransition openssh_sshd_t usr_lib_ssh_sshd_session_exec_t process openssh_sshd_session_t)
[...]

PR #910, opened 2025-05-21
sshd: label sshd-auth as sshd_exec_t #797

PPN-SD:
openssh-10.0 has split off the authentication logic into a new binary called sshd-auth. As sshd-session, relabel with sshd_exec_t now before a reworking policy.

See #797


PPN-SD:
Add path for debian too : 
https://packages.debian.org/trixie/amd64/openssh-server/filelist

0xC0ncord:
LGTM!

pebenito (approved):

PR #911, opened 2025-05-24
Fix boot issues with dracut-generated initramfs

gtrentalancia:
Update the kernel module in order to keep the initramfs generated by dracut bootable with SELinux in enforcing mode.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/377

 policy/modules/kernel/files.if  |   19 +++++++++++++++++++
 policy/modules/kernel/kernel.te |    1 +
 2 files changed, 20 insertions(+)


gtrentalancia:
The problem is most probably due to a SELinux kernel change which was introduced in an undetermined kernel version between 6.0.9 and 6.9.1 and it can be fixed in three ways (with different security impact):

1. by reverting (or ?fixing?) the kernel change which led to the need for the additional kernel domain "root_t:file entrypoint" SELinux permission upon booting (mounting) the system - this has probably the minimum security impact (or no security impact at all), however the kernel commit causing the issue has not been determined yet;
2. fixing dracut by loading the SELinux policy after switch_root and removing the selinux module (this has a small security impact in the first part of the boot process until switch_root) - see https://github.com/dracutdevs/dracut/pull/2652 and https://github.com/dracut-ng/dracut-ng/pull/376;
3. changing the SELinux policy in order to grant the kernel domain additional permission root_t:file entrypoint (this has the highest security impact in that such permission is always granted and not only during the initial boot phase) - see https://github.com/SELinuxProject/refpolicy/pull/911.

gtrentalancia:
As explained in https://github.com/SELinuxProject/refpolicy/pull/911#issuecomment-2906767123 a kernel patch would be the preferred way of dealing with the issue in order to avoid a security impact.

A kernel patch is proposed for testing: 
[linux-6.14.2-check-policy-capabilities-before-setting-the-initial-context-before-policy-load.txt](https://github.com/user-attachments/files/20424143/linux-6.14.2-check-policy-capabilities-before-setting-the-initial-context-before-policy-load.txt)

PR #923, opened 2025-06-14
fwupd-fixed

etbe:


pebenito:
The sechecker output is in the validation-* artifact of the checks.

```
Check name: PRIVILEGE-CAP_SYS_RAWIO

Description: Verify only expected domains can use CAP_SYS_RAWIO

    * allow fwupd_t fwupd_t:capability { dac_override dac_read_search linux_immutable sys_admin sys_rawio };

Check FAILED
```

etbe:
OK, how do we tell it that fwupd_t actually needs this?

PR #929, opened 2025-06-16
opensnitch daemon

etbe:


policy/modules/system/opensnitch.te
```diff
@@ -0,0 +1,98 @@
...
+## </p>
+## </desc>
+gen_tunable(allow_opensnitch_tmp, true)
+
+## <desc>
+## <p>
+## allow writing config files
+## </p>
+## </desc>
+gen_tunable(allow_opensnitch_write_config, true)
```
pebenito: Please remove `allow` from the names. I'd prefer to also set them as false by default unless it would lead to a very surprising result for users.

policy/modules/system/opensnitch.te
```diff
@@ -0,0 +1,98 @@
...
+allow opensnitchd_t self:perf_event { open kernel tracepoint cpu read write };
+allow opensnitchd_t self:netlink_tcpdiag_socket { create getattr bind setopt write read nlmsg_read };
+
+tunable_policy(`allow_opensnitch_write_config',`
+	allow opensnitchd_t opensnitchd_conf_t:dir { watch rw_dir_perms  };
+	allow opensnitchd_t opensnitchd_conf_t:file { watch manage_file_perms };
+', `
+	allow opensnitchd_t opensnitchd_conf_t:dir { watch list_dir_perms  };
+	allow opensnitchd_t opensnitchd_conf_t:file { watch read_file_perms };
+')
```
pebenito: Please move down with the other tunable block.

policy/modules/system/systemd.if
```diff
@@ -2699,6 +2699,25 @@ interface(`systemd_read_resolved_runtime',`
...
+## <summary>
+##  Allow domain to read directory containing resolv.conf
+## </summary>
+## <param name="domain">
+##     <summary>
+##     domain allowed access
+##     </summary>
+## </param>
+#
+interface(`systemd_list_resolved_runtime_dir',`
```
pebenito: `systemd_list_resolved_runtime`

policy/modules/system/opensnitch.te
etbe: Setting as false by default isn't a good option as we want things to generally just work for users. A system with SE Linux and an unfortunate location for the socket is more secure than running it in the same way without SE Linux.

policy/modules/system/opensnitch.te
```diff
@@ -0,0 +1,98 @@
...
+#
+# Declarations
+#
+
+## <desc>
+## <p>
+## allow user sockets in /tmp
+## </p>
+## </desc>
+gen_tunable(opensnitch_tmp, true)
```
pebenito: Please make the tunable name a bit more descriptive, e.g. something like `opensnitch_use_tmp_sockets` or `opensnitch_use_user_sockets`

pebenito (approved):

PR #930, opened 2025-06-17
newsystemd2

etbe:


policy/modules/kernel/filesystem.if
```diff
@@ -1362,6 +1362,24 @@ interface(`fs_watch_memory_pressure',`
...
+## <summary>
+##      Allow mounting a file ovr a memory.pressure file
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Source domain
+##      </summary>
+## </param>
+#
+interface(`fs_mounton_memory_pressure',`
```
pebenito: This isn't being used. Is this for `init_t` to call? If so, `init_mountpoint()` should be used with `memory_pressure_t` as the parameter.

policy/modules/system/systemd.if
```diff
@@ -1409,6 +1409,24 @@ interface(`systemd_use_logind_fds',`
...
+## <summary>
+##   Allow systemd-logind to use fds received from dbus
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`systemd_logind_use_fds',`
```
pebenito: `systemd_use_logind_fds`

policy/modules/system/authlogin.if
```diff
@@ -91,6 +91,7 @@ interface(`auth_use_pam_systemd',`
 	systemd_connect_machined($1)
 	systemd_dbus_chat_logind($1)
 	systemd_read_logind_state($1)
+	systemd_logind_use_fds($1)
```
pebenito: Are these pidfds?

policy/modules/system/userdomain.if
```diff
@@ -4274,6 +4274,24 @@ interface(`userdom_dontaudit_setattr_user_ttys',`
...
+## <summary>
+##	watch a user domain tty.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
+##	</summary>
+## </param>
+#
+interface(`userdom_watch_user_ttys',`
```
pebenito: Doesn't seem to be used.

policy/modules/system/systemd.te
```diff
@@ -1253,6 +1254,8 @@ kernel_dontaudit_getattr_proc(systemd_machine_id_setup_t)
 kernel_read_kernel_sysctls(systemd_machine_id_setup_t)
 kernel_read_system_state(systemd_machine_id_setup_t)
 
+dev_write_kmsg(systemd_machine_id_setup_t)
```
pebenito: Shouldn't be needed due to the systemd_log_parse_env_type() use.

etbe: ok I'll remove that

policy/modules/system/userdomain.if
etbe: I'll remove that.

policy/modules/system/systemd.if
etbe: ok

policy/modules/system/authlogin.if
etbe: no that's for logind to open devices for things like wayland before starting the session.

policy/modules/kernel/filesystem.if
etbe: OK I'll remove that.

pebenito (approved):

PR #933, opened 2025-06-17
bootloader

etbe:


policy/modules/admin/bootloader.fc
```diff
@@ -23,6 +23,7 @@
 /usr/sbin/grub2?-mkconfig	--	gen_context(system_u:object_r:bootloader_exec_t,s0)
 /usr/sbin/grub2?-probe	--	gen_context(system_u:object_r:bootloader_exec_t,s0)
 /usr/sbin/lilo.*	--	gen_context(system_u:object_r:bootloader_exec_t,s0)
+/usr/sbin/mkinitramfs		--	gen_context(system_u:object_r:bootloader_exec_t,s0)
```
pebenito: I don't think this fits. Much more than configuring the bootloader is done in mkinitramfs.

etbe: What are you suggesting as an alternative?

etbe: OK I'll remove that so you can merge the rest.

PR #936, opened 2025-06-17
mail

etbe:


policy/modules/services/dovecot.te
```diff
@@ -272,6 +272,8 @@ kernel_dontaudit_getattr_proc(dovecot_auth_t)
 
 kernel_getattr_proc(dovecot_auth_t)
 
+kernel_getattr_proc(dovecot_auth_t)
```
pebenito, line 271: Dupe line.

policy/modules/services/postfix.te
```diff
@@ -770,6 +770,10 @@ optional_policy(`
 	systemd_use_nss(postfix_showq_t)
 ')
 
+optional_policy(`
+	unconfined_run_to(postfix_showq_t, postfix_showq_exec_t)
```
pebenito: A proper domtrans interface for `postfix_showq_t` should be added instead. The "run_to" interfaces are only for downstream users.

etbe: I removed that, it seems that the issue was addressed in another way since I wrote that patch, it appears to work fine without it now.

pebenito (approved):

PR #939, opened 2025-06-17
device

etbe:


policy/modules/services/devicekit.te
```diff
@@ -143,6 +147,7 @@ auth_use_nsswitch(devicekit_disk_t)
 
 logging_send_syslog_msg(devicekit_disk_t)
 
+mount_watch_runtime_dirs(devicekit_disk_t)
```
pebenito: Please move up with other mount calls.

policy/modules/services/devicekit.te
```diff
@@ -116,6 +117,7 @@ files_getattr_all_files(devicekit_disk_t)
 files_getattr_all_pipes(devicekit_disk_t)
 files_manage_boot_dirs(devicekit_disk_t)
 files_manage_mnt_dirs(devicekit_disk_t)
+files_mounton_mnt(devicekit_disk_t)
```
pebenito, line 120: How is this mountpoint used?

etbe: Just for mounting stuff wherever udisksd etc want to do it with mnt_t as the label for the mountpoint.

pebenito (approved):

PR #941, opened 2025-06-18
udev

etbe:


policy/modules/system/udev.te
```diff
@@ -169,6 +177,10 @@ auth_read_pam_console_data(udev_t)
 auth_domtrans_pam_console(udev_t)
 auth_use_nsswitch(udev_t)
 
+# for /run/console-setup
+fs_manage_tmpfs_dirs(udev_t)
+fs_manage_tmpfs_files(udev_t)
```
pebenito: Can you elaborate? This seems like a labeling issue.

etbe: OK I'll remove that and do more tests

etbe: [    8.063825] audit: type=1400 audit(1750847044.008:6): avc:  denied  { read } for  pid=349 comm="udevadm" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:udev_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=dir permissive=0

How do I find out which of the tmpfs filesystems udev was trying to list before it was labelled?

etbe: I've put these lines back because I wasn't adding them, I was making rules that were conditional on not-systemd be unconditional. So apparently whatever requirements there are for having those rules have already been met in the non-systemd case

pebenito: Sorry, I'm a bit confused. Is udevadm emitting any of its own errors when the denial is hit? Is there a syscall audit message?

etbe: OK I've made another change to just allow listing tmpfs.

[    8.478782] audit: type=1400 audit(1750921750.320:6): avc:  denied  { read } for  pid=346 comm="udevadm" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:udev_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=dir permissive=0

pebenito (approved):

PR #943, opened 2025-06-18
login

etbe:


policy/modules/apps/xscreensaver.fc
```diff
@@ -6,3 +6,4 @@ HOME_DIR/XScreenSaver		--	gen_context(system_u:object_r:xscreensaver_config_t,s0
 /usr/bin/xscreensaver-gl-helper	--	gen_context(system_u:object_r:xscreensaver_helper_exec_t,s0)
 
 /usr/libexec/xscreensaver(/.*)?	--	gen_context(system_u:object_r:xscreensaver_helper_exec_t,s0)
+/usr/lib/aarch64-linux-gnu/libexec/kscreenlocker_greet	--	gen_context(system_u:object_r:xscreensaver_exec_t,s0)
```
pebenito: kscreenlocker_greet isn't specific t aarch64, so a more general fc would be better.

etbe:
@pebenito Does this need any changes?

policy/modules/apps/xscreensaver.fc
etbe: Turned out that on systems other plan plasma-mobile that caused some problems so I deleted it from this PR. Will add it in another one.

pebenito (approved):

PR #948, opened 2025-06-24
systemd-machined-modules-passwd

etbe:


policy/modules/system/systemd.te
```diff
@@ -1175,14 +1175,24 @@ allow systemd_machined_t self:cap_userns sys_chroot;
...
 
+ps_process_pattern(systemd_machined_t, systemd_nspawn_t)
+
 term_create_pty(systemd_machined_t, systemd_machined_devpts_t)
 allow systemd_machined_t systemd_machined_devpts_t:chr_file manage_chr_file_perms;
 
 manage_files_pattern(systemd_machined_t, systemd_machined_runtime_t, systemd_machined_runtime_t)
 allow systemd_machined_t systemd_machined_runtime_t:lnk_file manage_lnk_file_perms;
+allow systemd_machined_t systemd_machined_runtime_t:dir manage_dir_perms;
+allow systemd_machined_t systemd_machined_runtime_t:sock_file { create unlink };
```
pebenito: Please use `manage_sock_file_perms` for this and the below addition for systemd_userdb_runtime_t.

policy/modules/system/systemd.te
```diff
@@ -1259,7 +1273,12 @@ init_read_state(systemd_machine_id_setup_t)
 
 systemd_log_parse_environment(systemd_machine_id_setup_t)
 
-########################################
```
pebenito, line 1262: Please restore the comment header

policy/modules/system/systemd.te
```diff
@@ -1259,7 +1273,12 @@ init_read_state(systemd_machine_id_setup_t)
...
+logging_send_syslog_msg(systemd_machine_id_setup_t)
```
pebenito: This shouldn't be necessary due to the `systemd_log_parse_environment(systemd_machine_id_setup_t)`

policy/modules/system/systemd.te
```diff
@@ -1627,6 +1662,9 @@ miscfiles_read_localization(systemd_passwd_agent_t)
 seutil_search_default_contexts(systemd_passwd_agent_t)
 
 userdom_use_user_terminals(systemd_passwd_agent_t)
+userdom_search_user_runtime_root(systemd_passwd_agent_t)
+userdom_search_user_runtime(systemd_passwd_agent_t)
+systemd_search_user_runtime(systemd_passwd_agent_t)
```
pebenito: `systemd_search_user_runtime()` should already call the other two lines.

policy/modules/system/systemd.te
```diff
@@ -1611,13 +1644,15 @@ selinux_get_enforce_mode(systemd_passwd_agent_t)
 selinux_getattr_fs(systemd_passwd_agent_t)
 
 term_read_console(systemd_passwd_agent_t)
+term_use_unallocated_ttys(systemd_passwd_agent_t)
```
pebenito, line 1646: In which situation is this used?

etbe: The passwd agent can run early in the boot process and when the system is in single user mode so the normal login process hasn't happened.

lvm_signull(systemd_passwd_agent_t)

The above policy line which is already in the refpolicy is related to cryptsetup password on boot which ties in to this.

pebenito (approved):

PR #952, opened 2025-06-25
user-bubblewrap

etbe:


policy/modules/apps/bubblewrap.if
```diff
@@ -0,0 +1,141 @@
...
+	userdom_use_user_ttys($1_bubblewrap_t)
+
+	ifndef(`enable_mls',`
+		fs_search_removable($1_bubblewrap_t)
+		fs_read_removable_files($1_bubblewrap_t)
+		fs_read_removable_symlinks($1_bubblewrap_t)
+	')
+
+	fs_list_auto_mountpoints($1_bubblewrap_t)
+	files_list_home($1_bubblewrap_t)
```
pebenito: Needs to move up with other fs and files calls.

pebenito (approved):

PR #953, opened 2025-06-26
bootloader-kmod

etbe:


policy/modules/admin/bootloader.te
```diff
@@ -41,10 +41,11 @@ dev_node(bootloader_tmp_t)
 # bootloader local policy
 #
 
-allow bootloader_t self:capability { chown dac_override dac_read_search fsetid mknod setgid sys_admin sys_rawio };
+allow bootloader_t self:capability { chown dac_override dac_read_search fsetid sys_chroot mknod setgid sys_admin sys_rawio };
```
pebenito, line 44: Which bootloader is using chroot?

policy/modules/admin/bootloader.te
```diff
@@ -53,6 +54,7 @@ allow bootloader_t bootloader_etc_t:file read_file_perms;
 
 manage_dirs_pattern(bootloader_t, bootloader_tmp_t, bootloader_tmp_t)
 manage_files_pattern(bootloader_t, bootloader_tmp_t, bootloader_tmp_t)
+allow bootloader_t bootloader_tmp_t:file map;
```
pebenito, line 57: Please change to `mmap_manage_files_pattern()`

policy/modules/admin/bootloader.te
```diff
@@ -78,13 +81,16 @@ storage_raw_read_removable_device(bootloader_t)
 storage_raw_write_removable_device(bootloader_t)
 storage_rw_fuse(bootloader_t)
 
+dev_create_null_dev(bootloader_t)
+dev_delete_null(bootloader_t)
```
pebenito, line 85: How does this come into play?

pebenito (changes requested):
Please add a commit message that has more information. Without knowing which bootloader(s) need this, I'm having trouble evaluating some of these rules.

policy/modules/admin/bootloader.te
etbe: That would be for mkinitramfs which you already rejected. I'll cancel this PR and make a new one with a separate doain for mkinitramfs.

PR #956, opened 2025-06-28
seatd: new policy module

WavyEbuilder:
May need one or two more rules in the future as I've only tested on one compositor I have policy for, but should be pretty much ready to go.

### DIFF
--- a/policy/modules/services/seatd.if
+++ b/policy/modules/services/seatd.if
@@ -1,0 +1,22 @@
+## <summary>Seat management daemon.</summary>
+
+########################################
+## <summary>
+##	Allow the specified domain to connect
+##	to seatd socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to allow
+##	</summary>
+## </param>
+#
+interface(`seatd_stream_connect',`
+	gen_require(`
+		type seatd_t, seatd_runtime_t;
+	')
+
+	allow $1 seatd_t:unix_stream_socket connectto;
+	allow $1 seatd_runtime_t:sock_file write;
+	files_search_runtime($1)
+')
--- a/policy/modules/services/seatd.te
+++ b/policy/modules/services/seatd.te
@@ -1,0 +1,34 @@
+policy_module(seatd)
+
+########################################
+#
+# Declarations
+#
+
+type seatd_t;
+type seatd_exec_t;
+init_daemon_domain(seatd_t, seatd_exec_t)
+# seatd may not have the same level as a client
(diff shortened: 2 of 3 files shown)


policy/modules/services/seatd.te
```diff
@@ -0,0 +1,37 @@
...
+
+auth_use_nsswitch(seatd_t)
+
+dev_getattr_dri_dev(seatd_t)
+dev_getattr_input_dev(seatd_t)
+dev_rw_dri(seatd_t)
+dev_rw_input_dev(seatd_t)
+
+# seatd may not have the same level as a client
+mls_fd_share_all_levels(seatd_t)
```
0xC0ncord: Usually MLS rules go right after the declaration of the type. Take a look at `podman.te` for example.

policy/modules/services/seatd.if
```diff
@@ -0,0 +1,39 @@
...
+##	Domain to allow
+##	</summary>
+## </param>
+#
+interface(`seatd_connect_runtime_sock_files',`
+       	gen_require(`
+               	type seatd_t;
+       	')
+
+        allow $1 seatd_t:unix_stream_socket connectto;
```
0xC0ncord: Would it be better to use `stream_connect_pattern()` here?

0xC0ncord (commented):
Looks good! Just some minor comments.

policy/modules/services/seatd.te
WavyEbuilder: Ah okay, thanks!

policy/modules/services/seatd.if
WavyEbuilder:
```
#
# unix domain socket patterns
#
# Parameters:
# 1. source domain type
# 2. container (directory) type
# 3. socket type
# 4. target domain type
#
define(`stream_connect_pattern',`
        allow $1 $2:dir search_dir_perms;
        allow $1 $3:sock_file write_sock_file_perms;
        allow $1 $4:unix_stream_socket connectto;
')
```

Ah yea this looks much better, I'll use that. Thanks again!

WavyEbuilder: I can't reference `var_run_t` directly in refpolicy I think (i.e. it isn't allowed), and so I don't think I can use the pattern here

0xC0ncord: You would do it like this:
```
files_search_runtime($1)
stream_connect_pattern($1, seatd_runtime_t, seatd_runtime_t, seatd_t)
```

PR #957, opened 2025-06-28
More various fixes

0xC0ncord:
Another collection of various fixes. Of note:

- Add a tunable to allow `kernel_t` to relabel container files, use case being a share with container files on it is exported with the `security_label` option
- Add a tunable to allow `spc_t` to mount/unmount NFS shares
- Add rules needed for Kubernetes and containers to use hugepages, e.g. when using hugepage-backed memory for KubeVirt VMs

### DIFF
--- a/policy/modules/roles/sysadm.te
+++ b/policy/modules/roles/sysadm.te
@@ -35,6 +35,8 @@ ifndef(`enable_mls',`
 
 # for networkctl and possibly other networking tools
 allow sysadm_t self:netlink_route_socket rw_netlink_socket_perms;
+# for debugging BPF programs
+allow sysadm_t self:capability2 bpf;
 
 corecmd_exec_shell(sysadm_t)
 
@@ -97,6 +99,10 @@ ifdef(`init_systemd',`
 	# Allow sysadm to resolve the username of dynamic users by calling
 	# LookupDynamicUserByUID on org.freedesktop.systemd1.
 	init_dbus_chat(sysadm_t)
+
+	# Debug BPF programs.
+	init_rw_bpf(sysadm_t)
+	init_run_bpf(sysadm_t)
 
 	# Allow sysadm to get the status of and set properties of other users,
 	# sessions, and seats on the system.
@@ -305,6 +311,12 @@ optional_policy(`
 
 optional_policy(`
 	container_admin(sysadm_t, sysadm_r)
+
+	# Debug BPF programs.
+	container_run_system_engine_bpf(sysadm_t)
+	container_rw_system_engine_bpf(sysadm_t)
+	container_run_system_container_bpf(sysadm_t)
+	container_rw_system_container_bpf(sysadm_t)
 ')
 
 optional_policy(`
--- a/policy/modules/services/container.if
+++ b/policy/modules/services/container.if
@@ -441,6 +441,42 @@ interface(`container_engine_dbus_chat',`
 
 ########################################
 ## <summary>
+##	Run system container engine BPF programs.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed access.
+## </summary>
+## </param>
+#
+interface(`container_run_system_engine_bpf',`
+	gen_require(`
+		attribute container_engine_system_domain;
+	')
+
+	allow $1 container_engine_system_domain:bpf prog_run;
+')
+
+########################################
+## <summary>
+##	Read and write system container engine BPF programs.
+## </summary>
+## <param name="domain">
+## <summary>
+##	Domain allowed access.
+## </summary>
+## </param>
+#
+interface(`container_rw_system_engine_bpf',`
+	gen_require(`
+		attribute container_engine_system_domain;
+	')
+
+	allow $1 container_engine_system_domain:bpf { map_read map_write };
+')
+
+########################################
+## <summary>
 ##	Allow the specified domain to be started
 ##	by systemd socket activation using a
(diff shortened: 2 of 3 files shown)


policy/modules/services/container.te
```diff
@@ -319,7 +335,7 @@ allow container_domain container_file_t:chr_file map;
 allow container_domain container_ro_file_t:blk_file read_blk_file_perms;
 allow container_domain container_ro_file_t:dir list_dir_perms;
 allow container_domain container_ro_file_t:chr_file read_chr_file_perms;
-allow container_domain container_ro_file_t:file { exec_file_perms read_file_perms };
+allow container_domain container_ro_file_t:file { exec_file_perms read_file_perms execmod };
```
pebenito: I don't think we want to do this by default. This should be a very rare thing these days.

policy/modules/services/sasl.if
```diff
@@ -57,6 +57,24 @@ interface(`sasl_mmap_read_keytab',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`sasl_lock_keytab',`
+	gen_require(`
+		type saslauthd_keytab_t;
+	')
+
+	allow $1 saslauthd_keytab_t:file lock;
```
pebenito, line 77: Needs the parent path access.

policy/modules/services/container.if
```diff
@@ -1362,6 +1362,24 @@ interface(`container_watch_dirs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`container_ioctl_dirs',`
+	gen_require(`
+		type container_file_t;
+	')
+
+	allow $1 container_file_t:dir ioctl;
```
pebenito, line 1453: Needs parent dir access.

policy/modules/services/container.if
```diff
@@ -1686,6 +1704,25 @@ interface(`container_getattr_all_ro_files',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`container_getattr_tmpfs_files',`
+	gen_require(`
+		type container_tmpfs_t;
+	')
+
+	allow $1 container_tmpfs_t:file getattr;
```
pebenito: needs parent dir access

policy/modules/kernel/filesystem.if
```diff
@@ -3439,6 +3493,42 @@ interface(`fs_list_hugetlbfs',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`fs_setattr_hugetlbfs_dirs',`
+	gen_require(`
+		type hugetlbfs_t;
+	')
+
+	allow $1 hugetlbfs_t:dir setattr_dir_perms;
```
pebenito: Here and the below hunk needs parent dir access.

policy/modules/roles/sysadm.te
```diff
@@ -44,6 +46,9 @@ corenet_ib_manage_subnet_unlabeled_endports(sysadm_t)
 dev_read_kmsg(sysadm_t)
 dev_rw_ipmi_dev(sysadm_t)
 
+domain_run_bpf_all_domains(sysadm_t)
```
pebenito: I'm unsure we want this. Can you elaborate on the use?

policy/modules/kernel/domain.if
```diff
@@ -569,6 +569,44 @@ interface(`domain_kill_all_domains',`
...
+##	Run any domain BPF program.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`domain_run_bpf_all_domains',`
```
pebenito: `domain_run_all_domains_bpf()`

policy/modules/kernel/domain.if
```diff
@@ -569,6 +569,44 @@ interface(`domain_kill_all_domains',`
...
+##	Read and write any domain BPF program.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`domain_rw_bpf_all_domains',`
```
pebenito: `domain_rw_all_domains_bpf()`

policy/modules/roles/sysadm.te
0xC0ncord: This is a very coarse way of allowing `sysadm_t` to run `bpftool prog` to list all BPF programs. Alternatively I can split this out into interfaces for each domain where I know I've seen this hit, namely `init_t` and various container-related domains.

policy/modules/services/container.te
0xC0ncord: Thanks. I made this into a tunable instead.

0xC0ncord:
I've dropped the fix implemented by #955 from this series.

policy/modules/roles/sysadm.te
pebenito: Sorry, I failed to put my full concern, which is for untrusted stuff like `user_t` bpf.

0xC0ncord: Ahh, I understand now. I'll go ahead and split this out so it's not so coarse then.

0xC0ncord: For now `sysadm_t` gets BPF access to `init_t` and the system container domains.

PR #959, opened 2025-06-29
ssh fixes

aerusso:
This is a small collection of fix-ups to the ssh policy that I think are sufficiently general to warrant inclusion here in refpolicy.

I probably messed up the style in these, so I'm happy to restyle these as desired.


policy/modules/services/ssh.te
```diff
@@ -174,6 +181,13 @@ userdom_use_user_terminals(ssh_t)
 # needs to read krb tgt
 userdom_read_user_tmp_files(ssh_t)
 
+# ssh re-executes itself during ProxyJump
+ssh_exec(ssh_t)
+
+tunable_policy(`ssh_connect_any_port',`
+	corenet_tcp_connect_generic_port(ssh_t)
```
pebenito: This doesn't seem to match the documented intent of the tunable above. I'd expect it to call one of the interfaces that have `all` in the name.

policy/modules/apps/gpg.if
```diff
@@ -287,6 +287,7 @@ interface(`gpg_stream_connect_agent',`
 
 	stream_connect_pattern($1, gpg_agent_tmp_t, gpg_agent_tmp_t, gpg_agent_t)
 	allow $1 { gpg_secret_t gpg_runtime_t }:dir search_dir_perms;
+	allow $1 { gpg_secret_t gpg_runtime_t }:lnk_file read;
```
pebenito: Please use `read_lnk_file_perms`'

pebenito (changes requested):

pebenito (approved):

PR #960, opened 2025-06-29
Initial policy for Kanidm

0xC0ncord:
Add an initial policy implementation for [Kanidm](https://github.com/kanidm/kanidm).

From the README:

> Kanidm is a simple and secure identity management platform, allowing other applications and  services to offload the challenge of authenticating and storing identities to Kanidm.
>
> The goal of this project is to be a complete identity provider, covering the broadest possible set of requirements and integrations. You should not need any other components (like Keycloak) when you use Kanidm - we already have everything you need!
>
> To achieve this we rely heavily on strict defaults, simple configuration, and self-healing components. This allows Kanidm to support small home labs, families, small businesses, and all the way to the largest enterprise needs.
>
> If you want to host your own authentication service, then Kanidm is for you!

It provides the Kanidm server itself as well as the optional Kanidm-unixd client which implements support for UNIX authentication via PAM.

### DIFF
--- a/policy/modules/services/kanidm.if
+++ b/policy/modules/services/kanidm.if
@@ -1,0 +1,183 @@
+## <summary>A simple, secure and fast identity management platform</summary>
+
+########################################
+## <summary>
+##	Execute kanidmd in the kanidmd domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`kanidm_domtrans',`
+	gen_require(`
+		type kanidmd_t, kanidmd_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, kanidmd_exec_t, kanidmd_t)
+')
+
+########################################
+## <summary>
+##	Execute kanidmd in the kanidmd domain, and
+##	allow the specified role the kanidmd domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`kanidm_run',`
+	gen_require(`
+		type kanidmd_t;
+	')
+
+	kanidm_domtrans($1)
+	role $2 types kanidmd_t;
+')
+
+########################################
+## <summary>
+##	Execute kanidmd-unixd in the
+##	kanidmd-unixd domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`kanidm_domtrans_unixd',`
+	gen_require(`
+		type kanidm_unixd_t, kanidm_unixd_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, kanidm_unixd_exec_t, kanidm_unixd_t)
+')
+
+########################################
+## <summary>
+##	Execute kanidm-unixd in the kanidm-unixd
+##	domain, and allow the specified role the
+##	kanidm-unixd domain.
+## </summary>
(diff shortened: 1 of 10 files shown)


policy/modules/services/kanidm.if
```diff
@@ -0,0 +1,183 @@
...
+## <summary>
+##	Execute kanidmd in the kanidmd domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`kanidm_domtrans_server',`
```
pebenito: I'd prefer to drop the `server` here and below.

policy/modules/services/kanidm.te
```diff
@@ -0,0 +1,229 @@
...
+##  </p>
+## </desc>
+gen_tunable(kanidm_unixd_create_home_dirs, false)
+
+## <desc>
+##  <p>
+##  Determine whether kanidm-unixd-tasks can
+##	modify the SELinux policy. This is to create
+##	local equivalence file context rules which map
+##	usernames to their UUID, SPN, or other identifier.
```
pebenito, line 21: whitespace issue

policy/modules/services/kanidm.if
```diff
@@ -0,0 +1,183 @@
...
+## <rolecap/>
+#
+interface(`kanidm_run_server',`
+	gen_require(`
+		type kanidmd_t;
+	')
+
+	kanidm_domtrans_server($1)
+	role $2 types kanidmd_t;
+')
```
pebenito: Is this really intended to be run interactively?

policy/modules/system/init.te
```diff
@@ -37,6 +37,13 @@ gen_tunable(init_daemons_use_tty, false)
 ## </desc>
 gen_tunable(init_mounton_non_security, false)
 
+## <desc>
+## <p>
+## Enable systemd to mount on the SELinux config directories.
```
pebenito, line 42: What does kanidm do here?

pebenito (changes requested):

policy/modules/services/kanidm.if
0xC0ncord: For a very small number of cases, yes. One in particular is to set up the initial default admin accounts after starting up the server for the first time with `kanidmd recover-account <ACCOUNT>`: https://kanidm.github.io/kanidm/stable/server_configuration.html#default-admin-accounts

policy/modules/system/init.te
0xC0ncord: This is to support `kanidm-unixd`'s SELinux feature. When a Kanidm user logs in, their home directory is created in `/home/<UUID>` and a symlink with their username pointing to it at `/home/<username>`. The SELinux feature creates an equivalence rule in the policy so that `/home/<username>` equals `/home/<UUID>`, but the policy needs to be writable.

This tunable allows an administrator to add `ReadWritePaths=/etc/selinux` to the systemd unit which by default has `ProtectSystem=strict` set. Then, `init` will create a read-writable bind mount in the service's mount namespace to support this.

0xC0ncord: To clarify, this tunable here is generic. In the context of Kanidm one would need to enable both this (if using systemd and upstream's systemd unit) and `kanidm_unixd_modify_policy` for it to work.

policy/modules/services/kanidm.te
```diff
@@ -0,0 +1,229 @@
...
+kernel_read_vm_overcommit_sysctl(kanidm_unixd_t)
+
+auth_use_nsswitch(kanidm_unixd_t)
+
+miscfiles_read_generic_certs(kanidm_unixd_t)
+miscfiles_read_localization(kanidm_unixd_t)
+
+logging_send_syslog_msg(kanidm_unixd_t)
+
+userdom_use_user_terminals(kanidm_unixd_t)
```
pebenito: This should only be inherited terminals.

pebenito (approved):

PR #962, opened 2025-07-01
portage: gatekeep portage_fetch_t accessing all ports behind a tunable

WavyEbuilder:
For the default (and most) configurations of portage, portage shouldn't require unrestricted access to ports when fetching distfiles, so set the tunable to false by default.


0xC0ncord:
Please also add `corenet_tcp_connect_http_port()` and `corenet_sendrecv_http_client_packets()` outside the tunable to make sure we don't lose standard HTTP(s) connectivity. ;)

WavyEbuilder:
> Please also add `corenet_tcp_connect_http_port()` and `corenet_sendrecv_http_client_packets()` outside the tunable to make sure we don't lose standard HTTP(s) connectivity. ;)

Will do soon, just currently searching through SRC_URI to see what other ports are used (notably ftp so far).

PR #965, opened 2025-07-08
chromium

etbe:


policy/modules/apps/chromium.te
```diff
@@ -90,7 +90,7 @@ xdg_cache_content(chromium_xdg_cache_t)
 #
 
 # execmem for load in plugins
-allow chromium_t self:process { execmem getcap getsched setcap setrlimit setsched sigkill signal signull };
+allow chromium_t self:process { execmem execheap getcap getsched setcap setrlimit setsched sigkill signal signull ptrace };
```
pebenito: It would be great if these two could be moved to tunable(s).

policy/modules/apps/chromium.te
```diff
@@ -276,7 +292,11 @@ optional_policy(`
 	')
 
 	optional_policy(`
+		systemd_list_resolved_runtime(chromium_t)
 		systemd_dbus_chat_hostnamed(chromium_t)
+		systemd_dbus_chat_logind(chromium_t)
+		systemd_start_all_user_sessions(chromium_t)
+		systemd_status_all_user_sessions(chromium_t)
```
pebenito: What are these used for?

policy/modules/apps/wm.if
```diff
@@ -184,6 +184,42 @@ interface(`wm_dontaudit_exec_tmp_files',`
...
+##	</summary>
+## </param>
+#
+interface(`wm_map_tmpfs_files',`
+	gen_require(`
+		type wm_tmpfs_t;
+	')
+
+	allow $1 wm_tmpfs_t:file map;
+')
```
pebenito: I'd prefer to merge these into a single `wm_mmap_rw_tmpfs_files()`. Also please add parent dir access.

policy/modules/apps/chromium.te
etbe: I'll remove those from the pr and put them in later with more info.

pebenito (approved):

PR #966, opened 2025-07-08
unconfined: Promote anon_inode access to full access.

pebenito:
This class inherits the file common.

Also see https://lore.kernel.org/selinux/48916a70-2a89-4d24-8e36-d15ccc112519@ieee.org/

### DIFF
--- a/policy/modules/system/unconfined.if
+++ b/policy/modules/system/unconfined.if
@@ -50,8 +50,7 @@ interface(`unconfined_domain_noaudit',`
 	# Write access is for setting attributes under /proc/self/attr.
 	allow $1 self:file rw_file_perms;
 
-	# io_uring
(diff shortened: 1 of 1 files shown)


policy/modules/system/unconfined.if
```diff
@@ -51,7 +51,7 @@ interface(`unconfined_domain_noaudit',`
 	allow $1 self:file rw_file_perms;
 
 	# io_uring
-	allow $1 self:anon_inode { create map read write };
+	allow $1 self:anon_inode { manage_file_perms mounton quotaon relabel_file_perms watch watch_mount watch_reads watch_sb watch_with_perm };
```
pcmoore: Anonymous inodes are used by more than just io_uring, see the recent secretmem discussion, so it might be a good idea to update that comment while you are at it.

PR #968, opened 2025-07-09
cloudinit: Add container engine admin access.

pebenito:


0xC0ncord:
Looks good to me. Good catch on the unneeded `corecmd_bin_domtrans()`.

PR #969, opened 2025-07-10
wayland: new policy module

WavyEbuilder:
A new type attribute wayland_compositor is introduced for shared rules for Wayland compositors. To support the security-context protocol[1], two new type attributes are introduced: wayland_client for regular Wayland clients, and wayland_client_sandboxed for sandboxed Wayland clients for use in applications run by sandbox engines such as Flatpak that support security-context[2].

As a fair amount of new policy modules can be expected to work with modern Wayland desktop sessions, a new policy layer has be created, session, to contain these new policy modules.

[1] https://wayland.app/protocols/security-context-v1
[2] https://github.com/flatpak/flatpak/pull/4920


policy/modules/session/wayland.te
```diff
@@ -0,0 +1,77 @@
...
+attribute wayland_client_sandboxed;
+
+type wayland_runtime_t;
+files_runtime_file(wayland_runtime_t)
+userdom_user_runtime_content(wayland_runtime_t)
+
+# For use by sandbox engines implementing the security_context protocol.
+type wayland_runtime_sandboxed_t;
+files_runtime_file(wayland_runtime_t)
+userdom_user_runtime_content(wayland_runtime_t)
```
0xC0ncord: I think you meant `wayland_runtime_sandboxed_t` here.

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,77 @@
...
+
+# No need to distinguish between the shm of sandboxed and unsandboxed clients;
+# all clients should have their own shm type, this attribute is only to group them
+# and grant wayland_compositor access to them.
+attribute wayland_client_tmpfs_type;
```
0xC0ncord, line 29: Would it be appropriate to go ahead and add `files_tmpfs_file()` here?

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,77 @@
...
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-0")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-1")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-2")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-3")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-4")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-5")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-6")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-7")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-8")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, socket, "wayland-9")
```
0xC0ncord: `sock_file` instead of `socket`.

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,77 @@
...
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-6.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-7.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-8.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-9.lock")
+
+# GPU Access
+dev_rw_dri(wayland_compositor)
+
+# For seats
+systemd_use_logind_fds(wayland_compositor)
```
0xC0ncord: This needs to be wrapped in `ifdef('init_systemd')`.

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,77 @@
...
+dev_rw_input_dev(wayland_compositor)
+
+files_read_usr_files(wayland_compositor)
+
+# Fonts
+miscfiles_read_fonts(wayland_compositor)
+xserver_use_user_fonts(wayland_compositor)
+
+xserver_read_xkb_libs(wayland_compositor)
+xserver_rw_mesa_shader_cache(wayland_compositor)
```
0xC0ncord, line 76: These `xserver_()` calls should go into an optional.

On the other hand I know it's perfectly possible to run a Wayland desktop without Xorg at all, so maybe this highlights a need to split some of these file types out of `xserver` in the future.

WavyEbuilder: Maybe that's worth doing now (or possibly noting a dependency on the xserver module for now)? A compositor that doesn't have access to the mesa shader cache is pretty much useless...

policy/modules/session/wayland.te
WavyEbuilder: Yea, seems reasonable, will do

WavyEbuilder:
Hey,

> I've been building up a patchset that gets kde on wayland working with a slightly different set of primitives than this introduces. I haven't opened a PR for it partly because I hadn't yet figured out the best way to capture some of the primitives (which are captured here).

Nice!

> If this patch set is expected to be stable, I'll start re-basing my changes on this, and open a PR soon.

One thing I'm solving (and the final concern I have at the moment) is the shared memory (tmpfs_t) of the compositors, so I'm going to make an attribute for that tonight or latest tomorrow. After that all the interfaces should remain stable, ofc I suspect we'll in the future be able to remove some modules from kwin and future compositor modules and consolidate them into the typeattribute more as we get more practical experience of the accesses shared by compositors, but the interfaces themselves should remain stable.

> And, thanks for getting this started!

No worries! Glad it could be of use. :)

policy/modules/session/metadata.xml
```diff
@@ -0,0 +1 @@
+<summary>Policy modules for desktop sessions.</summary>
```
pebenito, line 1: How would this differ from `apps`?

policy/modules/session/wayland.if
```diff
@@ -0,0 +1,79 @@
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
```
pebenito, line 31: This doesn't seem to fit `wayland_compositor_tmpfs_type`.

policy/modules/session/wayland.if
```diff
@@ -0,0 +1,79 @@
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+## 	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+## 	</summary>
+## </param>
```
pebenito, line 69: This doesn't seem to fit `wayland_client_tmpfs_type`.

policy/modules/session/wayland.if
```diff
@@ -0,0 +1,79 @@
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+## 	</summary>
+## </param>
```
pebenito, line 93: Same as above

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,84 @@
...
+attribute wayland_client_tmpfs_type;
+
+##############################
+#
+# Local Policy
+#
+
+allow wayland_client wayland_compositor_tmpfs_type:file { mmap_rw_inherited_file_perms };
+
+allow wayland_compositor wayland_client_tmpfs_type:file { mmap_rw_inherited_file_perms };
```
pebenito: nit: please drop the `{` `}`.

pebenito (changes requested):

policy/modules/session/wayland.te
pebenito: Perhaps we should generalize the X fonts to simply user fonts, in userdomain or xdg? (suggestions please).

policy/modules/session/wayland.te
pebenito: Is this resolved? I don't see a change.

policy/modules/session/metadata.xml
WavyEbuilder: So the logic is that apps would get crowded with stuff that constitutes a desktop session.  Given that the aim is to reasonably confine a fair amount of stuff from a desktop session, including but not limited to:

- Pipewire
- xdg-desktop-portal (and the various backends)
- Compositors (kwin, sway, mutter, etc)
- Screen lockers (swaylock, kscreenlocker, gtklock, waylock, etc)
- layershell clients (launchers such as krunner, fuzzel, etc)
- polkit agents

etc. The most important distinction to be made here in my opinion is that these components are privileged - they make use of some privileged Wayland protocols or desktop session components of some kind, xdg-desktop-portal uses plenty of private compositor apis, compositors have access to e.g. all keyboard events, screenlockers can use the privileged Wayland screenlocker protocols (differ from compositor to compositor but are nonetheless privileged on each one)

I also think it marks a nice separation from constitutes a desktop session (such as some of the things above) and what uses those things (clients/apps).

However, I'm happy to keep things in the apps layer if you wish.

policy/modules/session/wayland.te
WavyEbuilder: Forgot to reply, this interface doesn't appear to work on attributes?

policy/modules/session/wayland.te
WavyEbuilder: xdg sounds good to me, would you like a future PR for that?

policy/modules/session/wayland.if
WavyEbuilder: Should be resolved.

policy/modules/session/wayland.if
WavyEbuilder: Should be resolved.

policy/modules/session/wayland.if
WavyEbuilder: Should be resolved.

policy/modules/session/wayland.te
```diff
@@ -0,0 +1,84 @@
...
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-6.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-7.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-8.lock")
+userdom_user_runtime_filetrans(wayland_compositor, wayland_runtime_t, file, "wayland-9.lock")
+
+# GPU Access
+dev_rw_dri(wayland_compositor)
+
+# /dev/input
+dev_rw_input_dev(wayland_compositor)
```
aerusso: This is a nit, but this domain will be used by things like headless compositors and things like waypipe (which I have mostly working, and am planning on re-basing on this over the weekend).

Those domains won't need input access, and I would argue it makes more sense that those individual domains should include this intentionally, rather than as a blanket.  It's also quite powerful, and a route to inject/sniff keystrokes (unless I'm mistaken), and therefore particularly scary if the domain is also network controlled.

I'd contrast this with the dri access, which is presumably used even on headless compositors.

policy/modules/session/wayland.fc
```diff
@@ -0,0 +1,2 @@
+/run/user/%{USERID}/wayland-[0-9]+	-s	gen_context(system_u:object_r:wayland_runtime_t,s0)
+/run/user/%{USERID}/wayland-[0-9]+\.lock	--	gen_context(system_u:object_r:wayland_runtime_t,s0)
```
aerusso: It may not be super important, but waypipe includes alphabet characters here, too.

policy/modules/session/wayland.te
WavyEbuilder: Ah okay, that sounds reasonable then, and it isn't too big of a change - I'll drop that.

policy/modules/session/wayland.fc
WavyEbuilder: Thanks - I'll see what they do for sockets and fix up the filecon then.

policy/modules/session/wayland.if
pebenito: Perhaps I need to look more at wayland internals... is there a process to handle tmpfs stuff? The attribute naming seems like it should be a file type.

policy/modules/session/wayland.if
pebenito: Same question as above.

policy/modules/session/wayland.te
pebenito: No, unfortunately, it's a language limitation. You'd have to put it in the interface.

policy/modules/session/wayland.te
pebenito: Sure

policy/modules/session/wayland.if
WavyEbuilder: Okay so basically it's `/dev/shm` - Wayland clients have shared memory with the compositor that is written and read to. Take a look at https://wayland.app/protocols/wayland#wl_shm for associated interfaces. The compositor and the app have the tmpfs so basically imagine I have an app foo and a compositor bar, I'd have `foo_t` and `foo_tmpfs_t`, as well as `bar_t` and `bar_tmpfs_t`. They need to be able to rw each other's shared memory.

policy/modules/session/wayland.if
WavyEbuilder: Should be answered above.

policy/modules/session/wayland.if
pebenito: In that case, the doc is wrong (parameter name and summary), as the parameter is to be used as a file type, not a domain.

WavyEbuilder:
```
allow wayland_client wayland_compositor_tmpfs_type:file mmap_rw_inherited_file_perms;
allow wayland_client wayland_compositor:unix_stream_socket connectto;
allow wayland_client wayland_compositor:fd use;
allow wayland_client wayland_runtime_t:sock_file rw_sock_file_perms;
allow wayland_client wayland_client:fifo_file write;

allow wayland_compositor wayland_client_tmpfs_type:file mmap_rw_inherited_file_perms;
allow wayland_compositor wayland_client:fd use;
allow wayland_compositor wayland_client:fifo_file write;
allow wayland_compositor wayland_client_sandboxed:fd use;
```
These don't seem particularly egregious - what's up with that?

WavyEbuilder:
@aerusso @pebenito any updates?

aerusso:
@WavyEbuilder I've cherry picked your commit, and combined it with some other work of mine to get waypipe, kde, Xwayland, and basically all my other wayland applications working.

I was waiting for this patch to finalize, and then I was going to rebase (and clean up) those patches, and then submit them.  The KDE patches are pretty significant, so those would come second.   There are core improvements (e.g., Xwayland) and other somewhat simpler things (waypipe) that would be good examples to help work out an interface for the new module before diving into all of KDE.

PR #970, opened 2025-07-11
add extra # characters to commented interfaces

aerusso:
When `make config` is called, the interfaces are passed through xmllint, which validates that the docstrings are appropriately formatted, even if the interface body has been commented out.  This causes validation errors, since these spurious docstrings are concactenated with the following docstring.

By left-padding the commented-out docstrings with another # character, the docstrings are not used, and not checked by xmllint, allowing the test to succeed.


pebenito:
Thanks for catching this. The interfaces were removed instead in #972 .

PR #971, opened 2025-07-11
git client domain fixes

aerusso:
Using the git client domain comes with some pain points.  This series adds some tunables and defaults that make it much more functional.

I don't fully understand the various user_git_t domains, so I'm not sure if it makes sense to have user-specific temporary directories. (Though, given the ability to run user-controlled scripts, I suspect it does make sense to restrict inter-user temporary file reads).

I also had to add an interface to gpg, and I'm not sure it's idiomatic, so I suspect it may need to be adjusted to fit in with the overall project style.

### DIFF
--- a/policy/modules/apps/gpg.if
+++ b/policy/modules/apps/gpg.if
@@ -518,6 +518,26 @@ interface(`gpg_list_user_secrets',`
 
 ########################################
 ## <summary>
+##	Allow gpg to read a file type, and therefore encrypt,
+##	decrypt, sign, and validate files of that type.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain gpg is allowed access to.
+##	</summary>
+## </param>
+#
+interface(`gpg_crypto_content',`
+	gen_require(`
+		type gpg_t;
+	')
+
+	allow gpg_t $1:file read_file_perms;
+	allow gpg_t $1:dir search_dir_perms;
+')
+
+########################################
+## <summary>
 ##	Do not audit attempt to search gpg user secrets dirs.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/services/git.fc
+++ b/policy/modules/services/git.fc
@@ -7,7 +7,9 @@ HOME_DIR/\.git-credentials	--	gen_context(system_u:object_r:git_xdg_config_t,s0)
 /usr/bin/git-[^/]+	--	gen_context(system_u:object_r:git_exec_t,s0)
 /usr/bin/git2_cli	--	gen_context(system_u:object_r:git_exec_t,s0)
 
-/usr/lib/git-core/git-daemon	--	gen_context(system_u:object_r:gitd_exec_t,s0)
+/usr/lib/git-core/git	--	gen_context(system_u:object_r:git_exec_t,s0)
+/usr/lib/git-core/git-[^/]+ --	gen_context(system_u:object_r:git_exec_t,s0)
+/usr/lib/git-core/git-daemon --	gen_context(system_u:object_r:gitd_exec_t,s0)
 
 /usr/libexec/git-core/git	--	gen_context(system_u:object_r:git_exec_t,s0)
 /usr/libexec/git-core/git-[^/]+ --	gen_context(system_u:object_r:git_exec_t,s0)
--- a/policy/modules/services/git.if
+++ b/policy/modules/services/git.if
@@ -92,6 +92,7 @@ template(`git_client_role_template',`
 	gen_require(`
 		attribute git_client_domain;
 		type git_exec_t, git_home_t, git_home_hook_t;
+		type git_xdg_config_t;
 	')
 
 	########################################
@@ -117,6 +118,9 @@ template(`git_client_role_template',`
 	allow $2 git_home_hook_t:dir { manage_dir_perms relabel_dir_perms };
 	allow $2 git_home_hook_t:file { exec_file_perms manage_file_perms relabel_file_perms };
 	filetrans_pattern($2, git_home_t, git_home_hook_t, dir, "hooks")
+	xdg_config_filetrans($2, git_xdg_config_t, dir, "git")
+	userdom_user_home_dir_filetrans($2, git_xdg_config_t, file, ".gitconfig")
+	userdom_user_home_dir_filetrans($2, git_xdg_config_t, file, ".git-credentials")
 
 	allow $3 $1_git_t:process { ptrace signal_perms };
 	ps_process_pattern($3, $1_git_t)
--- a/policy/modules/services/git.if
+++ b/policy/modules/services/git.if
@@ -131,6 +131,10 @@ template(`git_client_role_template',`
 	exec_files_pattern($3, git_home_hook_t, git_home_hook_t)
 	# transition back to the user domain when executing git hooks
 	domtrans_pattern($1_git_t, git_home_t, $2)
+	# execute shell scripts
+	corecmd_exec_shell($1_git_t)
+	# execute user utilities, e.g., editor
+	corecmd_bin_domtrans($1_git_t, $2)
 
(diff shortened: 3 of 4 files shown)


policy/modules/services/git.fc
```diff
@@ -7,10 +7,11 @@ HOME_DIR/\.git-credentials	--	gen_context(system_u:object_r:git_xdg_config_t,s0)
...
 /usr/bin/git2_cli	--	gen_context(system_u:object_r:git_exec_t,s0)
 
-/usr/lib/git-core/git-daemon	--	gen_context(system_u:object_r:gitd_exec_t,s0)
-
+/usr/lib/git-core/git	--	gen_context(system_u:object_r:git_exec_t,s0)
 /usr/libexec/git-core/git	--	gen_context(system_u:object_r:git_exec_t,s0)
+/usr/lib/git-core/git-[^/]+ --	gen_context(system_u:object_r:git_exec_t,s0)
 /usr/libexec/git-core/git-[^/]+ --	gen_context(system_u:object_r:git_exec_t,s0)
+/usr/lib/git-core/git-daemon --	gen_context(system_u:object_r:gitd_exec_t,s0)
 /usr/libexec/git-core/git-daemon --	gen_context(system_u:object_r:gitd_exec_t,s0)
```
0xC0ncord, line 16: This is just my personal taste but I think this would be easier to read if the `/usr/lib` and `/usr/libexec` portions were separate sections.

policy/modules/services/git.if
```diff
@@ -117,16 +118,41 @@ template(`git_client_role_template',`
...
+	allow $2 $1_git_tmp_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
+	allow $1_git_t $1_git_tmp_t:dir manage_dir_perms;
+	allow $1_git_t $1_git_tmp_t:file mmap_manage_file_perms;
+	allow $1_git_t $1_git_tmp_t:lnk_file manage_lnk_file_perms;
+	files_tmp_filetrans($1_git_t, $1_git_tmp_t, {dir file})
+
 	# allow userdomains to exec git hooks
 	exec_files_pattern($3, git_home_hook_t, git_home_hook_t)
 	# transition back to the user domain when executing git hooks
 	domtrans_pattern($1_git_t, git_home_t, $2)
```
0xC0ncord, line 144: Could you please add `corecmd_bin_domtrans($1_git_t, $2)` and see if that alleviates the need for a separate `$2_git_tmp_t` type?

I'm not opposed to a separate tmp type for the git client, but it introduces some additional complexity that would be nice to avoid if we can.

0xC0ncord (commented):
> I don't fully understand the various user_git_t domains, so I'm not sure if it makes sense to have user-specific temporary directories.

The primary reason I wrote it this way was because `git` executes a user's editor of choice for things like interactive rebasing, but it looks like I forgot to add that back-transition to the user domain in the policy when I wrote it. I addressed that in a below comment.

policy/modules/services/git.if
aerusso: Unfortunately, the temporary file is created by the git binary itself.  For example, when you use `git tag -v`, `verify_gpg_signed_buffer` in `gpg-interface.c` makes a `temp = mks_tempfile_t(".git_vtag_tmpXXXXXX");` call that creates a temporary file.

There may be other reasons to do transitions to other domains, but this cannot help the need to handle temporary file creation by the git binary itself.

0xC0ncord: This should be fine then, but we still need `corecmd_bin_domtrans($1_git_t, $2)` to make sure the user's editor runs in the correct context.

aerusso: Also done! I was originally trying to provide some protection from a rogue (e.g.,) staff_git_t process, and this would trivially allow staff_t elevation.

I'm still trying to understand what exactly we're trying to stop/allow with each policy.

policy/modules/apps/gpg.if
```diff
@@ -516,6 +516,25 @@ interface(`gpg_list_user_secrets',`
...
+## <summary>
+##	Allow gpg to read a file type.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain gpg is allowed access to.
+##	</summary>
+## </param>
+#
+interface(`gpg_read_files',`
```
pebenito: This would be better with a name that isn't confused with an interface that gives access to gpg files. Something like `gpg_encrypted_content()`. It's a bit imprecise in this PR's case, since it is simply signing git commits, but it has clearer intent. The docs can be updated that it's for encryption and signing.

policy/modules/services/git.if
```diff
@@ -117,16 +118,43 @@ template(`git_client_role_template',`
...
 	allow $3 $1_git_t:process { ptrace signal_perms };
 	ps_process_pattern($3, $1_git_t)
 
 	auth_use_nsswitch($1_git_t)
 
+	type $1_git_tmp_t;
+	userdom_user_tmp_file($1_git_tmp_t)
+
+	allow $2 $1_git_tmp_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $2 $1_git_tmp_t:file { exec_file_perms manage_file_perms relabel_file_perms };
```
pebenito, line 134: How is the exec used? This is an obvious path for arbitrary code execution.

pebenito (changes requested):

policy/modules/apps/gpg.if
aerusso: Renamed to `gpg_crypto_content`.

policy/modules/services/git.if
aerusso: Removed, and tested for a while now.

aerusso:
Another branch I failed to push.

policy/modules/apps/gpg.if
```diff
@@ -516,6 +516,26 @@ interface(`gpg_list_user_secrets',`
...
+##	</summary>
+## </param>
+#
+interface(`gpg_crypto_content',`
+	gen_require(`
+		type gpg_t;
+	')
+
+	allow gpg_t $1:file read_file_perms;
+	allow gpg_t $1:dir search_dir_perms;
```
pebenito, line 536: nit: please use `read_files_pattern()`.

policy/modules/services/git.if
```diff
@@ -117,16 +118,43 @@ template(`git_client_role_template',`
...
+	userdom_user_home_dir_filetrans($2, git_xdg_config_t, file, ".gitconfig")
+	userdom_user_home_dir_filetrans($2, git_xdg_config_t, file, ".git-credentials")
 
 	allow $3 $1_git_t:process { ptrace signal_perms };
 	ps_process_pattern($3, $1_git_t)
 
 	auth_use_nsswitch($1_git_t)
 
+	type $1_git_tmp_t;
+	userdom_user_tmp_file($1_git_tmp_t)
```
pebenito, line 131: This should be up in the declaration section

policy/modules/services/git.if
pebenito: You removed an `exec_file_perms` from a hunk at the top of the file, but not here.

policy/modules/services/git.if
```diff
@@ -117,16 +118,43 @@ template(`git_client_role_template',`
 	allow $2 git_home_hook_t:dir { manage_dir_perms relabel_dir_perms };
 	allow $2 git_home_hook_t:file { exec_file_perms manage_file_perms relabel_file_perms };
 	filetrans_pattern($2, git_home_t, git_home_hook_t, dir, "hooks")
+	xdg_config_filetrans($2, git_xdg_config_t, dir, "git")
```
pebenito, line 121: nit: add a blank line above this one.

pebenito (changes requested):
mostly nits

PR #974, opened 2025-07-12
files: add a default file context spec for /proc

WavyEbuilder:
There are many situations where having a default label for /proc is helpful in early boot, generally before /proc is mounted in cases such as plymouth. Having a /proc label also doesn't really hurt here, and makes labelling more robust before the system is fully initalised.

Another example of this being useful is for mounting namespaces, such as used by systemd for auditd:

[   19.902620] audit: type=1400 audit(1663630933.439:3): avc:  denied  { mounton } for  pid=1062 comm="(auditd)" path="/run/systemd/unit-root/proc" dev="dm-3" ino=67581 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=dir permissive=1

Closes: https://bugs.gentoo.org/871966

### DIFF
--- a/policy/modules/kernel/files.fc
+++ b/policy/modules/kernel/files.fc
@@ -142,12 +142,6 @@ HOME_ROOT/lost\+found/.*	<<none>>
 /opt/(.*/)?var/lib(64)?(/.*)?	gen_context(system_u:object_r:var_lib_t,s0)
 
 #
-# /proc
-#
-/proc			-d	<<none>>
-/proc/.*			<<none>>
-
-#
 # /run
 #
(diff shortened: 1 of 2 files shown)


WavyEbuilder:
> files.fc:           147: (S): Type proc_t is declared in module kernel, but used in file context here. (S-002)

Hmm, not sure what to do about this as labelling of the labelling of `/proc` is defined in the files module, maybe move it to the kernel policy module? @pebenito would you have any thoughts?

pebenito (changes requested):
I'm fine with the change, but please move the `/proc` block to `kernel.fc`.

PR #979, opened 2025-07-15
acpi

etbe:


policy/modules/services/acpi.te
```diff
@@ -64,7 +64,10 @@ logging_send_syslog_msg(acpi_t)
 
 allow acpid_t self:capability { kill mknod sys_admin sys_nice sys_time };
 dontaudit acpid_t self:capability { dac_override dac_read_search setuid sys_ptrace sys_tty_config };
-allow acpid_t self:process { getsession signal_perms };
+allow acpid_t self:process { getsession signal_perms getsched };
+# for pidof and pgrep
+allow acpid_t self:cap_userns sys_ptrace;
```
pebenito: We've typically dontaudited this for `capability`. While `cap_userns` is far less concerning, is it really needed?

etbe: OK I removed this and if necessary will submit it later with more information.

pebenito (approved):

PR #984, opened 2025-07-15
various: make dbus optional

0xC0ncord:
There's a few modules where `dbus_()` interfaces are used outside of an optional which causes policy build failures on systems without it. `kubernetes` and `cloud-init` are two such modules that don't actually strictly require it.

Let's go ahead and just blanket make `dbus` optional everywhere.

### DIFF
--- a/policy/modules/admin/blueman.te
+++ b/policy/modules/admin/blueman.te
@@ -7,7 +7,9 @@ policy_module(blueman)
 
 type blueman_t;
 type blueman_exec_t;
-dbus_system_domain(blueman_t, blueman_exec_t)
+domain_type(blueman_t)
+domain_entry_file(blueman_t, blueman_exec_t)
+role system_r types blueman_t;
 
 type blueman_runtime_t alias blueman_var_run_t;
 files_runtime_file(blueman_runtime_t)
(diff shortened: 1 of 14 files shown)


policy/modules/admin/blueman.te
```diff
@@ -7,7 +7,6 @@ policy_module(blueman)
 
 type blueman_t;
 type blueman_exec_t;
-dbus_system_domain(blueman_t, blueman_exec_t)
```
pebenito, line 10: You'll need to add something back here, since `blueman_t` won't be a domain anymore if the dbus module  is removed.

policy/modules/services/accountsd.te
```diff
@@ -11,7 +11,6 @@ gen_require(`
 
 type accountsd_t;
 type accountsd_exec_t;
-dbus_system_domain(accountsd_t, accountsd_exec_t)
```
pebenito, line 14: Same as previous comment.

policy/modules/admin/blueman.te
0xC0ncord: Thanks!

pebenito (approved):

PR #986, opened 2025-07-16
services

etbe:


policy/modules/services/networkmanager.te
```diff
@@ -230,6 +230,10 @@ optional_policy(`
 	chronyd_domtrans_cli(NetworkManager_t)
 ')
 
+optional_policy(`
+	chronyd_domtrans_cli(NetworkManager_t)
+')
```
pebenito: Duplicate block

policy/modules/system/authlogin.if
```diff
@@ -976,6 +978,24 @@ interface(`auth_setattr_faillog_files',`
...
+## <summary>
+##	create the login failure log directory.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`auth_create_faillog_dir',`
```
pebenito: `auth_var_lib_filetrans_faillog()`

pebenito (changes requested):

PR #999, opened 2025-07-25
strict

etbe:


policy/modules/services/dbus.if
```diff
@@ -121,17 +129,32 @@ template(`dbus_role_template',`
 	corecmd_bin_domtrans($1_dbusd_t, $3)
 	corecmd_shell_domtrans($1_dbusd_t, $3)
 
+	dev_read_sysfs($1_dbusd_t)
```
pebenito: What part(s) of sysfs is this reading?

policy/modules/system/systemd.if
```diff
@@ -274,6 +297,30 @@ template(`systemd_user_daemon_domain',`
...
+##	Domain that is entered with noatsecure
+##	</summary>
+## </param>
+#
+template(`systemd_user_daemon_domain_noatsecure',`
+	gen_require(`
+		type $1_systemd_t;
+	')
+
+	allow $1_systemd_t $2:process noatsecure;
```
pebenito: I think this perm should be default for anything that `systemd --user` is running, same as with pid1.

pebenito (changes requested):

policy/modules/services/dbus.if
etbe: I don't know. I'll remove it and do more tests.

policy/modules/system/systemd.if
etbe: Why? I've only found one case that needs it. Anyway if we are going to make this a more general thing then I think we should discuss it on the list. I think this PR should be merged as-is in this regard and we can have another PR if we want to expand it.

https://blog.siphos.be/2011/04/selinux-and-noatsecure-or-why-portage-complains-about-ld_preload-and-libsandbox-so/

For anyone who's following this discussion the above has a good description of the issue.

pebenito: > Why?

Because, as with system units, you can specify environment variables in user units.

etbe: Fair enough, but I think we still need more discussion, could we just get this PR done and then do a follow-up for the other domains?

etbe:
@pebenito What else do we need to get this merged?

policy/modules/system/systemd.if
pebenito: I'm not merging this block as-is. If you'd like to discuss this further on-list, please do.

etbe: OK I made the change requested.

pebenito (approved):

PR #1002, opened 2025-07-28
strict2

etbe:


etbe:
@pebenito I removed the line that caused that error from the PR because it depends on stuff from another PR. So this problem will happen again later. It's really the best regex for that purpose.

policy/modules/apps/xscreensaver.if
```diff
@@ -100,3 +100,22 @@ interface(`xscreensaver_run',`
...
+## <summary>
+##	Transition to xscreensaver_helper_t when running xscreensaver_helper_exec_t
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+#
+interface(`xscreensaver_run_helper', `
```
pebenito: `xscreensaver_domtrans_helper`

policy/modules/apps/xscreensaver.if
```diff
@@ -100,3 +100,22 @@ interface(`xscreensaver_run',`
...
+	gen_require(`
+		type xscreensaver_helper_exec_t, xscreensaver_helper_t;
+	')
+
+	domtrans_pattern($1, xscreensaver_helper_exec_t, xscreensaver_helper_t)
+	allow xscreensaver_helper_t $1:fd use;
```
pebenito: This should already be part of the `domtrans_pattern`

policy/modules/roles/sysadm.te
```diff
@@ -38,11 +42,22 @@ allow sysadm_t self:netlink_route_socket rw_netlink_socket_perms;
 # for debugging BPF programs
 allow sysadm_t self:capability2 bpf;
 
+allow sysadm_t self:netlink_generic_socket { create setopt bind write read };
```
pebenito: Where does this come into play?

policy/modules/roles/sysadm.te
```diff
@@ -38,11 +42,22 @@ allow sysadm_t self:netlink_route_socket rw_netlink_socket_perms;
...
+
+# for ptrace
+allow sysadm_t self:netlink_tcpdiag_socket { create write nlmsg_read read };
```
pebenito: Dupe.

policy/modules/roles/sysadm.te
```diff
@@ -38,11 +42,22 @@ allow sysadm_t self:netlink_route_socket rw_netlink_socket_perms;
...
+
+allow sysadm_t self:capability audit_write;
```
pebenito: Where does this come into play?

policy/modules/roles/sysadm.te
```diff
@@ -38,11 +42,22 @@ allow sysadm_t self:netlink_route_socket rw_netlink_socket_perms;
...
+allow sysadm_t self:system status;
```
pebenito: This doesn't seem to fit. Typically, `system` userspace perms only apply to systemd system or user sessions. There is already `allow sysadm_t sysadm_systemd_t:system { disable enable reload start status stop };` in the policy, so there may be a labeling problem on your system. It may also explain some of the other perms you've added.

policy/modules/system/systemd.te
```diff
@@ -2242,6 +2242,7 @@ tunable_policy(`systemd_tmpfiles_manage_all',`
 ')
 
 optional_policy(`
+	dbus_manage_lib_files(systemd_tmpfiles_t)
 	dbus_read_lib_files(systemd_tmpfiles_t)
```
pebenito: Please remove the read line.

policy/global_tunables
```diff
@@ -129,3 +129,11 @@ gen_tunable(user_tcp_server,false)
 ## </p>
 ## </desc>
 gen_tunable(user_udp_server,false)
+
+## <desc>
+## <p>
+## Allow users to execmod tmpfs files, KDE plasmashell needs this
```
pebenito, line 135: Do you know if there is a bug filed about this? It seems extremely broken.

pebenito (changes requested):

policy/modules/apps/xscreensaver.if
etbe: It wasn't being used so I removed it.

policy/modules/roles/sysadm.te
etbe: I'll remove it.

policy/modules/roles/sysadm.te
etbe: How?

policy/global_tunables
etbe: I don't know of a bug. Pretty much all of KDE requires execmem so is this a surprise?

policy/modules/roles/sysadm.te
pebenito: https://github.com/SELinuxProject/refpolicy/blob/e570a8b238a4f15cbf9d5f1f6d82997bfe53991c/policy/modules/system/userdomain.if#L1365

etbe: OK, well I already removed that.

etbe:
@pebenito What else do we need to get this merged?

PR #1004, opened 2025-08-04
miscnetwork

etbe:


policy/modules/services/apache.te
```diff
@@ -436,6 +436,7 @@ manage_files_pattern(httpd_t, httpd_tmpfs_t, httpd_tmpfs_t)
 manage_lnk_files_pattern(httpd_t, httpd_tmpfs_t, httpd_tmpfs_t)
 manage_fifo_files_pattern(httpd_t, httpd_tmpfs_t, httpd_tmpfs_t)
 manage_sock_files_pattern(httpd_t, httpd_tmpfs_t, httpd_tmpfs_t)
+allow httpd_t httpd_tmpfs_t:file map;
```
pebenito: Please use `mmap_manage_files_pattern` on line 435 instead.

policy/modules/services/ssh.if
```diff
@@ -950,3 +950,21 @@ interface(`ssh_dontaudit_agent_tmp',`
...
+##	Domain to permit
+##	</summary>
+## </param>
+#
+interface(`ssh_search_sshd_runtime',`
+	gen_require(`
+		type sshd_runtime_t;
+	')
+
+	allow $1 sshd_runtime_t:dir search;
```
pebenito: `search_dir_perms` please

policy/modules/services/ssh.te
```diff
@@ -17,7 +17,7 @@ gen_tunable(allow_ssh_keysign, false)
 ## Allow ssh logins as sysadm_r:sysadm_t
 ## </p>
 ## </desc>
-gen_tunable(ssh_sysadm_login, false)
+gen_tunable(ssh_sysadm_login, true)
```
pebenito: Please revert.

policy/modules/services/ntp.if
```diff
@@ -192,7 +192,7 @@ interface(`ntp_filetrans_drift',`
 	')
 
 	files_search_var_lib($1)
-	files_var_lib_filetrans($1, ntp_drift_t, dir)
+	files_var_lib_filetrans($1, ntp_drift_t, dir, "ntpsec")
```
pebenito, line 195: Which ntp client is this for? Why do we need a name transition? I'm concerned this will break other clients.

policy/modules/system/sysnetwork.te
```diff
@@ -363,6 +363,7 @@ term_dontaudit_use_all_ttys(ifconfig_t)
 term_dontaudit_use_all_ptys(ifconfig_t)
 term_dontaudit_use_ptmx(ifconfig_t)
 term_dontaudit_use_generic_ptys(ifconfig_t)
+term_use_unallocated_ttys(ifconfig_t)
```
pebenito: What scenario is this for?

pebenito (changes requested):

policy/modules/services/ntp.if
etbe: This interface is only used by dpkg_script_t, so it can't break anything other than a Debian package install.

It's for the ntpsec package in Debian which is depended on by the ntp package which is now a transitional package.  From the description:
 This is the NTPsec version of NTP.  NTPsec is a secure, hardened,
 and improved implementation derived from the original NTP project.
 NTPsec supports Network Time Security (NTS) which provides
 cryptographically authenticated time.

etbe: Also what it's used for is when the Debian package installs a directory and doesn't run restorecon on it.

policy/modules/system/sysnetwork.te
etbe: I think it's from sulogin or similar system recovery. Shall I remove it and get more information?

pebenito: If we can get more info that could be helpful. It's not particularly concerning but understanding it could be helpful since sounds like a corner case.

pebenito (approved):

PR #1007, opened 2025-08-07
systemd

etbe:


policy/modules/roles/sysadm.te
```diff
@@ -123,6 +123,10 @@ ifdef(`init_systemd',`
 	systemd_dbus_chat_networkd(sysadm_t)
 	fs_read_nsfs_files(sysadm_t)
 
+	systemd_run_nspawn(sysadm_t, sysadm_r)
```
pebenito: Why would sysadm do this instead of having systemd start it?

etbe: Sometimes it's handy to run systemd-nspawn in the same manner that chroot has traditionally been run.

etbe:
@pebenito Does this need anything else?

policy/modules/roles/sysadm.te
pebenito: This sounds like a local customization.

policy/modules/system/systemd.if
```diff
@@ -2955,6 +3016,83 @@ interface(`systemd_status_all_user_sessions',`
...
+##    </summary>
+## </param>
+#
+interface(`systemd_user_sessions_manage_sock_file',`
+	gen_require(`
+		attribute systemd_user_session_type;
+	')
+
+	allow systemd_user_session_type $1:sock_file manage_sock_file_perms;
+')
```
pebenito, line 3071: These need to be changed into abstract socket and named socket activation, like is done with the pid 1 equivalents. Then the `sock_file` can go with the named socket activation interface.

policy/modules/services/ssh.te
```diff
@@ -192,6 +192,11 @@ tunable_policy(`allow_ssh_keysign',`
 	domtrans_pattern(ssh_t, ssh_keysign_exec_t, ssh_keysign_t)
 ')
 
+ifdef(`init_systemd',`
+	systemd_user_runtime_dir_unlink(ssh_agent_tmp_t, sock_file)
```
pebenito, line 196: This should probably allow `systemd --user` to clean up any user temp file instead.

pebenito (changes requested):
cc @0xC0ncord for thoughts on `systemd --user` changes.

policy/modules/apps/gpg.te
```diff
@@ -182,6 +182,10 @@ optional_policy(`
 	spamassassin_read_spamd_tmp_files(gpg_t)
 ')
 
+optional_policy(`
+	systemd_user_sessions_create_stream_socket(gpg_t)
+')
```
0xC0ncord, line 187: Shouldn't this instead go into `ifdef(init_systemd)`?

policy/modules/apps/gpg.te
```diff
@@ -304,6 +308,10 @@ optional_policy(`
 	pcscd_stream_connect(gpg_agent_t)
 ')
 
+optional_policy(`
+	systemd_user_sessions_create_sock_file(gpg_runtime_t)
+')
```
0xC0ncord, line 313: Ditto.

policy/modules/roles/sysadm.te
0xC0ncord: I see the usefulness of this so maybe it should be relegated to a tunable.

policy/modules/system/systemd.if
0xC0ncord: Doing this would definitely help readability of the policy too.

0xC0ncord (commented):
Just a couple comments. The rest of the `systemd --user` changes look good.

policy/modules/roles/sysadm.te
etbe: OK I added the tunable sysadm_run_nspawn

policy/modules/apps/gpg.te
pebenito: This is my preferred implementation. My preference is to only use `init_systemd` when a set of rules can't be reasonably done by an optional that depends on the systemd module.

pebenito (changes requested):

policy/modules/services/ssh.te
etbe: How would you suggest doing that? Have userdom_user_tmp_file() add an attribute to it named user_tempfile and then have an interface to allow unlinking user_tempfile attribute?

policy/modules/system/systemd.if
etbe: What do you mean?  init_t doesn't seem to be able to do the same sorts of things.

# sesearch -A -s init_t -c sock_file -p create
allow files_unconfined_type file_type:sock_file { append create execmod execute getattr ioctl link lock map mounton open quotaon read relabelfrom relabelto rename setattr unlink watch watch_mount watch_reads watch_sb watch_with_perm write };
allow filesystem_unconfined_type filesystem_type:sock_file { append create execmod execute getattr ioctl link lock map mounton open quotaon read relabelfrom relabelto rename setattr unlink watch watch_mount watch_reads watch_sb watch_with_perm write };
allow init_t init_mountpoint_type:sock_file { create open }; [ init_create_mountpoints ]:True
allow kern_unconfined unlabeled_t:sock_file { append create execmod execute getattr ioctl link lock map mounton open quotaon read relabelfrom relabelto rename setattr unlink watch watch_mount watch_reads watch_sb watch_with_perm write };
# sesearch -A -s init_t -c unix_stream_socket -p create
allow unconfined_domain_type domain:unix_stream_socket { accept append bind connect connectto create getattr getopt ioctl listen lock map name_bind read recvfrom relabelfrom relabelto sendto setattr setopt shutdown write };
root@xev:/home#

pebenito: https://github.com/SELinuxProject/refpolicy/blob/5a5814e0484b5066e1f14750bd4c62009e993b18/policy/modules/system/init.if#L385

policy/modules/services/ssh.te
pebenito: Yes, that sounds good.

policy/modules/roles/sysadm.te
```diff
@@ -24,6 +24,13 @@ gen_tunable(allow_ptrace, false)
 ## </desc>
 gen_tunable(sysadm_allow_rw_inherited_fifo, false)
 
+## <desc>
+## <p>
+## Allow sysadm to run systemd-nspawn directly
+## </p>
+## </desc>
+gen_tunable(sysadm_run_nspawn, false)
```
thesamesam, line 32: Yes please :)

I was just looking at this too.

PR #1015, opened 2025-08-31
Let the kernel load firmware files during boot.

gtrentalancia:
**Let the kernel load firmware files during boot.**

Without this patch set the latest kernel is not able to initialize devices that require one or more firmware files to be loaded (tested with kernel version 6.16.4).

### DIFF
--- a/policy/flask/access_vectors
+++ b/policy/flask/access_vectors
@@ -411,6 +411,10 @@ class system
 	module_request
 	module_load
 	firmware_load
+	kexec_image_load
+	kexec_initramfs_load
+	policy_load
+	x509_certificate_load
 
 	# these are overloaded userspace
 	# permissions from systemd


gtrentalancia:
For further reference, please see the [corresponding kernel change](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2c2b1e059792f610bae4fee8ed517b8ce9c585fb)

policy/modules/kernel/kernel.te
```diff
@@ -443,6 +443,7 @@ optional_policy(`
 ')
 
 optional_policy(`
+	libs_load_firmware(kernel_t)
```
cgzones, line 451: also `allow kernel_t self:system firmware_load;` might be needed.

policy/flask/access_vectors
```diff
@@ -410,6 +410,7 @@ class system
 	syslog_console
 	module_request
 	module_load
+	firmware_load
```
cgzones, line 413: Why not add the whole list of newly added permissions?
```
kexec_image_load
kexec_initramfs_load
policy_load
x509_certificate_load
```

policy/modules/kernel/kernel.te
gtrentalancia: I am aware of such failure, however I don't know which firmware is causing it...

The system appears to be working fine without granting such permission and even more importantly firmware files are normally labelled as _lib_t_ rather than _kernel_t_.

policy/flask/access_vectors
gtrentalancia: The pull request is specific about being able to load firmware files, the other permissions can be added later on...

policy/modules/kernel/kernel.te
cgzones: Hmm, it might have to be only needed for Linux 6.15; I removed this permission recently using 6.16.

gtrentalancia: I am using kernel version 6.16.4 and the _libs_load_firmware(kernel_t)_ permission is absolutely needed in order to properly load the firmware for all devices that rely on it.

The kernel (running as _kernel_t_) needs to load firmware files from _/lib/firmware_ which are binary files labelled as _lib_t_.

gtrentalancia: Please see the following additional change recently introduced: https://github.com/SELinuxProject/refpolicy/pull/1015/commits/b7827465c0134bd2af58dd51e9a5c55ee9ce3194

policy/flask/access_vectors
pebenito: Since the related kernel patch has the other permissions, please add them in this PR.

policy/modules/kernel/kernel.te
```diff
@@ -1,5 +1,9 @@
 policy_module(kernel)
 
+gen_require(`
+	class system firmware_load;
```
pebenito, line 4: This shouldn't be necessary since `system` is a kernel object class, thus is automatically required.

policy/modules/system/libraries.if
```diff
@@ -1,5 +1,24 @@
...
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`libs_load_firmware',`
+	gen_require(`
+		type lib_t;
+		class system firmware_load;
```
pebenito, line 16: Also not necessary.

gtrentalancia: It generates a SELint warning otherwise: _(W): No explicit declaration for userspace class system.  You should access it via interface call or use a require block. (W-001)_

policy/modules/kernel/kernel.te
gtrentalancia: It would generate a SELint warning similar to https://github.com/SELinuxProject/refpolicy/pull/1015#discussion_r2316414123 if not explicitly required.

policy/flask/access_vectors
gtrentalancia: Here you go: https://github.com/SELinuxProject/refpolicy/pull/1015/commits/7a0779dee785f905d84814004b103059257ceb74

pebenito (approved):

policy/modules/kernel/kernel.te
cgzones: >> Hmm, it might have to be only needed for Linux 6.15; I removed this permission recently using 6.16.
>
> I am using kernel version 6.16.4 and the libs_load_firmware(kernel_t) permission is absolutely needed in order to properly load the firmware for all devices that rely on it.
>
> The kernel (running as kernel_t) needs to load firmware files from /lib/firmware which are binary files labelled as lib_t.

That comment was about `allow kernel_t self:system firmware_load;`, i think 6.15 did not correctly always took the target label from the firmware file.
I agree that `libs_load_firmware(kernel_t)` is necessary.

gtrentalancia: It's not something related to kernel version 6.15, as it's still there in **6.16 up to at least sublevel 4**.

In my opinion it's a SELinux kernel bug, **but not the so-called "file descriptor leak" bug**...

Anyway, as far as the policy is concerned, only _libs_load_firmware(kernel_t)_ appears to be needed for the system to boot up properly and that is now merged in.

PR #1027, opened 2025-09-23
evolution

etbe:


policy/modules/apps/evolution.te
```diff
@@ -346,6 +368,10 @@ optional_policy(`
 	gnome_stream_connect_gconf(evolution_alarm_t)
 ')
 
+optional_policy(`
+	wm_send_fd(evolution_alarm_t)
```
pebenito, line 370: Do you know what this fd is used for?

Also, with the added `wm` call on line 335 (new line num), the `wm` module is already unconditionally required. Please make both lines uncondtional or optional.

pebenito (changes requested):

policy/modules/apps/evolution.te
etbe: I made them all optional. Not sure about the use of  the fd.

pebenito: Is there an error message in the regular log (not audit messages)? If not, how does the code functionally fail?

pebenito:
Some, if not all of this merged separately. Please open a new PR if there are still relevant pieces.

PR #1028, opened 2025-09-23
wireshark

etbe:


pebenito (changes requested):
I'd appreciate if you would put further details into the PR titles and description so the Changelog will have useful information.


https://github.com/SELinuxProject/refpolicy/blob/004ca3252282b52f525c24f3c874bf7ecf724be1/Changelog#L118

PR #1037, opened 2025-11-01
sway: new policy module

WavyEbuilder:
Very WIP atm. PR created early such that others can give feedback as it progresses given that Wayland is still very novel in reference policy.

@pebenito Couple questions while I'm working on this:
- I'm considering going the `$1_sway_t` approach - do you think this is useful? I noticed that we don't rely solely on UBAC for protection (e.g. with `$1_systemd_t` and hence I think given the scope of the compositor this might be desirable.
- I did consider giving the `wayland_compositor` typeattribute manage files perms on `wayland_compositor_tmpfs_type`, but decided against this so that different running compositors can't "contaminate" each others tmpfs's - I don't think this is a huge deal, but it adds a bit of repetition for consumers of the Wayland module (each compositor needs to give itself manage file perms + map on its tmpfs). Would you like me to change this? I personally think having them be separate is the best approach, but of course open to suggestions/ideas.

Anyone else please also free to leave reviews and/or comment, very much open to opinions for this.

cc @aerusso @0xC0ncord

Thanks!

### DIFF
--- a/policy/modules/roles/staff.te
+++ b/policy/modules/roles/staff.te
@@ -57,6 +57,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	sway_role(staff, staff_t, staff_r)
+')
+
+optional_policy(`
 	sysadm_role_change(staff_r, staff_t)
 	userdom_dontaudit_use_user_terminals(staff_t)
 ')
--- a/policy/modules/roles/sysadm.te
+++ b/policy/modules/roles/sysadm.te
@@ -1107,6 +1107,10 @@ optional_policy(`
 ')
 
 optional_policy(`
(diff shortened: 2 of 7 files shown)


pebenito:
Is this still relevant? It's very stale and I'd like to close it, if not.

riseupcoder:
> Is this still relevant? It's very stale and I'd like to close it, if not.

Yes, I still think this is relevant.

I recently ported the policy to Fedora, made a few minor changes, and tested it. It worked fine, no issues.

I haven't tested it on Gentoo, but I'd still like to see the Sway policy merged.

pebenito:
@WavyEbuilder will you be making any further changes? Based on @riseupcoder 's note above, it sounds like this is useable as-is. If we merge it now, others can contribute.

PR #1042, opened 2025-11-04
gstreamer2

etbe:


WavyEbuilder (commented):
Minor nit, but you might want to amend the commit to be `~/.cache` and not `~/..cache` then force push

PR #1043, opened 2025-11-08
chromium: drop the chromium_render_t domain

WavyEbuilder:
Upstream no longer supports the SELinux sandbox[1], so no dyntrans is attempted, making this domain redundent. It's been long enough now that it is reasonable to drop this domain.

[1] https://github.com/chromium/chromium/blob/0f8100c82dc431809d866e5d0218cbe3c482b20c/sandbox/linux/README.md?plain=1#L31

### DIFF
--- a/policy/modules/apps/chromium.if
+++ b/policy/modules/apps/chromium.if
@@ -29,14 +29,12 @@
 template(`chromium_role',`
 	gen_require(`
 		type chromium_t;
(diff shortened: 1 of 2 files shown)


WavyEbuilder:
I should note that this shouldn't really call breakage: the policy doesn't seem to work as is on either X11 or Wayland boxes running the latest version of chrome for me. More patches to come.

PR #1045, opened 2025-11-14
firewalld: Allow firewall-cmd to be called from systemd

dsugar100:
RedHat has added 'ExecStartPost=/usr/bin/firewall-cmd --state' in the firewalld.service file. As firewall-cmd was just labeled bin_t this ran in initrc_t.
This change sets the type for firewall-cmd so running from systemd will work.  Also added runas in the admin interface to allow admin users to still be able to run firewall-cmd.

node=localhost type=USER_AVC msg=audit(1763093230.354:10021): pid=1156 uid=81 auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t:s0 msg='avc:  denied  { send_msg } for  scontext=system_u:system_r:initrc_t:s0 tcontext=system_u:system_r:firewalld_t:s0 tclass=dbus permissive=1 exe="/usr/bin/dbus-broker" sauid=81 hostname=? addr=? terminal=?' UID="dbus" AUID="unset" SAUID="dbus"

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -1316,6 +1316,10 @@ optional_policy(`
 	init_dbus_chat(initrc_t)
 
 	optional_policy(`
+		firewalld_dbus_chat(initrc_t)
+	')
+
+	optional_policy(`
 		networkmanager_dbus_chat(initrc_t)
 	')
 


dsugar100:
Let me know if you want more here.   I did just a basic set of changes to get this working and not affect anything else (hopefully).  I can add a domtrans interface if you want, etc...

policy/modules/services/firewalld.te
```diff
@@ -123,3 +127,21 @@ optional_policy(`
...
+corecmd_exec_bin(firewalld_cli_t)
+
+files_read_etc_files(firewalld_cli_t)
+
+miscfiles_read_generic_certs(firewalld_cli_t)
+miscfiles_read_localization(firewalld_cli_t)
+
+dbus_system_bus_client(firewalld_cli_t)
+
+firewalld_dbus_chat(firewalld_cli_t)
```
pebenito: It looks like this is the distinctive perm. If that's the case, I'd rather add it to `initrc_t` directly.

dsugar100: I think you are saying - don't label firewall-cmd but have 'firewalld_dbus_chat(initrc_t)' - in init.te?

pebenito: Correct.

dsugar100: Thanks for the clarification - and I see where similar things were done.   Hopefully this is in the correct place.

PR #1052, opened 2025-12-28
matrixd: gatekeep postgresql calls in an optional policy block

WavyEbuilder:
postgresql is a seperate policy module[1], and hence is not required for the matrixd policy to work; it's already behind a tunable policy block, and hence the boolean matrix_postgresql_connect.

[1] https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/services/postgresql.te

Closes: https://bugs.gentoo.org/968058

### DIFF
--- a/policy/modules/services/matrixd.te
+++ b/policy/modules/services/matrixd.te
@@ -123,11 +123,6 @@ tunable_policy(`matrix_allow_federation',`
 	corenet_dontaudit_udp_bind_all_ports(matrixd_t)
 ')
 
-tunable_policy(`matrix_postgresql_connect',`
-	postgresql_stream_connect(matrixd_t)
-	postgresql_tcp_connect(matrixd_t)
-')
-
 tunable_policy(`matrix_bind_all_unreserved_tcp_ports',`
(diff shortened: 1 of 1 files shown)


policy/modules/services/matrixd.te
```diff
@@ -123,9 +123,11 @@ tunable_policy(`matrix_allow_federation',`
...
 
-tunable_policy(`matrix_postgresql_connect',`
-	postgresql_stream_connect(matrixd_t)
-	postgresql_tcp_connect(matrixd_t)
+optional_policy(`
+	tunable_policy(`matrix_postgresql_connect',`
+		postgresql_stream_connect(matrixd_t)
+		postgresql_tcp_connect(matrixd_t)
+	')
 ')
```
pebenito: Please move the block to the end, as that's where optional block go.

pebenito (changes requested):

policy/modules/services/matrixd.te
WavyEbuilder: Sorry, I forgot to do this; I thought I'd already done it. Fixed!

pebenito (approved):

PR #1055, opened 2026-01-11
Add specialized Networkmanager dispatch domains

aerusso:
Currently, NetworkManager dispatch scripts are tagged as `NetworkManager_initrc_exec_t`, and run in the `NetworkManager_t` domain itself.  However, those scripts may require permissions different than those of `NetworkManager_t`.

This patch introduces an interface to generate domains for these specialized scripts, and grants permission for NetworkManager to transition into them, as well as communicate with it.

This patch also removes some old, incompatible tagging that used `bin_t` in favor of `NetworkManager_initrc_exec_t`.  Alternatively, we could stop using (ab)using `NetworkManager_initrc_exec_t`, and just use `bin_t` eveywhere (the more I think about it, the more I prefer this, but I don't understand the choice, so maybe I'm missing something).

The last commit actually does something useful and creates a domain for the postfix NetworkManager dispatcher script so that it can properly propagate changes in `/etc/resolv.conf` to its chroot.


policy/modules/services/networkmanager.if
```diff
@@ -430,3 +431,34 @@ interface(`networkmanager_admin',`
...
+##	Create a set of derived types for
+##	networkmanager dispatcher scripts.
+## </summary>
+## <param name="prefix">
+##	<summary>
+##	The prefix to be used for deriving type names.
+##	</summary>
+## </param>
+#
+interface(`networkmanager_dispatch_script',`
```
pebenito: If you change the interface to a template, that should resolve the lint issue.

policy/modules/services/networkmanager.fc
```diff
@@ -15,6 +15,7 @@
 /etc/wicd/wireless-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 /etc/wicd/wired-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 
+/usr/lib/NetworkManager/dispatcher\.d(/.*)?	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
```
pebenito: I think we should stick with the original `bin_t` for generic dispatching. We should keep `NetworkManager_initrc_exec_t` for init/start scripts. Arguably the below nm-dispatcher fcs should probably be `NetworkManager_exec_t` instead.

policy/modules/services/networkmanager.if
```diff
@@ -430,3 +431,34 @@ interface(`networkmanager_admin',`
...
+	domain_type(NetworkManager_dispatch_$1_t)
+	corecmd_shell_entry_type(NetworkManager_dispatch_$1_t)
+	role system_r types NetworkManager_dispatch_$1_t;
+
+	type NetworkManager_dispatch_$1_exec_t, NetworkManager_dispatch_exec_type;
+	domain_entry_file(NetworkManager_dispatch_$1_t, NetworkManager_dispatch_$1_exec_t)
+
+	allow NetworkManager_dispatch_$1_t NetworkManager_dispatch_$1_exec_t:file entrypoint;
+	domtrans_pattern(NetworkManager_t, NetworkManager_dispatch_$1_exec_t, NetworkManager_dispatch_$1_t)
+')
```
pebenito: Since this is a template, it should go at the top of the file.

policy/modules/services/networkmanager.te
```diff
@@ -94,6 +98,11 @@ files_runtime_filetrans(NetworkManager_t, NetworkManager_runtime_t, { dir file s
 
 can_exec(NetworkManager_t, { NetworkManager_exec_t NetworkManager_initrc_exec_t wpa_cli_exec_t NetworkManager_tmp_t })
 
+allow NetworkManager_t NetworkManager_dispatch_domains:fifo_file { rw_fifo_file_perms };
+init_rw_inherited_stream_socket(NetworkManager_dispatch_domains)
+init_use_inherited_script_ptys(NetworkManager_dispatch_domains)
+kernel_read_system_state(NetworkManager_dispatch_domains)
```
pebenito: This should go in a separate block later in the file, i.e. a block for all rules that have `NetworkManager_dispatch_domains` as the subject/sources.

policy/modules/services/networkmanager.te
```diff
@@ -203,6 +212,8 @@ userdom_write_user_tmp_sockets(NetworkManager_t)
 userdom_dontaudit_use_unpriv_user_fds(NetworkManager_t)
 userdom_dontaudit_use_user_ttys(NetworkManager_t)
 
+exec_files_pattern(NetworkManager_t, NetworkManager_dispatch_exec_type, NetworkManager_dispatch_exec_type)
```
pebenito: This shouldn't be necessary since you already have a domtrans for each of the domains/entrypoints.

pebenito (changes requested):

policy/modules/services/networkmanager.fc
aerusso: I used `NetworkManager_exec_t` here, since it's going to run in `NetworkManager_t` domain and it's conceivable that an administrator would want to manually run one of these scripts manually.

aerusso:
Another thing that fell off my radar, sorry.

policy/modules/services/networkmanager.if
```diff
@@ -1,5 +1,36 @@
...
+	gen_require(`
+		attribute NetworkManager_dispatch_domains;
+		attribute NetworkManager_dispatch_exec_type;
+
+		type NetworkManager_t;
+	')
+
+	type NetworkManager_dispatch_$1_t, NetworkManager_dispatch_domains;
+	domain_type(NetworkManager_dispatch_$1_t)
+	corecmd_shell_entry_type(NetworkManager_dispatch_$1_t)
```
pebenito, line 24: This doesn't seem right. The transition should hit when NM does the exec on the script. The entrypoint would be the script too. The shell doesn't come in until after the transition.

policy/modules/services/networkmanager.fc
```diff
@@ -15,8 +15,9 @@
 /etc/wicd/wireless-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 /etc/wicd/wired-settings\.conf	--	gen_context(system_u:object_r:NetworkManager_etc_rw_t, s0)
 
-/usr/lib/NetworkManager/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
-/usr/lib/networkmanager/nm-dispatcher.*	--	gen_context(system_u:object_r:NetworkManager_initrc_exec_t,s0)
+/usr/lib/NetworkManager/dispatcher\.d(/.*)?	--	gen_context(system_u:object_r:NetworkManager_exec_t,s0)
```
pebenito, line 18: I think the default for these should be `bin_t`.

PR #1057, opened 2026-01-13
snmpd doesn't seem to need sys_ptrace capability

dsugar100:
Tested on RHEL9 with net-snmp 5.9.4.
I was able to query information about running processes.  I'm not sure what else sys_ptrace might be needed for by snmpd

### DIFF
--- a/policy/modules/services/snmp.te
+++ b/policy/modules/services/snmp.te
@@ -29,9 +29,8 @@ files_type(snmpd_var_lib_t)
 # Local policy
 #
 
-allow snmpd_t self:capability { chown dac_override ipc_lock kill net_admin setgid setuid sys_nice sys_ptrace sys_tty_config };
+allow snmpd_t self:capability { chown dac_override ipc_lock kill net_admin setgid setuid sys_nice sys_tty_config };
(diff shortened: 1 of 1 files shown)


pebenito:
@0xC0ncord looks like you added the `cap_userns` `sys_ptrace`; do you have more info on the `sys_ptrace`? https://github.com/SELinuxProject/refpolicy/commit/cec7f0d3e2cfaa6165451feade1daf77f17e360c

@dsugar100 looks like you have a cert problem for your commit signature.

0xC0ncord:
> @0xC0ncord you added the `cap_userns` `sys_ptrace`; do you have more info on the `sys_ptrace`? [cec7f0d](https://github.com/SELinuxProject/refpolicy/commit/cec7f0d3e2cfaa6165451feade1daf77f17e360c)

Hmm. This change was added along with a bunch of other `cap_userns` related permissions in the same PR that originally added the container policy. I seem to remember that I just copied every `cap_userns` related access from `container-selinux` at the time in an effort to get as many things working as possible.

So, it seems we should be able to drop the access for `cap_userns` here as well.

dsugar100:
Ok, then I will drop the cap_usrns sys_ptrace as well.

PR #1062, opened 2026-01-20
Various fixes

0xC0ncord:
- Add interfaces to init for automatic creation of unit directories. Some systemd units use these to rely on the init system (mostly systemd) to create the service's needed directories.
- Fixes for tuned to get it running on modern systems and modern versions.
- Allow named socket activation for `systemd-resolved`. This is mostly to fix an issue with init wanting to change the mode on its unix socket. If this fails, `systemd-resolved` will fail to start.
- Various container and kubernetes fixes.

### DIFF
--- a/policy/modules/services/crio.te
+++ b/policy/modules/services/crio.te
@@ -36,8 +36,13 @@ files_mounton_etc_dirs(crio_t)
 # watch /usr/share/containers/oci/hooks.d
 files_watch_usr_dirs(crio_t)
 
+fs_create_bpf_dirs(crio_t)
+fs_manage_bpf_files(crio_t)
+
 kernel_dgram_send(crio_t)
 kernel_read_irq_sysctls(crio_t)
+kernel_read_vm_overcommit_sysctl(crio_t)
+kernel_request_load_module(crio_t)
 
 auth_use_nsswitch(crio_t)
 
@@ -49,6 +54,8 @@ miscfiles_mounton_generic_cert_dirs(crio_t)
 container_read_home_config(crio_t)
 
 container_watch_config_dirs(crio_t)
+
+crio_read_conmon_state(crio_t)
 
 # Ensure conmon runs in s0 so that it can talk to the container
 podman_spec_rangetrans_conmon(crio_t, s0)


policy/modules/services/crio.te
```diff
@@ -64,6 +69,8 @@ kubernetes_relabelfrom_tmpfs_dirs(crio_t)
 kubernetes_relabelfrom_tmpfs_files(crio_t)
 kubernetes_relabelfrom_tmpfs_symlinks(crio_t)
 
+crio_read_conmon_state(crio_t)
```
pebenito: Should be after the `container_*` calls above

policy/modules/services/kanidm.te
```diff
@@ -97,6 +97,12 @@ miscfiles_read_generic_certs(kanidmd_t)
 
 userdom_use_inherited_user_terminals(kanidmd_t)
 
+ifdef(`init_systemd',`
+	init_create_unit_cache_dir(kanidm_unixd_cache_t, "kanidmd")
+	init_create_unit_var_lib_dir(kanidm_unixd_cache_t, "kanidmd")
+	init_manage_unit_runtime_dir(kanidm_unixd_cache_t, "kanidmd")
+')
```
pebenito, line 103: It doesn't seem necessary to have the name (same as below hunk). Is there a type transition conflict I'm missing?

policy/modules/system/init.te
```diff
@@ -618,6 +618,9 @@ ifdef(`init_systemd',`
 
 	optional_policy(`
 		container_remount_fs(init_t)
+
+		container_run_system_engine_bpf(init_t)
+		fs_manage_bpf_files(init_t)
```
pebenito: Please put the empty line between the `container_*` and `fs_*` calls.

policy/modules/services/kanidm.te
pebenito: I think it would also be good to have interfaces like `systemd_RuntimeDirectory` and `systemd_StateDirectory`, even if they simply call the above interfaces.

policy/modules/system/init.if
```diff
@@ -421,6 +421,138 @@ interface(`init_named_socket_activation',`
...
+## </param>
+#
+interface(`init_create_unit_cache_dir',`
+	ifdef(`init_systemd',`
+		gen_require(`
+			type init_t;
+		')
+
+		allow init_t $1:dir { create_dir_perms setattr };
+		files_var_filetrans(init_t, $1, dir, $2)
```
pebenito: Hmm, it seems like these are bugs in systemd that it doesn't do the file context lookup -- not that it blocks this PR. We should probably file a bug or PR to fix it.

policy/modules/services/kanidm.te
0xC0ncord: > It doesn't seem necessary to have the name (same as below hunk). Is there a type transition conflict I'm missing?

Whoops. There's some typos here. These are supposed to handle the var_lib and runtime types in the appropriate directories.

> I think it would also be good to have interfaces like `systemd_RuntimeDirectory` and `systemd_StateDirectory`, even if they simply call the above interfaces.

Great idea.

policy/modules/system/init.if
0xC0ncord: Let me double-check and make sure that this is still the case. If so, I'll add a comment here to clarify.

If that is the case then it's worth noting this would also cause issues with systemd's format specifiers like `%i` which can be used in `StateDirectory=` and the like.

0xC0ncord: OK, without the filetrans access the type still gets set correctly, so I don't think systemd is at fault here. What probably happened in my testing was I forgot to load the new filecons before the directory was created.

In any case, this lets us condense these interfaces and use the `systemd_*` ones for convenience.

PR #1063, opened 2026-01-20
Add a policy for installkernel and dracut

0xC0ncord:
Pull in the `dracut` policy module from Gentoo and get it working with modern systems. For this, we:
1. Add the `dracut` policy module as-is from Gentoo.
2. Also add a policy module for `installkernel`. This tool is used on several distros (including Gentoo) as a helper for calling the necessary utilities for installing a kernel. In our case, this tool can be used to call into `dracut`.
3. Apply a bunch of a fixes to the `dracut` module to get it working as needed.

The workflow I tested is installing a hostonly kernel via the kernel's `make install`, which calls `installkernel` and then `dracut`. Happy to share the `dracut` config if needed.

Normally `dracut` has helper scripts that get installed by other system tools (e.g. ZFS) that allow it to handle installing those components as needed. I only tested the tools that I am using personally, so there may be others that need looking into as well.

### DIFF
--- a/policy/modules/admin/dracut.te
+++ b/policy/modules/admin/dracut.te
@@ -1,11 +1,19 @@
 policy_module(dracut)
 
+########################################
+#
+# Declarations
+#
+
 type dracut_t;
 type dracut_exec_t;
 application_domain(dracut_t, dracut_exec_t)
 
 type dracut_var_log_t;
 logging_log_file(dracut_var_log_t)
+
+type dracut_runtime_t;
+files_runtime_file(dracut_runtime_t)
 
 type dracut_tmp_t;
 files_tmp_file(dracut_tmp_t)
@@ -15,13 +23,21 @@ files_tmp_file(dracut_tmp_t)
 # Local policy
 #
 
-allow dracut_t self:process setfscreate;
-allow dracut_t self:capability dac_override;
+allow dracut_t self:process { getsched setfscreate };
+# sys_admin needed for fsfreeze use
+# sys_chroot needed for ldconfig
+allow dracut_t self:capability { dac_override dac_read_search setfcap sys_admin sys_chroot };
+dontaudit dracut_t self:capability sys_resource;
 allow dracut_t self:fifo_file rw_fifo_file_perms;
 allow dracut_t self:unix_stream_socket create_stream_socket_perms;
+allow dracut_t self:alg_socket create_stream_socket_perms;
 
-manage_files_pattern(dracut_t, dracut_tmp_t, dracut_tmp_t)
-manage_dirs_pattern(dracut_t, dracut_tmp_t, dracut_tmp_t)
+allow dracut_t dracut_runtime_t:dir manage_dir_perms;
+allow dracut_t dracut_runtime_t:file manage_file_perms;
+files_runtime_filetrans(dracut_t, dracut_runtime_t, { dir file })
+
+allow dracut_t dracut_tmp_t:dir { manage_dir_perms relabelfrom };
+allow dracut_t dracut_tmp_t:file { manage_file_perms relabel_file_perms };
 manage_lnk_files_pattern(dracut_t, dracut_tmp_t, dracut_tmp_t)
 manage_chr_files_pattern(dracut_t, dracut_tmp_t, dracut_tmp_t)
 files_tmp_filetrans(dracut_t, dracut_tmp_t, dir)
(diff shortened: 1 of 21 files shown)


policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
...
+manage_chr_files_pattern(dracut_t, dracut_tmp_t, dracut_tmp_t)
+files_tmp_filetrans(dracut_t, dracut_tmp_t, dir)
+
+manage_files_pattern(dracut_t, dracut_var_log_t, dracut_var_log_t)
+logging_log_filetrans(dracut_t, dracut_var_log_t, file)
+
+corecmd_exec_bin(dracut_t)
+corecmd_exec_shell(dracut_t)
+corecmd_mmap_all_executables(dracut_t)
+corecmd_delete_all_executables(dracut_t)
```
pebenito, line 51: What's happening here (the delete in particular)?

pebenito: Related: there's also deletions in the `files_*` block and others.

policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
...
+udev_relabelto_rules_files(dracut_t)
+
+userdom_search_user_home_dirs(dracut_t)
+userdom_use_user_terminals(dracut_t)
+
+xdg_dontaudit_search_data_dirs(dracut_t)
+
+ssh_exec_keygen(dracut_t)
+ssh_read_server_keys(dracut_t)
+ssh_dontaudit_getattr_home_dirs(dracut_t)
```
pebenito: This should probably be optional

policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
...
+libs_setattr_lib_files(dracut_t)
+libs_relabelto_lib_files(dracut_t)
+
+logging_read_syslog_config(dracut_t)
+logging_delete_syslog_config_files(dracut_t)
+logging_setattr_syslog_config_files(dracut_t)
+logging_relabelto_syslog_config_files(dracut_t)
+logging_dontaudit_search_runtime_dirs(dracut_t)
+
+lvm_exec(dracut_t)
```
pebenito: Please move to the below lvm optional

policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
...
+udev_list_rules(dracut_t)
+udev_rw_rules_files(dracut_t)
+udev_delete_rules_files(dracut_t)
+udev_setattr_rules_files(dracut_t)
+udev_relabelto_rules_files(dracut_t)
+
+userdom_search_user_home_dirs(dracut_t)
+userdom_use_user_terminals(dracut_t)
+
+xdg_dontaudit_search_data_dirs(dracut_t)
```
pebenito: Optional

policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
...
+	init_manage_all_unit_files(dracut_t)
+	init_relabel_all_unit_files(dracut_t)
+
+	systemd_exec_sysusers(dracut_t)
+	systemd_read_user_units_files(dracut_t)
+')
+
+ifdef(`distro_gentoo',`
+	portage_exec_gcc_config(dracut_t)
+')
```
pebenito: Can this move to with the below `portage` optional block?

policy/modules/admin/dracut.te
```diff
@@ -0,0 +1,202 @@
+policy_module(dracut)
+
```
pebenito, line 2: Please add the "Declarations" header

policy/modules/roles/sysadm.te
```diff
@@ -555,6 +559,10 @@ optional_policy(`
 	ipsec_admin(sysadm_t, sysadm_r)
 ')
 
+optional_policy(`
+	installkernel_run(sysadm_t, sysadm_r)
+')
```
pebenito, line 564: Looks like the optionals are a bit out of order already; please move this block up after the inn one.

policy/modules/services/ssh.if
```diff
@@ -758,6 +758,25 @@ interface(`ssh_agent_exec',`
...
 ')
 
+########################################
+## <summary>
+##	Do not audit attempts to get the attributes
+##	of ssh home directory (~/.ssh).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
```
pebenito: Copy-paste error

policy/modules/system/libraries.if
```diff
@@ -149,6 +149,42 @@ interface(`libs_exec_ld_so',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`libs_delete_ld_so',`
+	gen_require(`
+		type ld_so_t;
+	')
+
+	allow $1 ld_so_t:file delete_file_perms;
```
pebenito: This should use a pattern macro so it gets the `lib_t` dir access.

policy/modules/system/libraries.if
```diff
@@ -313,6 +349,42 @@ interface(`libs_watch_lib_dirs',`
 	allow $1 lib_t:dir watch;
 ')
 
+########################################
+## <summary>
+##	Delete library files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to not audit.
```
pebenito: copy-paste error

policy/modules/system/logging.if
```diff
@@ -827,6 +827,81 @@ interface(`logging_read_syslog_config',`
...
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`logging_delete_syslog_config_files',`
+	gen_require(`
+		type syslog_conf_t;
+	')
+
+	allow $1 syslog_conf_t:file delete_file_perms;
```
pebenito: Please add parent dir access to `remove_name`

policy/modules/system/logging.if
```diff
@@ -827,6 +827,81 @@ interface(`logging_read_syslog_config',`
...
+	allow $1 syslog_conf_t:file relabelto;
+')
+
+########################################
+## <summary>
+##	Watch syslog runtime dirs.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
```
pebenito: copy-paste error

policy/modules/system/sysnetwork.if
```diff
@@ -493,6 +493,25 @@ interface(`sysnet_create_config',`
...
+##	</summary>
+## </param>
+#
+interface(`sysnet_delete_config',`
+	gen_require(`
+		type net_conf_t;
+	')
+
+	files_search_etc($1)
+	allow $1 net_conf_t:file delete_file_perms;
```
pebenito: Please add parent dir access

policy/modules/admin/portage.if
```diff
@@ -249,6 +249,25 @@ interface(`portage_run_fetch',`
 	roleattribute $2 portage_fetch_roles;
 ')
 
+########################################
+## <summary>
+##	Execute gcc-config in the caller domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
```
pebenito: copy-paste error

pebenito (changes requested):

policy/modules/admin/dracut.te
0xC0ncord: What's happening here is `installkernel` creates the staging area for the `initrd` in `/tmp`. `dracut` copies all the binaries, shared libs, config files, etc. for all the components needed to boot the system to its own directory in the staging area. `dracut` also relabels all the files to the correct type so that the SELinux contexts inside of the `initrd` are correct. After the `initrd` is created, `dracut` recursively deletes this temporary directory and all its contents.

The delete access here is a rather unfortunate byproduct of the fact that these objects are copied and then deleted while having their destination SELinux contexts.

pebenito: Can you file a bug to have it use mv/install `-Z`? I would prefer to have a comment in the policy with a reference to a bug.

policy/modules/admin/portage.if
```diff
@@ -353,6 +372,82 @@ interface(`portage_dontaudit_read_config',`
...
+## <summary>
+##	Read portage config files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`portage_read_config',`
```
pebenito, line 404: Please split to `portage_read_config` and `portage_mmap_read_config`.

pebenito (changes requested):

policy/modules/admin/dracut.te
0xC0ncord: I don't mind submitting an issue about the current behavior, but I'm not sure what `mv`/`install -Z` will do that is different from what is here currently. Let me know if I am misunderstanding something.

I think our shared concern is that the policy permits the deletion of sensitive files (e.g. `shadow_t`), so I think what we might really want is for `dracut` to be in a different domain that can only normally be transitioned to from `dracut_t` that does the cleanup, e.g. `dracut_cleanup_t`, and this domain is what is permitted to delete these files.

pebenito: If they use `install -Z` then it would be getting the expected label as it is written out, without having to label it in the tmp dir.

0xC0ncord: Ah, I see now. I'll do some investigating. If the change is easy enough then we might be able to just submit a patch.

0xC0ncord:
I haven't forgotten about this, but I need to fix a few things. It seems newer versions of dracut stopped doing the relabeling shenanigans, so the need to grant access to delete the many file types is no longer needed.

pebenito:
It seems like there may be some overlap with #1157

PR #1068, opened 2026-01-23
userdomain: Remove permissions from unprivileged user

dsugar100:
I am using userdom_unpriv_user_template() to create an unprivileged user.  That user was getting capabilities for chown, fowner and setgid.  Unrpivileged users were also allowed process ptrace, which I moved into
userdom_admin_user_template().

After dropping these permissions, those users were still able to login to the console and GUI (lightdm/icewm - on RHEL9)

Note that this does not affect admin users - which get the capabilities in userdom_admin_user_template()

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -991,10 +991,9 @@ template(`userdom_login_user_template', `
 	# User domain Local policy
 	#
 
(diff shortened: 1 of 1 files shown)


dsugar100:
With no comments at this point, can you please merge?

PR #1094, opened 2026-03-13
incus: update and extend module to work with recent incus versions

mschiff:
In this PR
- incus: Fix starting of virtual networking in Incus version >=6.15
- incus: Add rule that lets Incus detect SELinux on a system
- incus: Add new rules required for improved SELinux support in Incus 7
- virt: Add new interface required by Incus
- container: Add new interface required by Incus
- qemu:  Add new interfaces required by Incus
- qemu: Update Incus support for new Incus versions

Edit: Rewrote description

### DIFF
--- a/policy/modules/apps/qemu.if
+++ b/policy/modules/apps/qemu.if
@@ -208,6 +208,24 @@ interface(`qemu_read_state',`
 
 ########################################
 ## <summary>
+##	Get qemu process stats.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`qemu_getattr',`
+	gen_require(`
+		type qemu_t;
+	')
+
+	allow $1 qemu_t:process getattr;
+')
+
+########################################
+## <summary>
 ##	Set qemu scheduler.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/apps/qemu.if
+++ b/policy/modules/apps/qemu.if
@@ -360,6 +360,26 @@ interface(`qemu_domtrans_unconfined',`
 ########################################
 ## <summary>
 ##	Create, read, write, and delete
+##	qemu image symlinks.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`qemu_manage_image_symlinks',`
+	gen_require(`
+		type qemu_image_t;
+	')
+
+	manage_lnk_files_pattern($1, qemu_image_t, qemu_image_t)
+
+')
+
+########################################
+## <summary>
+##	Create, read, write, and delete
 ##	qemu temporary directories.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/apps/qemu.if
+++ b/policy/modules/apps/qemu.if
@@ -208,6 +208,27 @@ interface(`qemu_read_state',`
 
 ########################################
 ## <summary>
+##	Write qemu process state files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain to allow access.
+##	</summary>
+## </param>
+#
+interface(`qemu_write_state',`
+	gen_require(`
+		type qemu_t;
+	')
+
+	kernel_search_proc($1)
+	allow $1 qemu_t:dir list_dir_perms;
+	allow $1 qemu_t:file write_file_perms;
+	allow $1 qemu_t:lnk_file read_lnk_file_perms;
+')
+
+########################################
+## <summary>
 ##	Get qemu process stats.
 ## </summary>
 ## <param name="domain">
--- a/policy/modules/apps/qemu.te
+++ b/policy/modules/apps/qemu.te
@@ -57,12 +57,19 @@ tunable_policy(`qemu_full_network',`
 
 optional_policy(`
 	tunable_policy(`qemu_incus_managed',`
+		allow qemu_t self:process setrlimit;
+
 		incus_stream_connect_daemon(qemu_t)
 
 		files_create_generic_tmp_named_sockets(qemu_t)
+		# read /etc/machine-id
+		files_read_etc_runtime_files(qemu_t)
+
+		fs_read_cgroup_files(qemu_t)
 
 		kernel_read_kernel_sysctls(qemu_t)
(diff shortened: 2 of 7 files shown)


policy/modules/apps/qemu.if
```diff
@@ -206,6 +206,42 @@ interface(`qemu_read_state',`
...
+##	</summary>
+## </param>
+#
+interface(`qemu_write_qemu_files',`
+	gen_require(`
+		type qemu_t;
+	')
+
+	allow $1 qemu_t:file write_file_perms;
+')
```
pebenito, line 246: The summary and access don't match. Additionally, instead of these being "qemu files", this is qemu state.

policy/modules/apps/qemu.if
```diff
@@ -318,6 +354,26 @@ interface(`qemu_domtrans_unconfined',`
 	domtrans_pattern($1, qemu_exec_t, unconfined_qemu_t)
 ')
 
+########################################
+## <summary>
+##	Create, read, write, and delete
+##	qemu image files.
```
pebenito: Symlinks, not files

policy/modules/apps/qemu.te
```diff
@@ -56,13 +56,21 @@ tunable_policy(`qemu_full_network',`
 ')
 
 optional_policy(`
+	incusd_entry_type(qemu_t)
```
pebenito: Would you please elaborate on this? Having incusd running in qemu_t seems wrong.

policy/modules/services/container.if
```diff
@@ -702,6 +702,24 @@ interface(`container_read_all_container_state',`
...
+## </param>
+#
+interface(`container_write_container_files',`
+	gen_require(`
+		attribute container_domain;
+	')
+
+	allow $1 container_domain:file write_file_perms;
+')
+
```
pebenito, line 723: Similar to the qemu comment above, this is "all containers state"

policy/modules/services/incus.te
```diff
@@ -53,14 +53,25 @@ allow incusd_t self:vsock_socket create_stream_socket_perms;
...
+kernel_relabelfrom_unlabeled_blk_devs(incusd_t)
+kernel_relabelfrom_unlabeled_chr_devs(incusd_t)
+
 # incusd wants to load netdev-phys*
 kernel_request_load_module(incusd_t)
 
 kernel_setsched(incusd_t)
 
+selinux_getattr_dirs(incusd_t)
+selinux_read_policy(incusd_t)
```
pebenito: I couldn't find this in the incus code, can you clarify?

pebenito (changes requested):

policy/modules/services/incus.te
mschiff: It is in code yet to be merged: https://github.com/lxc/incus/pull/3216/changes

This PR adds a new selinux package in incus which in turn uses [go-selinux](https://github.com/opencontainers/selinux) from opencontainers where it is appropriate.

For example the code uses [GetEnabled()](https://github.com/lxc/incus/pull/3216/changes#diff-28c0ade07139ff2e76a707cb3b7f5fac0beea451b8cac0ddd1c010732a39c136R35) from the external go-selinux package.

Should I add comments to the policy which describe what is actually being read?

policy/modules/apps/qemu.te
mschiff: This is because incusd starts qemu like this: `/usr/bin/incusd forklimits ... -- /usr/bin/qemu-system-x86_64 `. So the exec label is set too early and incusd is starting itself with the `forklimits` arg before executing qemu.

I will try to find a solution.

policy/modules/services/incus.te
pebenito: > Should I add comments to the policy which describe what is actually being read?

I mainly am interested in this policy read. I can't think of a reason to do that off the top of my head.

mschiff: incusd reads `/sys/fs/selinux/mls` to detect if its running on a system with mls enabled so it can use different levels for each container or vm.

```
# audit(1778794152.902:3940624):                                                                                                                                                  
#  scontext="system_u:system_r:incusd_t:s0" tcontext="system_u:object_r:security_t:s0"                                                                                            
#  class="file" perms="read"                                                                                                                                                      
#  comm="incusd" exe="" path=""                                                                                                                                                   
#  message="type=AVC msg=audit(1778794152.902:3940624): avc:  denied  { read }                                                                                                    
#   for  pid=228774 comm="incusd" name="mls" dev="selinuxfs" ino=12                                                                                                               
#   scontext=system_u:system_r:incusd_t:s0                                               
#   tcontext=system_u:object_r:security_t:s0 tclass=file permissive=0 "                                                                                                           
allow incusd_t security_t:file read;
```

pebenito: In that case, we need an `selinux_read_fs()`, since `selinux_read_policy()` is an excessive permission (it has `allow $1 security_t:security read_policy;`)

policy/modules/apps/qemu.if
```diff
@@ -206,6 +206,45 @@ interface(`qemu_read_state',`
...
+#
+interface(`qemu_write_state',`
+	gen_require(`
+		type qemu_t;
+	')
+
+	kernel_search_proc($1)
+	allow $1 qemu_t:dir list_dir_perms;
+	allow $1 qemu_t:file write_file_perms;
+	allow $1 qemu_t:lnk_file write_lnk_file_perms;
```
pebenito: This should probably be `read_lnk_file_perms`, as it doesn't appear you meant to modify the symlinks under /proc/pid... I'm not sure that's even possible. Reading the symlink is sufficient to resolve the path to the real file.

policy/modules/services/incus.if
```diff
@@ -158,3 +158,22 @@ interface(`incus_admin',`
...
+##	Make incusd executable files an
+##	entrypoint for the specified domain.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	The domain for which incusd_exec_t is an entrypoint.
+##	</summary>
+## </param>
+#
+interface(`incusd_entry_type',`
```
pebenito: `incus_entry_type()`  without the `d`, for consistency with other interface naming.

pebenito (changes requested):
Minor changes.

Additionally, please reduce the number of commits.

pebenito (approved):

PR #1103, opened 2026-04-08
Add op-tee based tee supplicant policy

dburgener:


policy/modules/services/tee_supplicant.te
```diff
@@ -9,9 +9,19 @@ type tee_supplicant_t;
 type tee_supplicant_exec_t;
 init_daemon_domain(tee_supplicant_t, tee_supplicant_exec_t)
 
-########################################
+type tee_supplicant_var_lib_t;
+files_type(tee_supplicant_var_lib_t)
+files_var_lib_filetrans(tee_supplicant_t, tee_supplicant_var_lib_t, { file dir })
```
pebenito: This should be down with the `tee_supplicant_var_lib_t` access below.

pebenito (changes requested):

pebenito (approved):

PR #1104, opened 2026-04-10
refpolicy: label busybox.suid and restrict setuid BusyBox execution

jaihindy:
This change improves SELinux confinement by introducing a dedicated label for the setuid BusyBox binary, /usr/bin/busybox.suid.

Previously, both setuid and non-setuid BusyBox binaries were labeled with the generic bin_t type, making the setuid BusyBox executable broadly accessible. Treating privileged and non-privileged executables identically undermines SELinux confinement, as busybox.suid provides elevated privileges that should be tightly controlled.

To address this, a new executable type, busybox_suid_exec_t, and a reusable interface, domain_execute_busybox_suid(), are introduced to allow explicit execution of the setuid BusyBox binary only by domains with a legitimate functional requirement (e.g. ping, traceroute, passwd, vlock, mount, and local login helpers). To preserve existing functionality, the interface is applied selectively to domains that require execution of BusyBox applets with the setuid bit set.

This keeps access explicit and auditable while avoiding broad permissions and preserving least-privilege principles.

Additionally, ping is granted controlled PTY access via the existing init_use_script_ptys() interface to allow normal interactive operation (stdout/stderr and signal handling) without unnecessary device access.

### DIFF
--- a/policy/modules/admin/netutils.te
+++ b/policy/modules/admin/netutils.te
@@ -22,10 +22,13 @@ files_tmp_file(netutils_tmp_t)
 type ping_t;
 type ping_exec_t;
 init_system_domain(ping_t, ping_exec_t)
+domain_execute_busybox_suid(ping_t)
+init_use_script_ptys(ping_t)
 
 type traceroute_t;
 type traceroute_exec_t;
 init_system_domain(traceroute_t, traceroute_exec_t)
+domain_execute_busybox_suid(traceroute_t)
 
 ########################################
 #
--- a/policy/modules/admin/usermanage.te
+++ b/policy/modules/admin/usermanage.te
@@ -384,6 +384,7 @@ userdom_read_user_tmp_files(passwd_t)
 # user generally runs this from their home directory, so do not audit a search
 # on user home dir
 userdom_dontaudit_search_user_home_content(passwd_t)
+domain_execute_busybox_suid(passwd_t)
 
 optional_policy(`
 	nscd_run(passwd_t, passwd_roles)
--- a/policy/modules/apps/vlock.te
+++ b/policy/modules/apps/vlock.te
@@ -41,3 +41,5 @@ miscfiles_read_localization(vlock_t)
 
 userdom_dontaudit_search_user_home_dirs(vlock_t)
 userdom_use_user_terminals(vlock_t)
+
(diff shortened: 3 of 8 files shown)


pebenito:
The use of setuid is already controlled by capability:setuid. If the domain that runs busybox.suid doesn't have that, then they can't get the setuid effect by running the binary. This does not seem to appreciably improve things. Is there something I'm missing?

jaihindy:
Thanks for the clarification — I agree that the setuid behavior itself is ultimately gated by capability:setuid, and a domain without that capability will not gain the setuid effect simply by executing the binary. The motivation for this change is primarily about strengthening SELinux confinement at the MAC layer, rather than replacing or bypassing the DAC
and capability checks.
In the previous setup, both the setuid and non‑setuid BusyBox binaries were labeled bin_t. Since bin_t is executable by a large number of domains, the MAC policy did not meaningfully distinguish between privileged and non‑privileged BusyBox execution. While the capability check still applies, SELinux had no way to restrict or reason about which domainsare allowed to exec a privileged binary in the first place.This effectively makes the MAC check too generic in this case, and relies almost entirely on DAC/capabilities for controlling access to a setuid binary. By introducing a dedicated executable type (busybox_suid_exec_t) and requiring explicit policy via domain_execute_busybox_suid(), this change adds an additional MAC‑level control. Only domains with a clear functional need are permitted to execute the setuid BusyBox binary, making execution of a privileged binary explicit, auditable, and consistent with least‑privilege principles. In other words, capability:setuid remains the gate for what happens after exec, but this change allows SELinux to more precisely control who is allowed to exec a privileged binary at all, providing a more robust, layered defence.
If this distinction is still considered insufficient to justify the added labeling and interface, I’m happy to adjust or drop the change, but the intent was to avoid treating privileged and non‑privileged executables identically at the MAC layer.

jaihindy:
hi @pebenito , 
just a gentle ping on this in case my previous comment was missed.
Happy to clarify or adjust the change further if needed. 
Thanks for your time and review.

pebenito:
I still don't agree about changing the busybox label.

jaihindy:
Thanks for the review. Since changing the busybox label is not acceptable upstream,
I am closing this PR. I will resend a smaller change addressing only the functional
ping permission issue without relabeling.

PR #1105, opened 2026-04-16
tee_supplicant: Add necessary SELinux policy for tee_supplicant domain

wenjz-qualcomm:
Define some new interfaces to access /var/lib/tee, /var/lib/qtee_supplicant
and /var/lib/tee/qtee_supplicant.

Grant the nescessary permission to tee_supplicant for resolving
AVC denials in enforcing mode.

Upstream-Status: Inappropriate [embedded specific]

In the commit write the Upstream-status just because yocto project request it, the status isn't the truly status, just to maintain a uniform format.

### DIFF
--- a/policy/modules/services/tee_supplicant.fc
+++ b/policy/modules/services/tee_supplicant.fc
@@ -1,2 +1,4 @@
 /usr/bin/qtee_supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
 /usr/sbin/tee-supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
+
+/var/lib/tee(/.*)?                    gen_context(system_u:object_r:tee_supplicant_var_lib_t,s0)
--- a/policy/modules/services/tee_supplicant.if
+++ b/policy/modules/services/tee_supplicant.if
@@ -1,5 +1,5 @@
 ## <summary>tee_supplicant</summary>
-#
+##
 ## <desc>
 ## qtee_supplicant is a userspace supplicant daemon that
 ## services callback requests from QTEE via the Linux TEE subsystem.
@@ -8,3 +8,23 @@
 ##
 ## https://github.com/qualcomm/minkipc/tree/main/qtee_supplicant
 ## </desc>
+
+#####################
+## <summary>
+##  Allow the specified domain to create
+##  objects in /var/lib with an automatic
+##  transition to the tee_supplicant var lib type.
+## </summary>
+## <param name="domain">
+##  <summary>
+##  Domain allowed access.
+##  </summary>
+## </param>
+#
+interface(`tee_supplicant_var_lib_filetrans',`
+       gen_require(`
+               type tee_supplicant_var_lib_t;
+       ')
+
+       files_var_lib_filetrans($1, tee_supplicant_var_lib_t, dir, "qtee_supplicant")
+')
--- a/policy/modules/services/tee_supplicant.te
+++ b/policy/modules/services/tee_supplicant.te
@@ -5,12 +5,20 @@ policy_module(tee_supplicant)
 # Declarations
 #
 
+## <desc>
+##  <p>
+##  Enable rules specific to qtee_supplicant.
+##  </p>
+## </desc>
+gen_tunable(tee_supplicant_qtee, false)
+
 type tee_supplicant_t;
 type tee_supplicant_exec_t;
 init_daemon_domain(tee_supplicant_t, tee_supplicant_exec_t)
 
 type tee_supplicant_var_lib_t;
 files_type(tee_supplicant_var_lib_t)
+files_mountpoint(tee_supplicant_var_lib_t)
 
 #########################################
 #
@@ -25,3 +33,34 @@ dev_rw_tee(tee_supplicant_t)
 dev_rw_tee_priv(tee_supplicant_t)
 
 kernel_read_vm_overcommit_sysctl(tee_supplicant_t)
+
+# Access qtee_supplicant to access UFS BSG device
+storage_read_scsi_generic_cond(tee_supplicant_t,tee_supplicant_qtee)
+storage_write_scsi_generic_cond(tee_supplicant_t,tee_supplicant_qtee)
+
+tunable_policy(`tee_supplicant_qtee',`
+
+        # Access qtee_supplicant to request sys_rawio capability
+        allow tee_supplicant_t self:capability sys_rawio;
+
+        # Allow qtee_supplicant to block system suspend by wake_lock
+        allow tee_supplicant_t self:capability2 block_suspend;
+
+        # Access qtee_supplicant to open/read /sys/firmware/devicetree/base/compatible
+        dev_read_sysfs(tee_supplicant_t)
+
+        # Access qtee_supplicant to write /sys/power/wake_lock
+        dev_write_sysfs(tee_supplicant_t)
+
+        # Access tee_supplicant to read /var
+        files_list_var(tee_supplicant_t)
+
+        # Access qtee_supplicant to visit /var/lib
+        files_list_var_lib(tee_supplicant_t)
+
+        # Access qtee_supplicant to access /proc/cmdline
+        kernel_read_system_state(tee_supplicant_t)
+
+        # Access qtee_supplicant to send logs to systemd journal
+        logging_send_syslog_msg(tee_supplicant_t)
+')
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -1524,6 +1524,10 @@ optional_policy(`
 ')
 
 optional_policy(`
+	tee_supplicant_var_lib_filetrans(initrc_t)
+')
+
+optional_policy(`
 	udev_manage_runtime_files(initrc_t)
 	udev_manage_runtime_dirs(initrc_t)
 	udev_manage_rules_files(initrc_t)
--- a/policy/modules/kernel/storage.if
+++ b/policy/modules/kernel/storage.if
@@ -549,6 +549,36 @@ interface(`storage_read_scsi_generic',`
 
 ########################################
 ## <summary>
+##      Allow the caller to directly read, in a
+##      generic fashion, from any SCSI device
+##      if a tunable is set.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+## <param name="tunable">
+##      <summary>
+##      Tunable to depend on
+##      </summary>
+## </param>
+#
+interface(`storage_read_scsi_generic_cond',`
+        gen_require(`
+                attribute scsi_generic_read;
+                type scsi_generic_device_t;
+        ')
+
+        typeattribute $1 scsi_generic_read;
+        tunable_policy(`$2',`
+                dev_list_all_dev_nodes($1)
+                allow $1 scsi_generic_device_t:chr_file read_chr_file_perms;
+        ')
+')
+
+########################################
+## <summary>
 ##	Allow the caller to directly write, in a
 ##	generic fashion, from any SCSI device.
 ##	This is extremely dangerous as it can bypass the
@@ -570,6 +600,36 @@ interface(`storage_write_scsi_generic',`
 	dev_list_all_dev_nodes($1)
 	allow $1 scsi_generic_device_t:chr_file write_chr_file_perms;
 	typeattribute $1 scsi_generic_write;
+')
+
+########################################
+## <summary>
+##      Allow the caller to directly write, in a
+##      generic fashion, from any SCSI device
+##      if a tunable is set.
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed access.
+##      </summary>
+## </param>
+## <param name="tunable">
+##      <summary>
+##      Tunable to depend on
+##      </summary>
+## </param>
+#
+interface(`storage_write_scsi_generic_cond',`
+        gen_require(`
+                attribute scsi_generic_write;
+                type scsi_generic_device_t;
+        ')
+
+        typeattribute $1 scsi_generic_write;
+        tunable_policy(`$2',`
+                dev_list_all_dev_nodes($1)
+                allow $1 scsi_generic_device_t:chr_file write_chr_file_perms;
+        ')
 ')
 
 ########################################
--- a/testing/sechecker.ini
+++ b/testing/sechecker.ini
@@ -221,6 +221,7 @@ exempt_source = abrt_t              # Conditional access (allow_raw_memory_acces
                 sosreport_t         # Conditional access (allow_raw_memory_access)
                 spc_t
                 sysadm_t            # System admin role
+                tee_supplicant_t    # Access qtee_supplicant to request sys_rawio capability
                 udev_t
                 vbetool_t           # Conditional access (allow_raw_memory_access)
                 vmware_t


policy/modules/services/tee_supplicant.fc
```diff
@@ -1,2 +1,7 @@
 /usr/bin/qtee_supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
 /usr/sbin/tee-supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
+
+/var/lib/qtee_supplicant(/.*)?        gen_context(system_u:object_r:qtee_supplicant_data_t,s0)
```
dburgener: I had created `tee_supplicant_var_lib_t` for this.  I added the filetrans originally, but not the file_contexts.  The file_contexts should definitely match, and the var_lib_t suffix I believe is more standard here.

policy/modules/services/tee_supplicant.fc
```diff
@@ -1,2 +1,7 @@
...
+/var/lib/tee/qtee_supplicant(/.*)?    gen_context(system_u:object_r:qtee_supplicant_data_t,s0)
+
+/var/lib/tee(/.*)?                    gen_context(system_u:object_r:tee_supplicant_persist_t,s0)
```
dburgener: Is the distinction between the _data_t and _persist_t types meaningful security-wise?  It looks like this PR adds manage permissions on both.  I ask because I'm not sure the added granularity adds value, and also because the filetrans rules would need some tweaking to match this structure if we maintain the distinction.  But I'm not sure why these wouldn't both just be tee_supplicant_var_lib_t?

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,87 @@
...
+## /var/lib/tee/qtee_supplicant (data files for TA)
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed dir and file operations
+##      under /var/lib/qtee_supplicant
+##      </summary>
+## </param>
+#
+interface(`qtee_supplicant_manage_data',`
```
dburgener: Across the board, the convention in .if files is that they should start with the name of the module (`tee_supplicant` in this case)

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,87 @@
 ##
 ## https://github.com/qualcomm/minkipc/tree/main/qtee_supplicant
 ## </desc>
+
+#####################
+## <summary>
+## Allow dir and file operations
+## under /var/lib/qtee_supplicant and
```
dburgener: I think it's best to not explicitly mention the paths here.  In this case, the paths are encoding the qtee path assumptions, but we'll want the same labeling to handle the optee case as well.

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,87 @@
...
+##   </summary>
+## </param>
+#
+interface(`sfs_config_create_data_dir',`
+    gen_require(`
+        type var_lib_t;
+        type qtee_supplicant_data_t;
+        type initrc_t;
+    ')
+    filetrans_pattern(initrc_t, var_lib_t, qtee_supplicant_data_t, dir, "qtee_supplicant")
```
dburgener: This interface doesn't take any arguments as written.  It should not explicitly mention initrc_t here, that should be the `$1` argument, and then this interface should get called from init.te.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +25,40 @@ dev_rw_tee(tee_supplicant_t)
 dev_rw_tee_priv(tee_supplicant_t)
 
 kernel_read_vm_overcommit_sysctl(tee_supplicant_t)
+
+# Access /var/lib/qtee_supplicant
+qtee_supplicant_manage_data(tee_supplicant_t)
```
dburgener: Typically, when we're dealing with types in the same module, we just add the manage pattern rather than using an external interface for it.

In this case, if we switch to tee_supplicant_var_lib_t for these labels as I recommend, the manage patterns are already implemented above.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +25,40 @@ dev_rw_tee(tee_supplicant_t)
...
+
+# Access qtee_supplicant to access /proc/cmdline
+kernel_read_system_state(tee_supplicant_t)
+
+# Access qtee_supplicant to access UFS BSG device
+storage_read_scsi_generic(tee_supplicant_t)
+storage_write_scsi_generic(tee_supplicant_t)
+
+# Access qtee_supplicant to request sys_rawio capability
+allow tee_supplicant_t self:capability sys_rawio;
```
dburgener: Do we understand why qtee_supplicant needs this capability?  Is it related to the use of the UFS BSG device?

I ask because we're also using this policy for the optee case and haven't noticed this need.  Maybe it's just a gap in our optee testing, but if qtee needs it and optee doesn't, this might be a good candidate for a tunable, depending on what @pebenito thinks.  This and the scsi access could all go behind a `tee_supplicant_use_ufs_bsg` or something?

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,87 @@
...
+##   <summary>
+##   Domain allowed access.
+##   </summary>
+## </param>
+#
+interface(`tee_supplicant_search_data_dir',`
+    gen_require(`
+        type default_t;
+    ')
+    allow $1 default_t:dir search;
```
dburgener: This suggests that /data is mislabeled on your system.

If tee_supplicant really needs this access, this interface just reimplements `files_search_default()`, and that should be used instead.  Such a change is probably not appropriate for upstream IMO.

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: got it. will update.

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: got it. will update.

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: got it. will update. And if use the method you mentioned, will need add 
gen_require(`
    type initrc_t;
')
in the te file.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: got it. After we have a result about above issue, I will follow the result to change the content.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: I'm not sure why qtee_supplicant needs this capability. Just based on the avc denied information add this one. Maybe we could wait the comment of @pebenito

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: got it. I will directly use the files_search_default().

policy/modules/services/tee_supplicant.fc
wenjz-qualcomm: The /dev/sdf1 will mount on /var/lib/tee. The difference between /var/lib/tee with other two path is suggested by @hl_harshaldev@yahoo.co.in. Pls help reply to this comment.

policy/modules/services/tee_supplicant.fc
wenjz-qualcomm: ok, I think we should have a basic consensus about above three directories. Afak, the /dev/sdf1 will mount on /var/lib/tee and the /var/lib/qtee_supplicant is created by progress sfs_config.

harshaldev27: Okay this I agree with, we don't need separate labelling for things under `/var/lib` for folders used by either OPTEE's tee-supplicant or QTEE's qtee-supplicant.

policy/modules/services/tee_supplicant.fc
harshaldev27: I agree with @dburgener here. Let's keep the same label for all files under `/var/lib/tee` as well, same as `tee_supplicant_var_lib_t`. There is no security wise distinction between these two directories which requires further granularity. Either tee-supplicant or `qtee-supplicant` can access them.

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,69 @@
...
+## /var/lib/tee/qtee_supplicant (data files for TA)
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed dir and file operations
+##      under /var/lib/qtee_supplicant
+##      </summary>
+## </param>
+#
+interface(`tee_supplicant_manage_data',`
```
harshaldev27: We probably don't need this interface if both the `/var/lib/qtee_supplicant` and `/var/lib/tee/*` have the same `tee_supplicant_var_lib_t` label. Since the `files_var_lib_filetrans` already grants manage permissions.

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,69 @@
...
+##   under /var/lib/tee (PERSIST partition)
+## </summary>
+## <param name="domain">
+##      <summary>
+##      Domain allowed dir and file operations
+##      under /var/lib/tee
+##      </summary>
+## </param>
+#
+interface(`tee_supplicant_manage_persist',`
```
harshaldev27: Don't need this interface either.

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,69 @@
...
+##   Domain allowed access.
+##   </summary>
+## </param>
+#
+interface(`sfs_config_create_data_dir',`
+    gen_require(`
+        type var_lib_t;
+        type qtee_supplicant_data_t;
+    ')
+    filetrans_pattern($1, var_lib_t, qtee_supplicant_data_t, dir, "qtee_supplicant")
```
harshaldev27: I think we can omit this interface as well, instead of using this, we can use a macro similar to what we have in tee_supplicant.te line 22.
```
file_var_lib_filetrans(<domain_name_for_sfs_config>, tee_supplicant_var_lib_t, { file dir })
```

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +25,45 @@ dev_rw_tee(tee_supplicant_t)
 dev_rw_tee_priv(tee_supplicant_t)
 
 kernel_read_vm_overcommit_sysctl(tee_supplicant_t)
+
+# Access /var/lib/qtee_supplicant
+tee_supplicant_manage_data(tee_supplicant_t)
```
harshaldev27: Not needed.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +25,45 @@ dev_rw_tee(tee_supplicant_t)
...
+
+# Access /var/lib/tee
+tee_supplicant_manage_persist(tee_supplicant_t)
```
harshaldev27: Not needed as mentioned above.

policy/modules/services/tee_supplicant.te
harshaldev27: We need this because our RPMB listener writes RPMB packets received from QTEE after some processing directly into UFS_BSG device. I don't know how OPTEE's tee-supplicant is doing it without raw IO capability. But I agree, we can have a interface here called `tee_supplicant_use_ufs_bsg`. I don't think it really makes an issue for OPTEE tee-supplicant to have this permission.

harshaldev27 (changes requested):

harshaldev27:
Hi @wenjz-qualcomm , Can you please update the Commit message to explicitly state that we are adding these new interface for qtee-supplicant which requires more permissions than OPTEE's tee-supplicant. Also, please describe what all permissions (high level) we are adding and why. Please also fix the indentation on your commit message body.

Finally, drop this "Upstream-Status: Inappropriate [embedded specific]" from the commit message.

wenjz-qualcomm:
Hi @harshaldev27 and @dburgener, I already modified all mentioned comments, pls help me review the latest version. Thanks.

policy/modules/services/tee_supplicant.if
```diff
@@ -1,5 +1,5 @@
 ## <summary>tee_supplicant</summary>
-#
+##
```
harshaldev27, line 2: Please don't add unnecessary #s here. This file should be un-touched.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +26,42 @@ dev_rw_tee(tee_supplicant_t)
...
+dev_write_sysfs(tee_supplicant_t)
+
+# Allow qtee_supplicant to block system suspend (wake_lock)
+allow tee_supplicant_t self:capability2 block_suspend;
+
+# Allow qtee_supplicant to search /data directory
+files_search_default(tee_supplicant_t)
+
+# Allow sfs_config to relable /var/lib/qtee_supplicant
+optional_policy(`
```
harshaldev27: We need some guidance here. I believe the correct way to do this would be to label `/usr/bin/sfs_config`, and then have a `file_var_lib_filetrans()` macro that labels everything it creates as `tee_supplicant_var_lib_t`.
And if this is the way we go, you don't need to change anything in `tee_supplicant.fc`. However, we will still need to generate a context for `sfs_config` in some `*.fc` file.

@dburgener /  @pebenito  perhaps you can help guide here.

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: Done.

policy/modules/services/tee_supplicant.te
pebenito: My preference would be to put significant access that is specific to a particular implementation, such as this and the sg access, under tunables that reflect the specific tee implementation, such as `tee_supplicant_qtee` or `tee_supplicant_optee`.

policy/modules/services/tee_supplicant.te
pebenito: You would need to add an interface in `tee_supplicant.if` that has the below implementation, with `initrc_t` replaced with `$1`. Then in `initrc.te`, add an optional block that calls this new interface.

Similar interface: https://github.com/SELinuxProject/refpolicy/blob/bf750a6d8368741f4487c415e4fb0a0c7bc7238c/policy/modules/services/container.if#L2710

wenjz-qualcomm: Modified.

policy/modules/services/tee_supplicant.fc
```diff
@@ -1,2 +1,5 @@
 /usr/bin/qtee_supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
 /usr/sbin/tee-supplicant      --      gen_context(system_u:object_r:tee_supplicant_exec_t,s0)
+
+/var/lib/qtee_supplicant(/.*)?        gen_context(system_u:object_r:tee_supplicant_var_lib_t,s0)
```
harshaldev27: With the `tee_supplicant_var_lib_filetrans` interface now available, I don't think you need this. Check once.

policy/modules/services/tee_supplicant.te
harshaldev27: Hi @pebenito, does that mean that we should label `/usr/bin/qtee_supplicant` as `tee_supplicant_qtee`, and then for all the permissions that it needs common with `tee_supplicant_t` call the same interfaces? Or is there a better way to do this?

harshaldev27 (changes requested):

policy/modules/services/tee_supplicant.fc
wenjz-qualcomm: sure, you are right, already modified.

policy/modules/services/tee_supplicant.te
pebenito: > does that mean that we should label `/usr/bin/qtee_supplicant` as `tee_supplicant_qtee`

No, it would be a conditional policy block. Declare a tunable:
```
gen_tunable(aide_mmap_files, false)
```
There is a required documentation block not shown above.

Then use it in a tunable block:
```
tunable_policy(`tee_supplicant_qtee ',`
	allow tee_supplicant_t self:capability sys_rawio;

	# Access qtee_supplicant to access UFS BSG device
	storage_read_scsi_generic(tee_supplicant_t)
	storage_write_scsi_generic(tee_supplicant_t)

	...
')
```
See `aide.te`, which has a small policy with a couple tunables.

policy/modules/services/tee_supplicant.if
pebenito: That would be the correct thing to do for the access, but I'd rather not include this access. `default_t` means unknown security properties. The labeling should be corrected first and then the rule here can be updated accordingly with the specific required access.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: Hi @pebenito ,I already change these policy to tunable_policy, just because storage_read/write_scsi_generic will use typeattribute and it can't put in tunable policy, so I put them outside.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +33,34 @@ dev_rw_tee(tee_supplicant_t)
...
+
+tunable_policy(`tee_supplicant_qtee',`
+
+        # Access tee_supplicant to read /var
+        files_list_var(tee_supplicant_t)
+
+        # Access qtee_supplicant to visit /var/lib
+        files_list_var_lib(tee_supplicant_t)
+
+        # Accesss qtee_supplicant to access compatible of device tree
```
harshaldev27: Can you fix the spelling of Access? Code spell is complaining:

```
Run codespell --skip Changelog,Changelog.contrib,Changelog.old --ignore-words-list busses,chage,doesnt,lik,msdos,nd,racoon,shouldnt,startd,te,thats,xwindows --context 1 .
: 
>         # Accesss qtee_supplicant to access compatible of device tree
:         dev_read_sysfs(tee_supplicant_t)
./policy/modules/services/tee_supplicant.te:49: Accesss ==> Access
Error: Process completed with exit code 65.
```

harshaldev27 (changes requested):

policy/modules/services/tee_supplicant.te
pebenito: We have a way to work around that. You'll need to create new interfaces in `storage.if` that are similar to this: https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/kernel/storage.if#L208

Then in `tee_supplicant.te` you would call `storage_write_scsi_generic_cond(tee_supplicant_t, tee_supplicant_qtee)` and then the access would be controlled by the tunable.

pebenito:
> Hi @pebenito, does that mean that we should label /usr/bin/qtee_supplicant as tee_supplicant_qtee, and then for all the permissions that it needs common with tee_supplicant_t call the same interfaces? Or is there a better way to do this?

No, `/usr/bin/qtee_supplicant` would remain `tee_supplicant_exec_t`. The entire .te file should use tee_supplicant_t. Any permissions known to be specific to qtee_supplicant should be in the tunable. Then on your system, you would enable the `tee_supplicant_qtee` Boolean to enable the rules.

harshaldev27:
> > Hi @pebenito, does that mean that we should label /usr/bin/qtee_supplicant as tee_supplicant_qtee, and then for all the permissions that it needs common with tee_supplicant_t call the same interfaces? Or is there a better way to do this?
> 
> No, `/usr/bin/qtee_supplicant` would remain `tee_supplicant_exec_t`. The entire .te file should use tee_supplicant_t. Any permissions known to be specific to qtee_supplicant should be in the tunable. Then on your system, you would enable the `tee_supplicant_qtee` Boolean to enable the rules.

Understood. I believe @wenjz-qualcomm has already accomodated this direction in the lastest pushed changes.

policy/modules/services/tee_supplicant.te
```diff
@@ -5,12 +5,20 @@ policy_module(tee_supplicant)
 # Declarations
 #
 
+## <desc>
+##  <p>
+##  Control if tee_supplicant_t can access storage device.
```
pebenito: I think a description like "Enable rules specific to qtee_supplicant" is more accurate.

wenjz-qualcomm: Already modified.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: > We have a way to work around that. You'll need to create new interfaces in `storage.if` that are similar to this: https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/kernel/storage.if#L208
> 
> Then in `tee_supplicant.te` you would call `storage_write_scsi_generic_cond(tee_supplicant_t, tee_supplicant_qtee)` and then the access would be controlled by the tunable.

Hi @pebenito , pls review latest version.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: Done.

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +33,34 @@ dev_rw_tee(tee_supplicant_t)
...
+        kernel_read_system_state(tee_supplicant_t)
+
+        # Access qtee_supplicant to request sys_rawio capability
+        allow tee_supplicant_t self:capability sys_rawio;
+
+        # Access qtee_supplicant to write wake_lock
+        dev_write_sysfs(tee_supplicant_t)
+
+        # Allow qtee_supplicant to block system suspend (wake_lock)
+        allow tee_supplicant_t self:capability2 block_suspend;
```
pebenito: Please order the lines in the tunable block according to the [StyleGuide](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide).

policy/modules/services/tee_supplicant.te
```diff
@@ -25,3 +33,34 @@ dev_rw_tee(tee_supplicant_t)
...
+tunable_policy(`tee_supplicant_qtee',`
+
+        # Access tee_supplicant to read /var
+        files_list_var(tee_supplicant_t)
+
+        # Access qtee_supplicant to visit /var/lib
+        files_list_var_lib(tee_supplicant_t)
+
+        # Access qtee_supplicant to access compatible of device tree
+        dev_read_sysfs(tee_supplicant_t)
```
pebenito: Which files/directories in sysfs does this read? We are slowly adding more labeling in sysfs. Please put comments that summarize what is accessed in sysfs.

pebenito (changes requested):

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: Ordered.

policy/modules/services/tee_supplicant.te
wenjz-qualcomm: commented.

wenjz-qualcomm:
Hi @pebenito , could you please help me to do the further review?

harshaldev27 (approved):
This is looking good from my side.

policy/modules/system/init.te
```diff
@@ -1523,6 +1523,10 @@ optional_policy(`
 	sysnet_read_dhcpc_state(initrc_t)
 ')
 
+optional_policy(`
+	tee_supplicant_var_lib_filetrans(initrc_t)
```
pebenito, line 1527: This is too broad. This will make any file or directory it creates in `/var/lib` as `tee_supplicant_var_lib_t`. The interface should be revised like https://github.com/SELinuxProject/refpolicy/blob/436d78b106a7c336edb9accc01019df9f86c95a5/policy/modules/kernel/files.if#L2192 so the object class and dir name can be specified at the call site.

policy/modules/services/tee_supplicant.if
pebenito: This change still remains.

pebenito (changes requested):

policy/modules/services/tee_supplicant.if
wenjz-qualcomm: I must remain this changs. If not will encounter compile error. Previously could work well because there isn't present true interface.

policy/modules/system/init.te
wenjz-qualcomm: I think the interface just used for tee_supplicant. So I hard coding the filetrans_pattern in .if file.

policy/modules/services/tee_supplicant.if
```diff
@@ -8,3 +8,23 @@
...
+##  Domain allowed access.
+##  </summary>
+## </param>
+#
+interface(`tee_supplicant_var_lib_filetrans',`
+       gen_require(`
+               type tee_supplicant_var_lib_t;
+       ')
+
+       filetrans_pattern($1, var_lib_t, tee_supplicant_var_lib_t, dir, "qtee_supplicant")
```
pebenito: Sorry I was unclear with my previous comment: You should still use `files_var_lib_filetrans`. The call should have all the same parameters, except `var_lib_t`.

wenjz-qualcomm: It's ok. I already update this place. Pls review them again.

pebenito:
Policy looks good, one last change needed to pass the CI checks. You'll need to add `tee_supplicant_t` to the sechecker.ini here: https://github.com/SELinuxProject/refpolicy/blob/1275976c382e39e35c2fc98cab0f8b78f9f5c7d8/testing/sechecker.ini#L224 Please make sure to add a comment why CAP_SYS_RAWIO is necessary (same comment as in the policy is ok)

wenjz-qualcomm:
> Policy looks good, one last change needed to pass the CI checks. You'll need to add `tee_supplicant_t` to the sechecker.ini here:
> 
> https://github.com/SELinuxProject/refpolicy/blob/1275976c382e39e35c2fc98cab0f8b78f9f5c7d8/testing/sechecker.ini#L224
> 
> Please make sure to add a comment why CAP_SYS_RAWIO is necessary (same comment as in the policy is ok)

Thanks for the information. Added.

pebenito (approved):

Issue #1100, opened 2026-03-31
Is unconfined_mount_t allowed to run with system role ?


GargiQcom:
Hi,

In our setup, we use `adb shell` to perform certain operations, where the shell runs with developer (root) privileges.

Since `meta-selinux` does not provide an `adb` domain, the SELinux policy for `adb shell` has been added via `meta-openembedded`. The relevant patch is available here:  
<https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted/0001-Added-sepolicy-for-adb-service.patch>

On the target, the `adb shell` context is as follows:

```
adb shell id
uid=0(root) gid=0(root) groups=0(root) context=system_u:system_r:unconfined_t:s0
```

When running the `mount` command from this shell, we encounter the following SELinux error:

    type=SELINUX_ERR msg=audit(44.591:158): op=security_compute_sid
    invalid_context="system_u:system_r:unconfined_mount_t:s0"
    scontext=system_u:system_r:unconfined_t:s0
    tcontext=system_u:object_r:mount_exec_t:s0
    tclass=process

This suggests that `unconfined_mount_t` is not permitted to run with the `system` role.

Could you please advise whether it is acceptable to allow `unconfined_mount_t` to run with the `system` role as well?  
Specifically, is the following change to refpolicy considered acceptable?
<https://github.com/openembedded/meta-openembedded/pull/1035/changes>

jaihindy:
Thank you for the clarification.

We understand that `system_r:unconfined_t` was historically used early in the targeted policy, and that today both `init_t` and `initrc_t` are unconfined. From a policy model perspective, the transition may indeed seem unnecessary.

However, in practice, without this change, basic commands such as `mount` fail when executed from the shell. As per prior guidance from refpolicy maintainers, running `adb shell` in the `unconfined_t` domain made sense, since `unconfined_t` is also suitable for interactive shells. With `adb shell` running in `initrc_t` (i.e., without this change), we observe that even common commands like `ping` do not work as expected.

Given this, our intent with this change is to preserve functional parity for basic shell usage while remaining aligned with the refpolicy model. If this approach is not acceptable, then it likely means the earlier policy change of running `adb shell` in `unconfined_t` should also be reverted, as it no longer provides the intended benefits.

[...]

PR #1109, opened 2026-04-30
pipewire: add policy for PipeWire multimedia daemon

rchiluka29:
Add a new SELinux policy module for the PipeWire multimedia daemon. PipeWire is a low-latency audio and video server that replaces PulseAudio and JACK on modern Linux systems.

The policy confines the pipewire_t domain with:
- Process control: fork, signal handling and scheduling (sys_nice) for low-latency audio operation
- Runtime socket and directory management under /run/pipewire
- ALSA sound device access via dev_read_sound/dev_write_sound
- Shared memory (tmpfs/memfd) for zero-copy audio buffer handling
- sysfs and procfs read access for hardware discovery
- System logging via syslog and systemd journal
- PulseAudio runtime compatibility via optional_policy
- Suppressed audit noise for expected /proc scan denials

pipewire-pulse is a symlink to the pipewire binary and transitions into the same pipewire_t domain, so no separate domain is needed.

### DIFF
--- a/policy/modules/apps/pipewire.fc
+++ b/policy/modules/apps/pipewire.fc
@@ -1,0 +1,21 @@
+# PipeWire daemon executable.
+/usr/bin/pipewire		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+
+# pw-cat is a standalone client utility (plays/captures audio via a running
+# PipeWire daemon); it runs in pipewire_client_t, not pipewire_t.
+# pw-play, pw-record, pw-midiplay etc. are symlinks to pw-cat and
+# automatically inherit this label — no separate entries needed.
+/usr/bin/pw-cat			--	gen_context(system_u:object_r:pipewire_client_exec_t,s0)
+
+# System-service runtime directory (/run/pipewire).
+/run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
+
+# User-session runtime sockets live under $XDG_RUNTIME_DIR/pipewire and are
+# labelled at runtime via userdom_user_runtime_filetrans; no static entry needed.
+
+# Persistent system-service state.
+/var/lib/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_var_lib_t,s0)
+
+# User home configuration and state.
+HOME_DIR/\.config/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+HOME_DIR/\.local/share/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
--- a/policy/modules/apps/pipewire.fc
+++ b/policy/modules/apps/pipewire.fc
@@ -1,0 +1,10 @@
+HOME_DIR/\.config/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+HOME_DIR/\.local/share/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+
+/run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
+/run/user/%{USERID}/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_runtime_t,s0)
+
+/usr/bin/pipewire		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+/usr/bin/pw-cat		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+
+/var/lib/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_var_lib_t,s0)
--- a/policy/modules/apps/pipewire.if
+++ b/policy/modules/apps/pipewire.if
@@ -1,0 +1,225 @@
+## <summary>PipeWire multimedia daemon.</summary>
+
+########################################
+## <summary>
+##	Role access for PipeWire (user-service mode).
+##	Call this from unprivuser.te / staff.te in the same
+##	way pulseaudio_role is called.
+## </summary>
+## <param name="role_prefix">
+##	<summary>
+##	The prefix of the user role (e.g., "user" for user_r).
+##	</summary>
+## </param>
+## <param name="user_domain">
+##	<summary>
+##	User domain for the role (e.g., user_t).
+##	</summary>
+## </param>
+## <param name="user_exec_domain">
+##	<summary>
+##	User exec domain for execute and transition access
+##	(e.g., user_application_exec_domain).
+##	</summary>
+## </param>
+## <param name="role">
+##	<summary>
+##	Role allowed access (e.g., user_r).
+##	</summary>
+## </param>
+#
+template(`pipewire_role',`
+	gen_require(`
+		attribute_role pipewire_roles;
+		type pipewire_t, pipewire_exec_t;
+		type pipewire_client_t, pipewire_client_exec_t;
+		type pipewire_home_t, pipewire_tmpfs_t, pipewire_runtime_t;
+	')
+
+	roleattribute $4 pipewire_roles;
+
+	domtrans_pattern($2, pipewire_exec_t, pipewire_t)
+	domtrans_pattern($3, pipewire_exec_t, pipewire_t)
+	domtrans_pattern($2, pipewire_client_exec_t, pipewire_client_t)
+	domtrans_pattern($3, pipewire_client_exec_t, pipewire_client_t)
+
+	allow $3 pipewire_t:process { ptrace signal_perms };
+	allow $3 pipewire_t:fd use;
+	ps_process_pattern($3, pipewire_t)
+
+	allow pipewire_t $3:unix_stream_socket connectto;
+	allow pipewire_t $3:process signull;
+
+	# Allow the user domain to manage and relabel pipewire home state files
+	# (e.g. restorecon after reinstall, stale file cleanup).
+	allow $2 pipewire_home_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $2 pipewire_home_t:file { mmap_manage_file_perms relabel_file_perms };
+	allow $2 pipewire_home_t:lnk_file { manage_lnk_file_perms relabel_lnk_file_perms };
+
+	# Allow the user domain to manage and relabel runtime sock files
+	# (e.g. stale socket cleanup after daemon crash).
+	allow $2 pipewire_runtime_t:dir { manage_dir_perms relabel_dir_perms };
+	allow $2 pipewire_runtime_t:sock_file { manage_sock_file_perms relabel_sock_file_perms };
+
+	# Allow the user domain to manage shared memory files created by the daemon.
+	pipewire_mmap_rw_tmpfs_files($2)
+	allow $2 pipewire_tmpfs_t:file relabel_file_perms;
+
+	optional_policy(`
+		systemd_user_app_status($1, pipewire_t)
+		systemd_user_app_socket_create($1, pipewire_t, pipewire_runtime_t)
+	')
+')
+
+########################################
+## <summary>
+##	Execute a domain transition to run the PipeWire daemon.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`pipewire_domtrans',`
+	gen_require(`
+		type pipewire_t, pipewire_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, pipewire_exec_t, pipewire_t)
+')
+
+########################################
+## <summary>
+##	Execute a domain transition to run a PipeWire client utility
+##	(e.g. pw-cat).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`pipewire_domtrans_client',`
+	gen_require(`
+		type pipewire_client_t, pipewire_client_exec_t;
+	')
+
+	corecmd_search_bin($1)
+	domtrans_pattern($1, pipewire_client_exec_t, pipewire_client_t)
+')
+
+########################################
+## <summary>
+##	Connect to the PipeWire daemon over a Unix stream socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_stream_connect',`
+	gen_require(`
+		type pipewire_t, pipewire_runtime_t;
+	')
+
+	files_search_runtime($1)
+	stream_connect_pattern($1, pipewire_runtime_t, pipewire_runtime_t, pipewire_t)
+')
+
+########################################
+## <summary>
+##	Read and write connected PipeWire daemon stream sockets.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_rw_stream_sockets',`
+	gen_require(`
+		type pipewire_t;
+	')
+
+	allow $1 pipewire_t:unix_stream_socket rw_socket_perms;
+')
+
+########################################
+## <summary>
+##	Allow a domain to use file descriptors passed from PipeWire.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_use_fds',`
+	gen_require(`
+		type pipewire_t;
+	')
+
+	allow $1 pipewire_t:fd use;
+')
+
+########################################
+## <summary>
+##	Use file descriptors inherited from PipeWire clients.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_use_client_fds',`
+	gen_require(`
+		type pipewire_client_t;
+	')
+
+	allow $1 pipewire_client_t:fd use;
+')
+
+########################################
+## <summary>
+##	Allow a domain to read, write, and map PipeWire shared memory (tmpfs) files.
+##	These are memfd files passed between the daemon and its clients for
+##	zero-copy audio buffer sharing.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_mmap_rw_tmpfs_files',`
+	gen_require(`
+		type pipewire_tmpfs_t;
+	')
+
+	fs_search_tmpfs($1)
+	mmap_rw_files_pattern($1, pipewire_tmpfs_t, pipewire_tmpfs_t)
+')
+
+########################################
+## <summary>
+##	Allow a domain to read PipeWire home directory content.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_read_home_files',`
+	gen_require(`
+		type pipewire_home_t;
+	')
+
+	userdom_search_user_home_dirs($1)
+	read_files_pattern($1, pipewire_home_t, pipewire_home_t)
+	read_lnk_files_pattern($1, pipewire_home_t, pipewire_home_t)
+')
--- a/policy/modules/apps/pipewire.if
+++ b/policy/modules/apps/pipewire.if
@@ -1,0 +1,83 @@
+## <summary>PipeWire multimedia daemon.</summary>
+
+########################################
+## <summary>
+##	Role access for PipeWire in user-service mode.
+## </summary>
+## <param name="role">
+##	<summary>
+##	Role allowed access.
+##	</summary>
+## </param>
+## <param name="domain">
+##	<summary>
+##	User domain for the role.
+##	</summary>
+## </param>
+#
+interface(`pipewire_role',`
+	gen_require(`
+		type pipewire_t, pipewire_exec_t;
+	')
+
+	role $1 types pipewire_t;
+	domtrans_pattern($2, pipewire_exec_t, pipewire_t)
+')
+
+########################################
+## <summary>
+##	Allow a domain to use file descriptors
+##	passed from PipeWire.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_use_fds',`
+	gen_require(`
+		type pipewire_t;
+	')
+
+	allow $1 pipewire_t:fd use;
+')
+
+########################################
+## <summary>
+##	Connect to PipeWire over a unix stream socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_stream_connect',`
+	gen_require(`
+		type pipewire_t, pipewire_runtime_t;
+	')
+
+	files_search_runtime($1)
+	stream_connect_pattern($1, pipewire_runtime_t, pipewire_runtime_t, pipewire_t)
+')
+
+########################################
+## <summary>
+##	Allow a domain to read and write
+##	PipeWire tmpfs shared memory files.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_rw_tmpfs_files',`
+	gen_require(`
+		type pipewire_tmpfs_t;
+	')
+
+	fs_search_tmpfs($1)
+	rw_files_pattern($1, pipewire_tmpfs_t, pipewire_tmpfs_t)
+')
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -1,0 +1,229 @@
+policy_module(pipewire, 1.0.0)
+
+########################################
+#
+# Declarations
+#
+
+
+type pipewire_t;
+type pipewire_exec_t;
+ifdef(`pipewire_system_service',`
+	init_daemon_domain(pipewire_t, pipewire_exec_t)
+',`
+	userdom_user_application_domain(pipewire_t, pipewire_exec_t)
+')
+
+type pipewire_home_t;
+userdom_user_home_content(pipewire_home_t)
+
+type pipewire_var_lib_t;
+files_type(pipewire_var_lib_t)
+
+type pipewire_runtime_t;
+files_type(pipewire_runtime_t)
+
+type pipewire_tmpfs_t;
+userdom_user_tmp_file(pipewire_tmpfs_t)
+
+########################################
+#
+# Local policy
+#
+
+allow pipewire_t self:capability sys_nice;
+allow pipewire_t self:process { getsched setsched signal sigchld };
+allow pipewire_t self:fifo_file rw_fifo_file_perms;
+allow pipewire_t self:unix_stream_socket { create_stream_socket_perms connectto };
+
+manage_dirs_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_sock_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+
+manage_files_pattern(pipewire_t, pipewire_tmpfs_t, pipewire_tmpfs_t)
+allow pipewire_t pipewire_tmpfs_t:file map;
+fs_tmpfs_filetrans(pipewire_t, pipewire_tmpfs_t, file)
+
+dev_read_sound(pipewire_t)
+dev_read_sysfs(pipewire_t)
+dev_write_sound(pipewire_t)
+
+kernel_read_system_state(pipewire_t)
+
+domain_dontaudit_read_all_domains_state(pipewire_t)
+
+files_read_etc_files(pipewire_t)
+files_read_usr_files(pipewire_t)
+ifdef(`files_map_usr_files',`
+	files_map_usr_files(pipewire_t)
+')
+
+logging_send_syslog_msg(pipewire_t)
+
+# Allow pw-play to read files from /tmp
+files_search_tmp(pipewire_t)
+files_read_all_tmp_files(pipewire_t)
+
+miscfiles_read_localization(pipewire_t)
+
+ifdef(`pipewire_system_service',`
+	allow pipewire_t self:capability dac_override;
+
+	manage_dirs_pattern(pipewire_t, pipewire_var_lib_t, pipewire_var_lib_t)
+	manage_files_pattern(pipewire_t, pipewire_var_lib_t, pipewire_var_lib_t)
+	files_var_lib_filetrans(pipewire_t, pipewire_var_lib_t, { dir file })
+
+	gen_require(`
+		type var_run_t;
+	')
+	allow pipewire_t var_run_t:dir { search add_name write remove_name };
+	filetrans_pattern(pipewire_t, var_run_t, pipewire_runtime_t, dir, "pipewire")
+
+	# Needed only when launched from init/system scripts.
+	init_rw_script_stream_sockets(pipewire_t)
+	dev_rw_dma_dev(pipewire_t)
+
+	# Allow pw-record to create/write output files in /tmp (e.g. pw-record /tmp/test.wav).
+	ifdef(`files_write_tmp_files',`
+		files_write_tmp_files(pipewire_t)
+	',`
+		gen_require(`
+			type tmp_t;
+		')
+		allow pipewire_t tmp_t:dir { search add_name write };
+		allow pipewire_t tmp_t:file { create open read write getattr setattr };
+	')
+
+	ifdef(`dbus_system_domain',`
+		optional_policy(`
+			dbus_system_domain(pipewire_t, pipewire_exec_t)
+		')
+	')
+',`
+	manage_dirs_pattern(pipewire_t, pipewire_home_t, pipewire_home_t)
+	manage_files_pattern(pipewire_t, pipewire_home_t, pipewire_home_t)
+
+	ifdef(`userdom_user_runtime_filetrans',`
+		userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, dir)
+	',`
+		gen_require(`
+			type user_tmp_t;
+		')
+		filetrans_pattern(pipewire_t, user_tmp_t, pipewire_runtime_t, dir, "pipewire")
+	')
+
+	ifdef(`userdom_user_home_dir_filetrans',`
+		userdom_user_home_dir_filetrans(pipewire_t, pipewire_home_t, dir)
+	',`
+		# Older trees fallback: keep compile compatibility; rely on FC/restorecon for home paths.
+	')
+
+	# Session/system D-Bus socket access for PipeWire helpers.
+	optional_policy(`
+		gen_require(`
+			type session_dbusd_tmp_t;
+			type staff_dbusd_t;
+		')
+		allow pipewire_t session_dbusd_tmp_t:sock_file write;
+		allow pipewire_t staff_dbusd_t:unix_stream_socket connectto;
+	')
+
+	optional_policy(`
+		gen_require(`
+			type system_dbusd_var_run_t;
+			type system_dbusd_t;
+		')
+		allow pipewire_t system_dbusd_var_run_t:sock_file write;
+		allow pipewire_t system_dbusd_t:unix_stream_socket connectto;
+	')
+
+	# Name-service and ALSA config reads observed in enforcing logs.
+	optional_policy(`
+		gen_require(`
+			type passwd_file_t;
+		')
+		allow pipewire_t passwd_file_t:file { getattr open read };
+	')
+
+	optional_policy(`
+		gen_require(`
+			type alsa_etc_rw_t;
+		')
+		allow pipewire_t alsa_etc_rw_t:file { getattr open read map };
+	')
+
+	ifdef(`dbus_all_session_bus_client',`
+		optional_policy(`
+			dbus_all_session_bus_client(pipewire_t)
+		')
+	')
+	ifdef(`dbus_connect_all_session_bus',`
+		optional_policy(`
+			dbus_connect_all_session_bus(pipewire_t)
+		')
+	')
+')
+
+ifdef(`pulseaudio_stream_connect',`
+	optional_policy(`
+		pulseaudio_stream_connect(pipewire_t)
+	')
+')
+
+ifdef(`rtkit_scheduled',`
+	optional_policy(`
+		rtkit_scheduled(pipewire_t)
+	')
+')
+
+ifdef(`systemd_read_journal_files',`
+	optional_policy(`
+		systemd_read_journal_files(pipewire_t)
+	')
+')
+
+########################################
+#
+# Unconfined access rules
+#
+
+ifdef(`pipewire_system_service',`',`
+    gen_require(`
+        type staff_t;
+        type user_t;
+        role staff_r;
+        role user_r;
+    ')
+    role staff_r types pipewire_t;
+    role user_r types pipewire_t;
+    domtrans_pattern(staff_t, pipewire_exec_t, pipewire_t)
+    domtrans_pattern(user_t, pipewire_exec_t, pipewire_t)
+    allow staff_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+    allow user_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+    allow staff_t pipewire_t:process { signal sigkill signull };
+    allow user_t pipewire_t:process { signal sigkill signull };
+    allow staff_t pipewire_t:process2 nnp_transition;
+    allow user_t pipewire_t:process2 nnp_transition;
+
+    # PipeWire stores lock/pid/socket files directly under /run/user/<uid>.
+    gen_require(`
+        type user_tmp_t;
+    ')
+    manage_files_pattern(pipewire_t, user_tmp_t, user_tmp_t)
+    manage_sock_files_pattern(pipewire_t, user_tmp_t, user_tmp_t)
+
+    # WirePlumber (left in staff_t/user_t) maps PipeWire memfd files.
+    allow staff_t pipewire_tmpfs_t:file { read write map getattr open };
+    allow user_t pipewire_tmpfs_t:file { read write map getattr open };
+
+    optional_policy(`
+        gen_require(`
+            type systemd_user_t;
+        ')
+        domtrans_pattern(systemd_user_t, pipewire_exec_t, pipewire_t)
+        allow systemd_user_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+        allow systemd_user_t pipewire_t:process { signal sigkill signull };
+        allow systemd_user_t pipewire_t:process2 nnp_transition;
+        allow systemd_user_t pipewire_tmpfs_t:file { read write map getattr open };
+    ')
+')
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -1,0 +1,214 @@
+policy_module(pipewire)
+
+########################################
+#
+# Declarations
+#
+
+## <desc>
+##	<p>
+##	Run PipeWire as a system-wide daemon launched by the systemd system
+##	instance.  When false (default), PipeWire runs as a per-user session
+##	service launched by systemd --user.
+##	Define pipewire_system_service at build time to enable system-service mode.
+##	</p>
+## </desc>
+
+attribute_role pipewire_roles;
+
+#
+# pipewire_t — the PipeWire multimedia daemon.
+#
+type pipewire_t;
+type pipewire_exec_t;
+ifdef(`pipewire_system_service',`
+	init_daemon_domain(pipewire_t, pipewire_exec_t)
+',`
+	userdom_user_application_domain(pipewire_t, pipewire_exec_t)
+	role pipewire_roles types pipewire_t;
+')
+
+#
+# pipewire_client_t — domain for pw-cat and other PipeWire client utilities.
+# pw-cat, pw-play, pw-record, etc. are standalone clients that connect to a
+# running PipeWire daemon; they are separate programs and must not run in
+# pipewire_t.
+#
+type pipewire_client_t;
+type pipewire_client_exec_t;
+userdom_user_application_domain(pipewire_client_t, pipewire_client_exec_t)
+role pipewire_roles types pipewire_client_t;
+
+type pipewire_home_t alias pipewire_var_lib_t;
+userdom_user_home_content(pipewire_home_t)
+
+type pipewire_runtime_t;
+files_runtime_file(pipewire_runtime_t)
+
+type pipewire_tmpfs_t;
+userdom_user_tmpfs_file(pipewire_tmpfs_t)
+
+########################################
+#
+# pipewire_t — common policy (both system and user instances)
+#
+
+allow pipewire_t self:capability sys_nice;
+allow pipewire_t self:process { getsched setsched signal sigchld };
+allow pipewire_t self:fifo_file rw_fifo_file_perms;
+allow pipewire_t self:unix_stream_socket { create_stream_socket_perms connectto };
+
+# Manage the daemon's own runtime socket/pid directory.
+manage_dirs_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_sock_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+
+# Shared memory (memfd) files passed to clients via file-descriptor passing
+# for zero-copy audio buffer sharing.
+mmap_manage_files_pattern(pipewire_t, pipewire_tmpfs_t, pipewire_tmpfs_t)
+fs_tmpfs_filetrans(pipewire_t, pipewire_tmpfs_t, file)
+
+# Audio hardware access.
+dev_read_sound(pipewire_t)
+dev_write_sound(pipewire_t)
+
+# Device enumeration via sysfs (discovering ALSA/USB audio devices at startup).
+dev_read_sysfs(pipewire_t)
+
+# Suppress noisy denials when PipeWire probes /proc/<pid>/status of other
+# processes during client enumeration; these are harmless stat-only calls.
+domain_dontaudit_read_all_domains_state(pipewire_t)
+
+# Read /proc state for scheduling and memory info.
+kernel_read_system_state(pipewire_t)
+
+files_read_etc_files(pipewire_t)
+files_read_usr_files(pipewire_t)
+files_map_usr_files(pipewire_t)
+
+logging_send_syslog_msg(pipewire_t)
+
+miscfiles_read_localization(pipewire_t)
+
+# NSS/name-service lookups to resolve user names for session ownership checks.
+auth_use_nsswitch(pipewire_t)
+
+# ALSA configuration files (/etc/alsa/*, ~/.asoundrc).
+optional_policy(`
+	alsa_read_config(pipewire_t)
+')
+
+# PulseAudio compatibility layer (pipewire-pulse connects to PulseAudio clients).
+optional_policy(`
+	pulseaudio_stream_connect(pipewire_t)
+')
+
+# Real-time scheduling priority via rtkit-daemon.
+optional_policy(`
+	rtkit_scheduled(pipewire_t)
+')
+
+# Read systemd journal for log correlation.
+optional_policy(`
+	systemd_read_journal_files(pipewire_t)
+')
+
+########################################
+#
+# pipewire_t — system-service-only policy
+#
+
+ifdef(`pipewire_system_service',`
+	# CAP_DAC_OVERRIDE is required to open /dev/snd/* devices that are owned
+	# by root:audio when PipeWire runs as a system-wide daemon (not as the
+	# session user).
+	allow pipewire_t self:capability dac_override;
+
+	# Persistent state directory under /var/lib/pipewire.
+	manage_dirs_pattern(pipewire_t, pipewire_var_lib_t, pipewire_var_lib_t)
+	manage_files_pattern(pipewire_t, pipewire_var_lib_t, pipewire_var_lib_t)
+	files_var_lib_filetrans(pipewire_t, pipewire_var_lib_t, { dir file })
+
+	# Runtime socket/pid directory transition under /run/pipewire.
+	files_runtime_filetrans(pipewire_t, pipewire_runtime_t, dir, "pipewire")
+
+	# DMA buffer access for zero-copy audio with hardware that supports it
+	# (e.g. /dev/dma_heap on newer kernels).
+	dev_rw_dma_dev(pipewire_t)
+
+	# System D-Bus registration (org.PipeWire.* service name on the system bus).
+	optional_policy(`
+		dbus_system_domain(pipewire_t, pipewire_exec_t)
+	')
+')
+
+########################################
+#
+# pipewire_t — user-service-only policy
+#
+
+ifndef(`pipewire_system_service',`
+	# Home directory state (~/.local/share/pipewire).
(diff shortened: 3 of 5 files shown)


policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+')
+
+########################################
+# Types & Domains
+########################################
+
+# PipeWire daemon
+type pipewire_t;
+type pipewire_exec_t;
+init_daemon_domain(pipewire_t, pipewire_exec_t)
```
pebenito: I'm a bit concerned. The common case for pipewire isn't as a system service, but as a user service.

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+# File and Directory Types
+########################################
+
+# Config file contexts
+type pipewire_etc_t;
+files_config_file(pipewire_etc_t)
+
+# Shared configuration and presets
+type pipewire_share_t;
+files_type(pipewire_share_t)
```
pebenito: Why are these needed over the general type (`usr_t`) since they are read-only to pipewire?

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+########################################
+
+# ALSA sound devices (/dev/snd/*, /dev/dsp*, etc.)
+allow pipewire_t sound_device_t:chr_file { open read write ioctl getattr };
+
+########################################
+# System Logging
+########################################
+
+systemd_read_journal_files(pipewire_t)
```
pebenito: This seems strange. How is this access used?

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+
+########################################
+# IPC and Memory Handling
+########################################
+
+# Self unix socket: create, bind, listen, accept, connect
+allow pipewire_t self:unix_stream_socket { create bind listen accept connectto read write getattr };
+
+# memfd shared memory (tmpfs-backed)
+allow pipewire_t tmpfs_t:file { read write getattr map };
```
pebenito: This will need a new `pipewire_tmpfs_t` type.

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+allow pipewire_t pulseaudio_runtime_t:dir { getattr search };
+
+# DMA device access for hardware audio offload
+allow pipewire_t dma_device_t:chr_file { read open getattr ioctl };
+
+# PulseAudio runtime dir - pid file creation
+allow pipewire_t pulseaudio_runtime_t:dir { write add_name remove_name };
+allow pipewire_t pulseaudio_runtime_t:file { create write read open getattr unlink lock };
+
+# Suppress /proc scan denials - pipewire enumerates /proc for process discovery
```
pebenito: You should use `domain_dontaudit_read_all_domains_state()` instead. Then all domains will be dontaudited instead of specific ones.

pebenito (changes requested):
Please see the [StyleGuide](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide), as there are numerous style issues. The lint check below also identified some issues.

policy/modules/services/pipewire.fc
```diff
@@ -0,0 +1,12 @@
+# --- Executables ---
+/usr/bin/pipewire                   --  gen_context(system_u:object_r:pipewire_exec_t,s0)
+
+# --- Config and data ---
+/etc/pipewire(/.*)?                     gen_context(system_u:object_r:pipewire_etc_t,s0)
+/usr/share/pipewire(/.*)?               gen_context(system_u:object_r:pipewire_share_t,s0)
+
+# --- Runtime directories ---
+/run/pipewire(/.*)?                     gen_context(system_u:object_r:pipewire_run_t,s0)
```
WavyEbuilder: Where is this coming from...? Surely this should be relative to XDG_RUNTIME_DIR no?

policy/modules/services/pipewire.fc
```diff
@@ -0,0 +1,12 @@
...
+
+# --- Persistent state ---
+/var/lib/pipewire(/.*)?                 gen_context(system_u:object_r:pipewire_var_lib_t,s0)
```
WavyEbuilder: Same here... pipewire is not a system service; it's a user service.

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+########################################
+
+allow pipewire_t self:process { fork sigchld signal getsched setsched };
+allow pipewire_t self:capability { sys_nice };
+
+########################################
+# Runtime Directory and Socket Handling
+########################################
+
+# PipeWire native runtime (/run/pipewire)
```
WavyEbuilder: as mentioned above this doesn't make sense to me.

WavyEbuilder (changes requested):

policy/modules/services/pipewire.te
WavyEbuilder: Indeed, this seems to be missing all user runtime labelling and configuration too?

WavyEbuilder:
Other things which seem to be missing:
- dbus
- rtkit
- pipewire pulse

policy/modules/services/pipewire.te
```diff
@@ -0,0 +1,152 @@
...
+files_type(pipewire_var_lib_t)
+
+########################################
+# Baseline Capabilities and Library Access
+########################################
+
+corecmd_exec_bin(pipewire_t)
+files_read_etc_files(pipewire_t)
+files_read_usr_files(pipewire_t)
+libs_use_ld_so(pipewire_t)
```
WavyEbuilder: ...? This should be implied by domain:

https://github.com/SELinuxProject/refpolicy/blob/c875d7da32bbd2418287a5eafe5c828d0091e757/policy/modules/kernel/domain.te#L149

WavyEbuilder (changes requested):

policy/modules/services/pipewire.te
rchiluka29: This policy was initially developed for a Qualcomm embedded platform where PipeWire runs as a system service (init_daemon_domain). 
Changed to userdom_user_application_domain and moved the module to policy/modules/apps/ to reflect that PipeWire runs as a user session service

policy/modules/services/pipewire.te
rchiluka29: Done.

policy/modules/services/pipewire.te
rchiluka29: done

policy/modules/services/pipewire.te
rchiluka29: Fixed.

policy/modules/services/pipewire.fc
rchiluka29: This policy was initially developed for a Qualcomm embedded platform where PipeWire runs as a system service (init_daemon_domain). 
Modified now accordingly.

policy/modules/services/pipewire.te
rchiluka29: This policy was initially developed for a Qualcomm embedded platform where PipeWire runs as a system service (init_daemon_domain). so, this line was added. 

Changed.

policy/modules/services/pipewire.fc
rchiluka29: Done.

policy/modules/services/pipewire.te
rchiluka29: modified.

policy/modules/services/pipewire.te
rchiluka29: Fixed. 
Removed libs_use_ld_so as it is already granted to all domains via domain.te line 149. Also removed libs_read_lib_files as it is covered by files_read_usr_files.

policy/modules/services/pipewire.te
pebenito: My preference for the primary focus of the policy is for it to cover the common case: a userspace service. However, I am also open to having it support system service, as long as it does not provide privileged/dangerous access when the system uses it as a user service. If that's not possible, we can also look at making the system service rules tunable.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+
+########################################
+#
+# Local policy
+#
+
+allow pipewire_t self:capability sys_nice;
+allow pipewire_t self:process { getsched setsched signal sigchld };
+allow pipewire_t self:fifo_file rw_fifo_file_perms;
+allow pipewire_t self:unix_stream_socket { accept bind connectto create getattr listen read write };
```
pebenito: Please use `create_stream_socket_perms` for a comprehensive permission set.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+
+manage_dirs_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+manage_sock_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
+files_runtime_filetrans(pipewire_t, pipewire_runtime_t, { dir file sock_file })
+
+manage_files_pattern(pipewire_t, pipewire_tmpfs_t, pipewire_tmpfs_t)
+allow pipewire_t pipewire_tmpfs_t:file map;
```
pebenito: These lines can be merged by using the `mmap_manage_files_pattern` macro.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+dev_read_sound(pipewire_t)
+dev_read_sysfs(pipewire_t)
+dev_write_sound(pipewire_t)
+
+files_read_etc_files(pipewire_t)
+files_read_usr_files(pipewire_t)
+
+kernel_read_system_state(pipewire_t)
+
+
```
pebenito, line 88: nit: Please remove this extra blank line.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+logging_send_syslog_msg(pipewire_t)
+
+miscfiles_read_localization(pipewire_t)
+
+userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, dir)
```
pebenito: This should go on line 35.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+
+domain_dontaudit_read_all_domains_state(pipewire_t)
```
pebenito: This should go between the `dev_*` and `files_*` calls (blank line above and below)

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+
+optional_policy(`
+	pulseaudio_stream_connect(pipewire_t)
+')
```
pebenito, line 104: This should go after the dbus block.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,76 @@
...
+type pipewire_exec_t;
+userdom_user_application_domain(pipewire_t, pipewire_exec_t)
+
+type pipewire_runtime_t;
+files_runtime_file(pipewire_runtime_t)
+
+type pipewire_tmpfs_t;
+userdom_user_tmpfs_file(pipewire_tmpfs_t)
+
+type pipewire_var_lib_t;
```
pebenito: Probably makes sense to rename this to pipewire_home_t (?) and alias pipewire_var_lib_t so it makes sense in system mode.

pebenito (changes requested):

policy/modules/apps/pipewire.te
rchiluka29: Done. Replaced manage_files_pattern + file map for tmpfs with mmap_manage_files_pattern.

policy/modules/apps/pipewire.te
rchiluka29: Moved

policy/modules/apps/pipewire.te
rchiluka29: Updated. Renamed to pipewire_home_t and added alias pipewire_var_lib_t. updated the file context label
    accordingly

policy/modules/apps/pipewire.te
rchiluka29: Reordered

policy/modules/apps/pipewire.te
rchiluka29: done

policy/modules/apps/pipewire.te
rchiluka29: updated. Replaced the explicit unix_stream_socket permission list with create_stream_socket_perms.

policy/modules/services/pipewire.te
rchiluka29: Agreed. This series keeps pipewire focused on the userspace service path and avoids introducing privileged system-service-only access in default policy. I’ll handle explicit system-service support separately in a follow-up, with tunable-gated rules if needed to avoid privilege leakage into user-service deployments.

policy/modules/apps/pipewire.fc
```diff
@@ -0,0 +1,6 @@
+/run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
+/run/user/%{USERID}/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_runtime_t,s0)
+
+/usr/bin/pipewire		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+
+/var/lib/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_home_t,s0)
```
thesamesam: This path does not exist for user-services use. My understanding is that the PR is scoped for user-services for now.

thesamesam (commented):
I'm a bit surprised to not see WirePlumber in here, which is needed with remotely modern PipeWire unless you are using a custom session manager. Do you have plans for that?

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,20 @@
...
+## <summary>
+##	Connect to PipeWire over a unix stream socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_stream_connect',`
```
thesamesam, line 123: Do `pw-cli` and friends need this?

```
$ equery f pipewire | grep -i bin/
/usr/bin/pipewire
/usr/bin/pipewire-aes67
/usr/bin/pipewire-avb
/usr/bin/pipewire-pulse
/usr/bin/pw-cli
/usr/bin/pw-config
/usr/bin/pw-container
/usr/bin/pw-dot
/usr/bin/pw-dump
/usr/bin/pw-jack
/usr/bin/pw-link
/usr/bin/pw-loopback
/usr/bin/pw-metadata
/usr/bin/pw-mididump
/usr/bin/pw-mon
/usr/bin/pw-profiler
/usr/bin/pw-reserve
/usr/bin/pw-top
/usr/bin/pw-v4l2
/usr/bin/spa-acp-tool
/usr/bin/spa-inspect
/usr/bin/spa-json-dump
/usr/bin/spa-monitor
```

policy/modules/apps/pipewire.fc
```diff
@@ -0,0 +1,6 @@
+/run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
```
thesamesam, line 11: I think this line is for system services.

rchiluka29:
Thanks for all the review feedback. I'll be pushing an updated version shortly that addresses all open comments.

rchiluka29:
One additional change I'd like to include: adding a pipewire_system_service tunable (defaulting to false) to support deployments where PipeWire runs as a system service rather than a user service. This is a real use-case on embedded Linux platforms, and downstream consumers of this policy have also requested that both modes be covered in a single upstream implementation rather than maintaining a separate out-of-tree patch.

This aligns with pebenito's earlier suggestion: "I am also open to having it support system service, as long as it does not provide privileged/dangerous access when the system uses it as a user service."

pebenito:
> One additional change I'd like to include: adding a pipewire_system_service tunable (defaulting to false) to support deployments where PipeWire runs as a system service rather than a user service. This is a real use-case on embedded Linux platforms, and downstream consumers of this policy have also requested that both modes be covered in a single upstream implementation rather than maintaining a separate out-of-tree patch.

Yes, this is still acceptable.

policy/modules/apps/pipewire.fc
rchiluka29: this is for system-service mode. User-service mode uses the HOME_DIR/.config/pipewire and HOME_DIR/.local/share/pipewire. will add entries above it.

policy/modules/apps/pipewire.if
rchiluka29: yes, pw-cli and friends need pipewire_stream_connect  called for their respective domains

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,99 @@
...
+files_map_usr_files(pipewire_t)
+
+logging_send_syslog_msg(pipewire_t)
+
+# Allow pw-play to read files from /tmp
+files_search_tmp(pipewire_t)
+files_read_all_tmp_files(pipewire_t)
+
+# Allow use of inherited stream sockets from init scripts (e.g. ADB shell)
+init_rw_script_stream_sockets(pipewire_t)
```
thesamesam: I'm not sure if we really want to be doing this in every single consumer (e.g. https://github.com/SELinuxProject/refpolicy/pull/1137) and whether it should be handled elsewhere, but I'm still new to refpolicy development and I defer to Chris et. al.

GargiQcom: Thanks for raising this. A few points worth noting:
 
- Most distros don't use ADB shell as a primary shell, so this is a narrow, specific case.
- refpolicy itself doesn't provide a dedicated domain for ADB shell — it falls into initrc_t by default.
- initrc_t is effectively unconfined (we confirmed this with Chris Pebenito directly: https://github.com/SELinuxProject/refpolicy/issues/1100#issuecomment-4183481277), so granting init_rw_script_stream_sockets to pipewire_t doesn't open up any meaningful new attack surface.
- This rule won't disturb anything in practice — it's a safe, no-op permission in the vast majority of deployments.
 
Given this, we believe keeping the rule as-is is fine. It follows the existing pattern used in policy modules and poses no real security concern.

pebenito:
I think I'm ok with this, but there are some lint issues. Additionally, other commenters, please mark conversations as resolved, if that's the case.

policy/modules/apps/pipewire.fc
thesamesam: This is OK now and my comment can be marked as resolved (I don't see the button).

policy/modules/apps/pipewire.if
thesamesam: Pending still AFAICT, to be clear for Chris.

policy/modules/apps/pipewire.te
thesamesam: > Most distros don't use ADB shell as a primary shell, so this is a narrow, specific case.

That's exactly why it's odd.

I'm still not sure it makes sense to put into every single policy w/ an init script but I won't argue over it, someone who knows refpol better can suggest an alternative.

WavyEbuilder: > Thanks for raising this. A few points worth noting:
> 
> * Most distros don't use ADB shell as a primary shell, so this is a narrow, specific case.

Why is it initrc_t? That just doesn't make sense.

> * refpolicy itself doesn't provide a dedicated domain for ADB shell — it falls into 
initrc_t by default.

Well this is quite clearly needed by a lot of stuff. I don't understand why not just create a private domain for the adb shell?

> * initrc_t is effectively unconfined (we confirmed this with Chris Pebenito directly: [Is unconfined_mount_t allowed to run with system role ? #1100 (comment)](https://github.com/SELinuxProject/refpolicy/issues/1100#issuecomment-4183481277)), so granting init_rw_script_stream_sockets to pipewire_t doesn't open up any meaningful new attack surface.

I'd disagree. In fact, it actually opens up oodles of attack surface because it allows talking to a highly privileged process. 

> * This rule won't disturb anything in practice — it's a safe, no-op permission in the vast majority of deployments.
> 
> Given this, we believe keeping the rule as-is is fine. It follows the existing pattern used in policy modules and poses no real security concern.

I think at the minimum this should really be a tunable.

I am not an authority here, so judgement reserved for @pebenito, but I don't think it's wise to start adding the near endless talking to initrc_t that will be necessary because adb shell seems to break the tradition model of a "shell", and probably breaks rbac _to_ it given that it's system_r. I think really there should be some kind of private domain for it.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,99 @@
...
+
+miscfiles_read_localization(pipewire_t)
+
+ifdef(`pipewire_system_service',`
```
WavyEbuilder, line 121: can pipewire run simultaneously as both a system and a user service?

WavyEbuilder (changes requested):
I think we might want $1_pipewire_t for the user services as it's a per user instance. As such, we also probably want $1_pipewire_tmpfs_t, $1_pipewire_runtime_t, etc.

This also seems to be missing role access calls for user_r/staff_r/sysadm_r.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,99 @@
...
+
+type pipewire_tmpfs_t;
+userdom_user_tmpfs_file(pipewire_tmpfs_t)
+
+########################################
+#
+# Local policy
+#
+
+allow pipewire_t self:capability { sys_nice dac_override };
```
WavyEbuilder: is dac_override required when running as a user service? If not, I think this is an AV rule that should be gated behind the `pipewire_system_service` ifdef.

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,60 @@
...
+##	</summary>
+## </param>
+#
+interface(`pipewire_rw_tmpfs_files',`
+	gen_require(`
+		type pipewire_tmpfs_t;
+		class file { read write map getattr open };
+	')
+
+	allow $1 pipewire_tmpfs_t:file { read write map getattr open };
```
WavyEbuilder: Does this need `fs_search_tmpfs`?

Also, `rw_files_pattern`?

WavyEbuilder (commented):
How was this tested? I'm not seeing RBAC for the user roles, nor am I seeing any process transitions either to actually run pipewire in the pipewire domain. I don't see a `pipewire_role` interface, nor do I see any `domtrans_pattern` either for user domains. I'm a bit concerned about whether this works correctly.

policy/modules/apps/pipewire.te
rchiluka29: No, on any given build it runs in one mode. The ifdef is a compile-time switch, same pattern as init_systemd in docker.te. Distros that need system-service mode pass -D pipewire_system_service=true at build time

rchiluka29:
> I think we might want $1_pipewire_t for the user services as it's a per user instance. As such, we also probably want $1_pipewire_tmpfs_t, $1_pipewire_runtime_t, etc.
> 
> This also seems to be missing role access calls for user_r/staff_r/sysadm_r.

added pipewire_role($1, $2) interface with roleattribute and domtrans_pattern. On the per-user type question: pulseaudio.te uses the same single-domain approach and it's the established upstream pattern.

policy/modules/apps/pipewire.te
rchiluka29: Moved dac_override inside the pipewire_system_service ifdef — it's only needed when running as a dedicated system user accessing per-user files. sys_nice stays unconditional as both modes need RT scheduling

policy/modules/apps/pipewire.if
rchiluka29: Modified. pipewire_rw_tmpfs_files now uses fs_search_tmpfs($1) + rw_files_pattern(...), matching pulseaudio_rw_tmpfs_files style. Also removed the redundant class file from gen_require (kernel classes don't need to be required — fixes the selint W-003).

rchiluka29:
> How was this tested? I'm not seeing RBAC for the user roles, nor am I seeing any process transitions either to actually run pipewire in the pipewire domain. I don't see a `pipewire_role` interface, nor do I see any `domtrans_pattern` either for user domains. I'm a bit concerned about whether this works correctly.

 Tested on Qualcomm embedded in system-service mode. For user-service mode: userdom_user_application_domain sets up the exec entrypoint. pipewire_role (now added) wires up the role side for completeness. On the per-user $1_pipewire_t question, as pebenito noted on #1138, a derived per-user domain is only needed when there are transitions out of that domain to another derived domain or back to the user domain. PipeWire doesn't have those requirements, so a single pipewire_t with UBAC separation is sufficient, same as pulseaudio.te

policy/modules/apps/pipewire.te
pebenito: > I don't think it's wise to start adding the near endless talking to initrc_t that will be necessary because adb shell seems to break the tradition model of a "shell

This is a good point; I think the access should remain a patch in the Qualcomm policy. Once we have a proper adb policy, you can drop it.

thesamesam: I suggest https://github.com/SELinuxProject/refpolicy/pull/1137 should probably be reverted then.

WavyEbuilder:
> > How was this tested? I'm not seeing RBAC for the user roles, nor am I seeing any process transitions either to actually run pipewire in the pipewire domain. I don't see a `pipewire_role` interface, nor do I see any `domtrans_pattern` either for user domains. I'm a bit concerned about whether this works correctly.
> 
> Tested on Qualcomm embedded in system-service mode.

Ok, so I'm assuming that's what has been tested.

> For user-service mode: userdom_user_application_domain sets up the exec entrypoint. 
pipewire_role (now added) wires up the role side for completeness.

Well, yes, I left that note on my review. Stating my own review comments back to me does not alleviate my primary concern.

My concern is that user services appear to be entirely untested given that this stuff is only being caught by review.

I therefore don't think it's wise to write policy to target the user domain if it's not actually being tested by Qualcomm at all (which very much appears to be the case). Else, it'll end up either as:

1. Writing user services pipewire by proxy, or
2. Based on guesses.

As a side question, are you using an LLM for replies? I'm concerned that comments are not really being taken to heart or acknowledged, and I'd rather not partake in a debate with an LLM by proxy. I am finding it concerning that seemingly LLM written replies stating review comments back are used as justification.

> On the per-user $1_pipewire_t question, as pebenito noted on #1138, a derived per-user domain is only needed when there are transitions out of that domain to another derived domain or back to the user domain. PipeWire doesn't have those requirements, so a single pipewire_t with UBAC separation is sufficient, same as pulseaudio.te

At least for the user pipewire instance of which I can speak to, pipewire clients also needs to have their own tmpfs type, which pipewire needs to be able to read and write. The Wayland policy module is a good reference for the design there in my opinion as the semantics are _fairly_ similar (though I might point out very much not a direct 1-1 clone at all; some design is still required).

I'm not seeing any interfaces to allow clients to communicate properly with pipewire. I'm not seeing pipewire-pulse support or testing. Frankly, I'm not seeing _any_ testing of reasonable quality.

I'm concerned that this did not show up in testing either. I'm not saying it _would_ require for certain at all, but my understanding of PipeWire leads me to ask, and the other omissions here make me doubt that it's been carefully checked.

If Qualcomm does not wish to support and actually test an implementation of user pipewire, that is fine. I would however wish that we would be clear about that from the get go and then see if Chris is happy to accept a patch plainly for the system service rather than doing incomplete, insufficiently tested policy for user services.

We have our own needs for a user service and I'd rather it be left for someone else to implement if you aren't able to test it properly.

pebenito:
I'm willing to accept either an incomplete user implementation or a system-only implementation since I believe there is a substantial overlap in system vs user.  However, an incomplete user implementation would delay merging until after the next refpolicy release (which is very soon).

rchiluka29:
validated in User-service mode on a Fedora vm (aarch64 kernel). 

> > > How was this tested? I'm not seeing RBAC for the user roles, nor am I seeing any process transitions either to actually run pipewire in the pipewire domain. I don't see a `pipewire_role` interface, nor do I see any `domtrans_pattern` either for user domains. I'm a bit concerned about whether this works correctly.
> > 
> > 
> > Tested on Qualcomm embedded in system-service mode.
> 
> Ok, so I'm assuming that's what has been tested.
> 
> > For user-service mode: userdom_user_application_domain sets up the exec entrypoint.
> > pipewire_role (now added) wires up the role side for completeness.
> 
> Well, yes, I left that note on my review. Stating my own review comments back to me does not alleviate my primary concern.
> 
> My concern is that user services appear to be entirely untested given that this stuff is only being caught by review.
> 
> I therefore don't think it's wise to write policy to target the user domain if it's not actually being tested by Qualcomm at all (which very much appears to be the case). Else, it'll end up either as:
> 
> 1. Writing user services pipewire by proxy, or
> 2. Based on guesses.
> 
> As a side question, are you using an LLM for replies? I'm concerned that comments are not really being taken to heart or acknowledged, and I'd rather not partake in a debate with an LLM by proxy. I am finding it concerning that seemingly LLM written replies stating review comments back are used as justification.
> 
> > On the per-user $1_pipewire_t question, as pebenito noted on #1138, a derived per-user domain is only needed when there are transitions out of that domain to another derived domain or back to the user domain. PipeWire doesn't have those requirements, so a single pipewire_t with UBAC separation is sufficient, same as pulseaudio.te
> 
> At least for the user pipewire instance of which I can speak to, pipewire clients also needs to have their own tmpfs type, which pipewire needs to be able to read and write. The Wayland policy module is a good reference for the design there in my opinion as the semantics are _fairly_ similar (though I might point out very much not a direct 1-1 clone at all; some design is still required).
> 
> I'm not seeing any interfaces to allow clients to communicate properly with pipewire. I'm not seeing pipewire-pulse support or testing. Frankly, I'm not seeing _any_ testing of reasonable quality.
> 
> I'm concerned that this did not show up in testing either. I'm not saying it _would_ require for certain at all, but my understanding of PipeWire leads me to ask, and the other omissions here make me doubt that it's been carefully checked.
> 
> If Qualcomm does not wish to support and actually test an implementation of user pipewire, that is fine. I would however wish that we would be clear about that from the get go and then see if Chris is happy to accept a patch plainly for the system service rather than doing incomplete, insufficiently tested policy for user services.
> 
> We have our own needs for a user service and I'd rather it be left for someone else to implement if you aren't able to test it properly.

I might have missed some points due to unaware or in progress.
Anyways, validated User-service mode on Fedora VM (aarch64 kernel) and system-service on qualcomm targets respectively. updated the PR.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+files_read_usr_files(pipewire_t)
+ifdef(`files_map_usr_files',`
+	files_map_usr_files(pipewire_t)
+')
+
+logging_send_syslog_msg(pipewire_t)
+
+# Allow pw-play to read files from /tmp
+files_search_tmp(pipewire_t)
+files_read_all_tmp_files(pipewire_t)
```
WavyEbuilder: This does not look right at all... could you provide an AVC for this?

I also don't think it makes sense to have pw-cat running in pipewire_t.

policy/modules/apps/pipewire.fc
```diff
@@ -0,0 +1,10 @@
+HOME_DIR/\.config/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+HOME_DIR/\.local/share/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+
+/run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
+/run/user/%{USERID}/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_runtime_t,s0)
+
+/usr/bin/pipewire		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+/usr/bin/pw-cat		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
```
WavyEbuilder: This should be a separate domain to the daemon, it's a utility?

The manual page indicates it's a client for pipewire:

> pw-cat is a simple tool for playing back or capturing raw or encoded media files on a PipeWire server.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+		allow pipewire_t system_dbusd_t:unix_stream_socket connectto;
+	')
+
+	# Name-service and ALSA config reads observed in enforcing logs.
+	optional_policy(`
+		gen_require(`
+			type passwd_file_t;
+		')
+		allow pipewire_t passwd_file_t:file { getattr open read };
+	')
```
WavyEbuilder: ... what?

If it's the name service, which you seem to have indicated in your comment, then I'm not following as to why you're manually adding passwd_file_t instead of using the various interfaces in authlogin, such as `auth_use_nsswitch`, for this?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+	optional_policy(`
+		gen_require(`
+			type system_dbusd_var_run_t;
+			type system_dbusd_t;
+		')
+		allow pipewire_t system_dbusd_var_run_t:sock_file write;
+		allow pipewire_t system_dbusd_t:unix_stream_socket connectto;
+	')
+
+	# Name-service and ALSA config reads observed in enforcing logs.
```
WavyEbuilder: Comments like this aren't really useful at all. What's more useful is explaining _why_ pipewire needs for example the name service access. The ALSA config is somewhat self explanatory, so that can be left out imo.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+			type passwd_file_t;
+		')
+		allow pipewire_t passwd_file_t:file { getattr open read };
+	')
+
+	optional_policy(`
+		gen_require(`
+			type alsa_etc_rw_t;
+		')
+		allow pipewire_t alsa_etc_rw_t:file { getattr open read map };
```
WavyEbuilder: Why not use the interface for this? `alsa_read_config`?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+	')
+')
+
+ifdef(`rtkit_scheduled',`
+	optional_policy(`
+		rtkit_scheduled(pipewire_t)
+	')
+')
+
+ifdef(`systemd_read_journal_files',`
```
WavyEbuilder: Please add a comment explaining what this is for, as this is an unusual access.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    allow user_t pipewire_t:process2 nnp_transition;
+
+    # PipeWire stores lock/pid/socket files directly under /run/user/<uid>.
+    gen_require(`
+        type user_tmp_t;
+    ')
+    manage_files_pattern(pipewire_t, user_tmp_t, user_tmp_t)
+    manage_sock_files_pattern(pipewire_t, user_tmp_t, user_tmp_t)
+
+    # WirePlumber (left in staff_t/user_t) maps PipeWire memfd files.
```
WavyEbuilder: This should be addressed in #1138 (which seems to be out of date now).

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    allow staff_t pipewire_tmpfs_t:file { read write map getattr open };
+    allow user_t pipewire_tmpfs_t:file { read write map getattr open };
```
WavyEbuilder: `mmap_rw_file_perms`?

Also, you have this gated behind a conditional for `pipewire_system_service` but this seems to pertain to the user instance?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    domtrans_pattern(staff_t, pipewire_exec_t, pipewire_t)
+    domtrans_pattern(user_t, pipewire_exec_t, pipewire_t)
+    allow staff_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+    allow user_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+    allow staff_t pipewire_t:process { signal sigkill signull };
+    allow user_t pipewire_t:process { signal sigkill signull };
+    allow staff_t pipewire_t:process2 nnp_transition;
+    allow user_t pipewire_t:process2 nnp_transition;
+
+    # PipeWire stores lock/pid/socket files directly under /run/user/<uid>.
```
WavyEbuilder: Is this supposed to be gated behind `pipewire_system_service` considering it pertains to the user instance?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    gen_require(`
+        type user_tmp_t;
```
WavyEbuilder: What is this for? Why is there not a private domain for this? And nevertheless, you should not just `gen_require` like this, you should be using the interfaces.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    gen_require(`
+        type staff_t;
+        type user_t;
+        role staff_r;
+        role user_r;
+    ')
+    role staff_r types pipewire_t;
+    role user_r types pipewire_t;
+    domtrans_pattern(staff_t, pipewire_exec_t, pipewire_t)
+    domtrans_pattern(user_t, pipewire_exec_t, pipewire_t)
```
WavyEbuilder: Please look at examples of how other user services are handled, this is not right at all. There should be an interface which is then called in the various staff/user, etc policy modules.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    allow staff_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
```
WavyEbuilder: Use an interface for this?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    allow user_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+    allow staff_t pipewire_t:process { signal sigkill signull };
+    allow user_t pipewire_t:process { signal sigkill signull };
```
WavyEbuilder: ??? This should be implicit no?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    optional_policy(`
+        gen_require(`
+            type systemd_user_t;
+        ')
+        domtrans_pattern(systemd_user_t, pipewire_exec_t, pipewire_t)
+        allow systemd_user_t pipewire_t:unix_stream_socket { create setopt bind listen connectto getopt };
+        allow systemd_user_t pipewire_t:process { signal sigkill signull };
+        allow systemd_user_t pipewire_t:process2 nnp_transition;
+        allow systemd_user_t pipewire_tmpfs_t:file { read write map getattr open };
+    ')
```
WavyEbuilder: This is also not correct. As stated above, please look at other examples in the repo of systemd user services.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+    allow staff_t pipewire_t:process { signal sigkill signull };
+    allow user_t pipewire_t:process { signal sigkill signull };
+    allow staff_t pipewire_t:process2 nnp_transition;
+    allow user_t pipewire_t:process2 nnp_transition;
+
+    # PipeWire stores lock/pid/socket files directly under /run/user/<uid>.
+    gen_require(`
+        type user_tmp_t;
+    ')
+    manage_files_pattern(pipewire_t, user_tmp_t, user_tmp_t)
```
WavyEbuilder: What is this for? This is a very broad access.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+		gen_require(`
+			type user_tmp_t;
+		')
+		filetrans_pattern(pipewire_t, user_tmp_t, pipewire_runtime_t, dir, "pipewire")
+	')
+
+	ifdef(`userdom_user_home_dir_filetrans',`
+		userdom_user_home_dir_filetrans(pipewire_t, pipewire_home_t, dir)
+	',`
+		# Older trees fallback: keep compile compatibility; rely on FC/restorecon for home paths.
```
WavyEbuilder: I'm not following? Could you make this comment more clear?

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+		allow pipewire_t staff_dbusd_t:unix_stream_socket connectto;
+	')
+
+	optional_policy(`
+		gen_require(`
+			type system_dbusd_var_run_t;
+			type system_dbusd_t;
+		')
+		allow pipewire_t system_dbusd_var_run_t:sock_file write;
+		allow pipewire_t system_dbusd_t:unix_stream_socket connectto;
```
WavyEbuilder: Please use the interfaces.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,229 @@
...
+		type var_run_t;
+	')
+	allow pipewire_t var_run_t:dir { search add_name write remove_name };
+	filetrans_pattern(pipewire_t, var_run_t, pipewire_runtime_t, dir, "pipewire")
+
+	# Needed only when launched from init/system scripts.
+	init_rw_script_stream_sockets(pipewire_t)
+	dev_rw_dma_dev(pipewire_t)
+
+	# Allow pw-record to create/write output files in /tmp (e.g. pw-record /tmp/test.wav).
```
WavyEbuilder: Please do not conflate the client and the daemon. They are separate programs.

WavyEbuilder (changes requested):
I've left some review comments, but I'm still extremely concerned about the quality of this PR and the testing going into it.

The client and daemon seem to have been conflated, and interfaces seem to be entirely ignored. Private labels are lacking throughout, and unorthodox accesses are placed without any explanation.

Adding to my concerns, it seems that the conditional for `pipewire_system_service` is wrong?

Various `gen_require` calls seem to exist within the policy module itself, including where interfaces exist, even when the appropriate places for said interfaces are _mentioned in comments above_.

After various review cycles, it is now unclear what accesses are actually needed, which is something that's almost certainly going to become much, much more confusing once the client and daemon are split into separate domains.

policy/modules/apps/pipewire.te
rchiluka29: ok, changed

policy/modules/apps/pipewire.te
rchiluka29: Done

policy/modules/apps/pipewire.te
rchiluka29: Yes. This was added incorrectly for pw-play/pw-cat which is running in pipewire_t. 
Made seperate domain pipewire_client_t which labelled as pipewire_client_exec_t in .fc. pw-cat runs in it.

policy/modules/apps/pipewire.fc
rchiluka29: Fixed. pw-cat is now labelled pipewire_client_exec_t

policy/modules/apps/pipewire.te
rchiluka29: This was added becuase userdom_user_runtime_filetrans is not available in fedora vm (validation done on its SElinux policy which was not full aligned with upstream refpolicy). But now changed to use proper runtime directory transition.

policy/modules/apps/pipewire.te
rchiluka29: By that, pipewire reads the systemd journal for log correlation. This is the same access used by pulseaudio in upstream tree.

policy/modules/apps/pipewire.te
rchiluka29: Done. Replaced with interface. gen_require block is not optimal here to use.

policy/modules/apps/pipewire.te
rchiluka29: comment updated to explain the reason. For Alsa config, removed comment as you suggested.

policy/modules/apps/pipewire.te
rchiluka29: yes, it was added from an AVC I saw during enforcing validation.  Replaced with auth_use_nsswitch interface for NSS/name-service lookups.

policy/modules/apps/pipewire.te
rchiluka29: Agreed. Removed the wireplumber block entirely. the user domain's access to pipewire shared memory now handled properly through the pipewire_role template via pipewire_rw_tmpfs_files.

policy/modules/apps/pipewire.te
rchiluka29: used mmap_rw_file_perms instead of having permissions out manually. 

Corrected. And user domain tmpfs access now present inside pipewire_role template. This is invoked for  user-service mode via  unprivuser.te and staff.te, not inline in .te.

If this is needed any change, Please let me know.

policy/modules/apps/pipewire.te
rchiluka29: yes, that block should pertained to user instance only. removed from .te completely. user domain access to pipewire is now handled through pipewire_role template in .if (for user-service) via unprivuser.te and staff.te.

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,226 @@
...
+##	Execute a domain transition to run a PipeWire client utility
+##	(e.g. pw-cat).
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed to transition.
+##	</summary>
+## </param>
+#
+interface(`pipewire_client_domtrans',`
```
pebenito: Please rename to `pipewire_domtrans_client`.

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,226 @@
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_rw_stream_sockets',`
+	gen_require(`
+		type pipewire_t;
+	')
+
+	allow $1 pipewire_t:unix_stream_socket { getattr getopt read setopt write };
```
pebenito: Please use `rw_socket_perms` for the perimssions.

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,226 @@
...
+## <summary>
+##	Use file descriptors inherited from PipeWire clients.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_client_use_fds',`
```
pebenito: `pipewire_use_client_fds`

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,226 @@
...
+## </param>
+#
+interface(`pipewire_rw_tmpfs_files',`
+	gen_require(`
+		type pipewire_tmpfs_t;
+	')
+
+	fs_search_tmpfs($1)
+	rw_files_pattern($1, pipewire_tmpfs_t, pipewire_tmpfs_t)
+	allow $1 pipewire_tmpfs_t:file map;
```
pebenito: These two lines can be merged with `mmap_rw_files_pattern`.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
+policy_module(pipewire, 2.0.0)
```
pebenito: The version parameter here is deprecated.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+userdom_user_home_content(pipewire_home_t)
+
+type pipewire_var_lib_t;
+files_type(pipewire_var_lib_t)
+
+type pipewire_runtime_t;
+files_runtime_file(pipewire_runtime_t)
+
+type pipewire_tmpfs_t;
+userdom_user_tmpfs_file(pipewire_tmpfs_t)
```
pebenito, line 49: Please move these after the `pipewire_home_t` lines above.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+
+# Device enumeration via sysfs (discovering ALSA/USB audio devices at startup).
+dev_read_sysfs(pipewire_t)
+
+# Read /proc state for scheduling and memory info.
+kernel_read_system_state(pipewire_t)
+
+# Suppress noisy denials when PipeWire probes /proc/<pid>/status of other
+# processes during client enumeration; these are harmless stat-only calls.
+domain_dontaudit_read_all_domains_state(pipewire_t)
```
pebenito, line 80: Please move above the `kernel_read_system_state(pipewire_t)` line.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+
+files_read_etc_files(pipewire_t)
+files_read_usr_files(pipewire_t)
+files_map_usr_files(pipewire_t)
+
+logging_send_syslog_msg(pipewire_t)
```
pebenito, line 89: There should be a blank line between the files and logging calls.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+')
+
+# Read systemd journal for log correlation.
+optional_policy(`
+	systemd_read_journal_files(pipewire_t)
+')
+
+# PulseAudio compatibility layer (pipewire-pulse connects to PulseAudio clients).
+optional_policy(`
+	pulseaudio_stream_connect(pipewire_t)
```
pebenito: This should be after the alsa optional.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+	optional_policy(`
+		systemd_user_daemon_domain(user, pipewire_exec_t, pipewire_t)
+		systemd_user_daemon_domain(staff, pipewire_exec_t, pipewire_t)
+	')
+
+	# Session D-Bus (register org.PipeWire.* on the per-user session bus).
+	optional_policy(`
+		dbus_all_session_bus_client(pipewire_t)
+		dbus_connect_all_session_bus(pipewire_t)
+	')
```
pebenito, line 164: This should move above the systemd optional.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+
+allow pipewire_client_t self:process { getsched signal };
+allow pipewire_client_t self:fifo_file rw_fifo_file_perms;
+allow pipewire_client_t self:unix_stream_socket { create_stream_socket_perms connectto };
+
+files_read_usr_files(pipewire_client_t)
+files_map_usr_files(pipewire_client_t)
+kernel_read_system_state(pipewire_client_t)
+miscfiles_read_localization(pipewire_client_t)
+auth_use_nsswitch(pipewire_client_t)
```
pebenito: There should be blank lines between the module group calls and auth should be the first one.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,215 @@
...
+pipewire_stream_connect(pipewire_client_t)
+pipewire_rw_stream_sockets(pipewire_client_t)
+pipewire_rw_tmpfs_files(pipewire_client_t)
+pipewire_use_fds(pipewire_client_t)
+pipewire_client_use_fds(pipewire_t)
+
+optional_policy(`
+	dbus_all_session_bus_client(pipewire_client_t)
+	dbus_connect_all_session_bus(pipewire_client_t)
+')
```
pebenito: This should be at the bottom.

pebenito (changes requested):
Thanks for all your effort on this.

Only some minor style fixes.

policy/modules/apps/pipewire.te
rchiluka29: This was added because on our earlier test system pipewire's runtime files were landing in user_tmp_t directly. Updated now.

policy/modules/apps/pipewire.te
rchiluka29: Done.

policy/modules/apps/pipewire.te
rchiluka29: Removed.

policy/modules/apps/pipewire.if
rchiluka29: Replaced those 2 lines with mmap_rw_files_pattern.

policy/modules/apps/pipewire.if
rchiluka29: Renamed from pipewire_client_use_fds to pipewire_use_client_fds

policy/modules/apps/pipewire.if
rchiluka29: Done.

policy/modules/apps/pipewire.if
rchiluka29: Renamed to pipewire_domtrans_client

policy/modules/apps/pipewire.te
rchiluka29: Moved

policy/modules/apps/pipewire.te
rchiluka29: Done

policy/modules/apps/pipewire.te
rchiluka29: Added blank lines between files and logging calls.

policy/modules/apps/pipewire.te
rchiluka29: Done

policy/modules/apps/pipewire.te
rchiluka29: Done. Added blank lines between them and moved auth to the first of above.

policy/modules/apps/pipewire.te
rchiluka29: Done.

policy/modules/apps/pipewire.te
rchiluka29: Done. removed raw allow staff_t pipewire_t:unix_stream_socket block entirely. The entire role wiring and domain transition setup has been moved into the pipewire_role template in .if, following the pulseaudio_role pattern.

The socket access between the daemon and the exec domain is handled inside the template via allow pipewire_t $3:unix_stream_socket connectto and the pipewire_rw_stream_sockets interface

policy/modules/apps/pipewire.te
rchiluka29: These have been removed. The signal permissions are now on $3 (exec domain) only inside the pipewire_role template

policy/modules/apps/pipewire.te
rchiluka29: Apologies for the confusing comment. That entire block has been removed in the updated code. We now use userdom_user_home_dir_filetrans(pipewire_t, pipewire_home_t, dir, ".local") directly without any fallback ifdef. the interface is available in upstream refpolicy and no compatibility workaround is needed.

policy/modules/apps/pipewire.te
pebenito: I realized that there are still separate types here. `pipewire_var_lib_t` is not aliased to `pipewire_home_t`.

policy/modules/apps/pipewire.te
pebenito: This change also wasn't made.

policy/modules/apps/pipewire.te
```diff
@@ -0,0 +1,218 @@
...
+		dbus_system_domain(pipewire_t, pipewire_exec_t)
+	')
+')
+
+########################################
+#
+# pipewire_t — user-service-only policy
+#
+
+ifdef(`pipewire_system_service',`',`
```
pebenito: There is a `ifndef` macro in refpolicy, so you don't need to do the empty true block trick.

policy/modules/apps/pipewire.if
```diff
@@ -0,0 +1,225 @@
...
+##	These are memfd files passed between the daemon and its clients for
+##	zero-copy audio buffer sharing.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_rw_tmpfs_files',`
```
pebenito: I missed this on my last review, but this should be named `pipewire_mmap_rw_tmpfs_files`.

pebenito (changes requested):

policy/modules/apps/pipewire.te
rchiluka29: Apologies, this was missed in the previous update. Now fixed.

policy/modules/apps/pipewire.if
rchiluka29: Done

pebenito (approved):
Thanks for all your effort on this.

PR #1110, opened 2026-05-02
newrole_t, run_init_t: call auth_run_pam()

WavyEbuilder:
newrole is linked with PAM:

rsandhu@carbon ~ $ ldd /usr/bin/newrole | ugrep 'pam'
	libpam.so.0 => /usr/lib64/libpam.so.0 (0x0000738cd6060000)
	libpam_misc.so.0 => /usr/lib64/libpam_misc.so.0 (0x0000738cd605a000)

And as can be seen from an AVC as well:

type=AVC msg=audit(1777733274.580:51): avc:  denied  { open } for  pid=1555 comm="newrole" path="/usr/share/pam/security/faillock.conf" dev="vda" ino=59482 scontext=root:staff_r:newrole_t tcontext=system_u:object_r:usr_t tclass=file permissive=1

And drop the redundant call to auth_use_nsswitch(): that is implied by auth_use_pam()[1].

[1] https://github.com/SELinuxProject/refpolicy/blob/c875d7da32bbd2418287a5eafe5c828d0091e757/policy/modules/system/authlogin.if#L59

Bug: https://bugs.gentoo.org/973557


WavyEbuilder:
Also done the same for run_init_t.

pebenito (approved):

PR #1111, opened 2026-05-02
file_contexts.subs_dist: treat /sbin and /usr/sbin the same as /usr/bin

WavyEbuilder:
Various distros, such as Gentoo, treat /bin, /sbin, and /usr/sbin the same as /usr/bin by symlinking them to there[1].

And on a Gentoo system:

$ for path in /bin /sbin /usr/sbin; do realpath "${path}"; done /usr/bin
/usr/bin
/usr/bin

As such, use subs_dist to treat them the same. Commands used to bulk modify file_contexts specifications:

$ for f in $(find policy -name '*.fc'); do sed -i 's|^/usr/sbin|/usr/bin|' "${f}"; done $ for f in $(find policy -name '*.fc'); do sed -i 's|^/sbin|/usr/bin|' "${f}"; done

[1] https://www.gentoo.org/support/news-items/2022-12-01-systemd-usrmerge.html

Bug: https://bugs.gentoo.org/960682


WavyEbuilder:
cc @perfinion @0xC0ncord

WavyEbuilder:
Ugh, there appear to be some duplicates. I'll weed them out manually.

thesamesam:
@pebenito If possible, would really love this in the next release, as it's conflict-prone and requires patching various modules if using it locally.

It fixes a source of pain with mismatched paths across distros. Fortunately many of the earlier cases were fixed for critical things already but they pop up every so often.

dburgener:
The title of this seems to have gotten mangled.  I assume it's supposed to say "/usr/bin", but it just says "/usr".

policy/modules/admin/shutdown.fc
```diff
@@ -1,16 +1,13 @@
...
 /usr/lib/upstart/shutdown	--	gen_context(system_u:object_r:shutdown_exec_t,s0)
 
-/usr/sbin/halt		--	gen_context(system_u:object_r:shutdown_exec_t,s0)
-/usr/sbin/shutdown	--	gen_context(system_u:object_r:shutdown_exec_t,s0)
+/usr/bin/halt		--	gen_context(system_u:object_r:shutdown_exec_t,s0)
 
 /run/shutdown\.pid	--	gen_context(system_u:object_r:shutdown_runtime_t,s0)
 
 ifdef(`distro_gentoo',`
-/usr/bin/halt		--	gen_context(system_u:object_r:shutdown_exec_t,s0)
```
thesamesam, line 14: This block is now empty and can be removed.

policy/modules/system/libraries.fc
```diff
@@ -15,13 +15,9 @@ ifdef(`distro_gentoo',`
 ')
 
 ifdef(`distro_redhat',`
-/emul/ia32-linux/usr(/.*)?/lib(/.*)?		gen_context(system_u:object_r:lib_t,s0)
```
thesamesam, line 18: I think you may have cleaned up duplicates here but it wasn't scoped to within an ifdef, i.e. these lines should stay I think?

thesamesam (changes requested):

policy/modules/admin/netutils.fc
```diff
@@ -12,15 +12,10 @@
...
-/usr/sbin/iptstate	--	gen_context(system_u:object_r:netutils_exec_t,s0)
-/usr/sbin/send_arp	--	gen_context(system_u:object_r:ping_exec_t,s0)
-/usr/sbin/tcpdump	--	gen_context(system_u:object_r:netutils_exec_t,s0)
-/usr/sbin/traceroute.*	--	gen_context(system_u:object_r:traceroute_exec_t,s0)
+/usr/bin/arping	--	gen_context(system_u:object_r:netutils_exec_t,s0)
+/usr/bin/fping 	--	gen_context(system_u:object_r:ping_exec_t,s0)
+/usr/bin/hping2	--	gen_context(system_u:object_r:ping_exec_t,s0)
+/usr/bin/iftop		--	gen_context(system_u:object_r:netutils_exec_t,s0)
 
 ifdef(`distro_gentoo',`
```
thesamesam: Empty block.

policy/modules/services/clamav.fc
```diff
@@ -10,8 +10,7 @@
 
 /usr/lib/systemd/system/clamd.*\.service -- gen_context(system_u:object_r:clamd_unit_t,s0)
 
-/usr/sbin/clamd	--	gen_context(system_u:object_r:clamd_exec_t,s0)
-/usr/sbin/clamav-milter	--	gen_context(system_u:object_r:clamd_exec_t,s0)
+/usr/bin/clamd	--	gen_context(system_u:object_r:clamd_exec_t,s0)
```
thesamesam: clamd is here twice but the first one has 2 tabs for the 2nd column..

policy/modules/services/smartmon.fc
```diff
@@ -2,13 +2,11 @@
...
-/usr/sbin/smartd	--	gen_context(system_u:object_r:fsdaemon_exec_t,s0)
-/usr/sbin/update-smart-drivedb	--	gen_context(system_u:object_r:smartmon_update_drivedb_exec_t,s0)
+/usr/bin/update-smart-drivedb	--	gen_context(system_u:object_r:smartmon_update_drivedb_exec_t,s0)
 
 /run/smartd\.pid	--	gen_context(system_u:object_r:fsdaemon_runtime_t,s0)
 
 /var/lib/smartmontools(/.*)?	gen_context(system_u:object_r:fsdaemon_var_lib_t,s0)
 
 ifdef(`distro_gentoo',`
-/usr/bin/update-smart-drivedb	--	gen_context(system_u:object_r:smartmon_update_drivedb_exec_t,s0)
```
thesamesam, line 13: Empty block.

policy/modules/system/authlogin.fc
```diff
@@ -28,21 +28,11 @@ ifdef(`distro_redhat', `
...
-/usr/sbin/unix_verify		--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
-/usr/sbin/utempter		--	gen_context(system_u:object_r:utempter_exec_t,s0)
-/usr/sbin/validate		--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
+/usr/bin/pwhistory_helper	--	gen_context(system_u:object_r:updpwd_exec_t,s0)
 ifdef(`distro_suse', `
-/usr/sbin/unix2_chkpwd	--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
+/usr/bin/unix2_chkpwd	--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
 ')
 ifdef(`distro_gentoo',`
-/usr/bin/pwhistory_helper	--	gen_context(system_u:object_r:updpwd_exec_t,s0)
```
thesamesam, line 45: Empty block.

policy/modules/system/authlogin.fc
```diff
@@ -28,21 +28,11 @@ ifdef(`distro_redhat', `
...
-/usr/sbin/tcb_unconvert		--	gen_context(system_u:object_r:updpwd_exec_t,s0)
-/usr/sbin/unix_chkpwd		--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
-/usr/sbin/unix_update		--	gen_context(system_u:object_r:updpwd_exec_t,s0)
-/usr/sbin/unix_verify		--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
-/usr/sbin/utempter		--	gen_context(system_u:object_r:utempter_exec_t,s0)
-/usr/sbin/validate		--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
+/usr/bin/pwhistory_helper	--	gen_context(system_u:object_r:updpwd_exec_t,s0)
 ifdef(`distro_suse', `
-/usr/sbin/unix2_chkpwd	--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
+/usr/bin/unix2_chkpwd	--	gen_context(system_u:object_r:chkpwd_exec_t,s0)
```
thesamesam, line 34: You've now made this suse-specific! We don't want that, we want it unconditionally.

thesamesam (changes requested):

policy/modules/system/authlogin.fc
thesamesam: Nevermind, I misread.

WavyEbuilder: This is `unix2_chkpwd` so I think we should be good for this one. :)

WavyEbuilder:
I've dropped the uwimap policy module as the executable entrypoint file seems to conflict with another policy module, and upstream for it seems to be long dead. Repology indicates that no distro seems to package it, with us being the only distro to appear in the list (due to the policy package for it, which is not pulled in by anything). Hopefully this should be okay?

thesamesam (approved):
I've gone over it again and it looks good to me.

pebenito:
> @pebenito If possible, would really love this in the next release, as it's conflict-prone and requires patching various modules if using it locally.
> 
> It fixes a source of pain with mismatched paths across distros. Fortunately many of the earlier cases were fixed for critical things already but they pop up every so often.

Sorry, this is too big of a change to put in at the last minute. It'll have to wait to be one of the first merges, if not the first merge, after the release.

thesamesam:
Understandable. That works for me too.

pebenito (approved):

PR #1114, opened 2026-05-12
selinux: allow seatd to use unallocated TTYs

jaihindy:
seatd requires access to unallocated TTY devices such as /dev/tty0 to manage seat activation and VT handling for Wayland compositors.

### DIFF
--- a/policy/modules/services/seatd.te
+++ b/policy/modules/services/seatd.te
@@ -32,3 +32,7 @@ auth_use_nsswitch(seatd_t)
 
 dev_rw_dri(seatd_t)
 dev_rw_input_dev(seatd_t)
+
+# seatd requires access to unallocated TTYs (e.g. /dev/tty0) to manage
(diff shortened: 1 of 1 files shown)


policy/modules/services/seatd.te
```diff
@@ -32,3 +32,6 @@ auth_use_nsswitch(seatd_t)
 
 dev_rw_dri(seatd_t)
 dev_rw_input_dev(seatd_t)
+# seatd requires access to unallocated TTYs (e.g. /dev/tty0) to manage
```
pebenito, line 36: Nit: please add an extra blank line.

pebenito (changes requested):
One trivial change

policy/modules/services/seatd.te
jaihindy: Thanks for review.
I addressed the comment , can you please review it once more.

pebenito (approved):

PR #1115, opened 2026-05-12
selinux: allow ModemManager to send DBus messages to init scripts

jaihindy:
ModemManager sends DBus method replies to init-managed services.

### DIFF
--- a/policy/modules/system/init.te
+++ b/policy/modules/system/init.te
@@ -1327,6 +1327,11 @@ optional_policy(`
 	')
 
 	optional_policy(`
+		# Needed by wireplumber to interact with ModemManager over D-Bus
+		modemmanager_dbus_chat(initrc_t)
+	')
+
+	optional_policy(`
 		policykit_dbus_chat(initrc_t)
 	')
 ')


pebenito:
I don't see any access for `initrc_t` to send to `modemmanager_t`, except for the unconfined `initrc_t` case. If services are running as `initrc_t`, then that needs to be fixed first. If init scripts or systemd units are running dbus methods in modemmanager, e.g., by `busctl` or some other similar command, then you should add a `modemmanager_dbus_chat(initrc_t)` call in `init.te`.

jaihindy:
Thanks for the review and for the clarification.
The AVCs we are seeing are actually caused by wireplumber running in initrc_t, which I agree is not ideal and should eventually run in its own domain. Below is the relevant denial:
avc:  denied  { send_msg } for msgtype=method_return
scontext=system_u:system_r:modemmanager_t:s0
tcontext=system_u:system_r:initrc_t:s0
tclass=dbus tclass=permissive=0 exe="/usr/bin/dbus-daemon" sauid=996 hostname=? addr=? terminal=?'UID="messagebus" AUID="unset" SAUID="messagebus"
type=ANOM_ABEND msg=audit(173.215:1133): auid=4294967295 uid=993 gid=989 ses=4294967295 subj=system_u:system_r:initrc_t:s0 pid=1755 comm="wireplumber" exe="/usr/bin/wireplumber" sig=11 res=1AUID="unset" UID="pipewire" GID="pipewire"

While moving wireplumber into its own domain is the correct long‑term fix, doing so right now is risky—It's upstream and we don’t yet have full visibility into its required permissions and use cases, and introducing a new domain could easily break existing functionality.
Given this, would it be acceptable to address the immediate issue by adding:
modemmanager_dbus_chat(initrc_t)
in  init.te, as you suggested? 

Thanks for review and guidance.

pebenito:
> Given this, would it be acceptable to address the immediate issue by adding:
> modemmanager_dbus_chat(initrc_t)

Yes, but please add a comment that it is for wireplumber, so we remember the reasoning.

jaihindy (commented):
Added the suggested rule with a comment indicating it is required for wireplumber D-Bus interaction with ModemManager.
can you please review it again.

policy/modules/system/init.te
```diff
@@ -966,6 +966,9 @@ userdom_read_user_home_content_files(initrc_t)
 # started from init should be placed in their own domain.
 userdom_use_inherited_user_terminals(initrc_t)
 
+# Needed by wireplumber to interact with ModemManager over D-Bus
+modemmanager_dbus_chat(initrc_t)
```
pebenito: This should move to an optional block that is nested in the dbus optional block below, like `networkmanager_dbus_chat(initrc_t)`:

https://github.com/SELinuxProject/refpolicy/blob/a160b33e0ac2762c36f0230f04bc7a8ad5f8559d/policy/modules/system/init.te#L1326

pebenito (approved):

PR #1116, opened 2026-05-13
tor

etbe:


policy/modules/services/dirmngr.te
```diff
@@ -90,3 +90,7 @@ optional_policy(`
 	gpg_secret_filetrans(dirmngr_t, dirmngr_home_t, dir)
 	gpg_stream_connect_agent(dirmngr_t)
 ')
+
+optional_policy(`
+	corenet_tcp_connect_tor_port(dirmngr_t)
```
pebenito: This needn't be optional, as it always exists.

pebenito (changes requested):

ZhanYF:
I added a few other paths to pluggable transports packaged for Debian.

Also it might be worth it to revisit the obfs4proxy policy and merge it with tor_pt. Lyrebird can already handle obfs4 and the dedicate obfs4proxy program will be deprecated.

PR #1118, opened 2026-05-13
gpg: adapt to Gentoo's app-alternatives/gpg

thesamesam:
Gentoo recently introduced app-alternatives/gpg which installs these as symlinks to a provider chosen by the user via the package manager:
* /usr/bin/gpg
* /usr/bin/gpg2

They may point to /usr/bin/gpg-reference from vanilla GnuPG or /usr/bin/gpg-freepg from FreePG [0]. We want gpg_exec_t for those.

[0] https://freepg.org/ (a fork of GnuPG)

Bug: https://bugs.gentoo.org/965532

### DIFF
--- a/policy/modules/apps/gpg.fc
+++ b/policy/modules/apps/gpg.fc
@@ -8,6 +8,10 @@ HOME_DIR/\.gnupg/S\.scdaemon		-s	gen_context(system_u:object_r:gpg_agent_tmp_t,s
 /usr/bin/gpg-agent			--	gen_context(system_u:object_r:gpg_agent_exec_t,s0)
 /usr/bin/pinentry.*			--	gen_context(system_u:object_r:gpg_pinentry_exec_t,s0)
 
+ifdef(`distro_gentoo', `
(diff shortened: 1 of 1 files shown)


thesamesam:
It's my first time really dealing with policy changes but I've checked with `restorecon` on `/usr/bin/gpg-reference` and it has the right context after this change.

pebenito:
This looks ok to me, but I'd like confirmation from @perfinion or @0xC0ncord

perfinion (approved):
Looks fine to me. Are other distros packaging alternatives too? Should these be just without the if Gentoo block?

thesamesam:
Another option I didn't handle here (yet) is `app-alternatives/gpg[sequoia]` because you can't test it yet in Gentoo, as some Portage dependencies don't work with it so far.

But Debian has https://packages.debian.org/unstable/gpg-from-sq for this, where I think it installs as `gpg -> gpg-sq`. For freepg and reference, they just call it gpg I think.

The setup this PR is fixing may exist on other distros but there's no guarantee of the same naming. I'd say better to leave it as-is for now?

pebenito (approved):

PR #1120, opened 2026-05-16
systemd: allow user systemd-tmpfiles to setfscreate

thesamesam:
systemd-tmpfiles needs to be able to setfscreate when run as a user too:
```
AVC avc:  denied  { setfscreate } for  pid=1656 comm="systemd-tmpfile"
scontext=staff_u:staff_r:staff_systemd_tmpfiles_t:s0-s0:c0.c1023
tcontext=staff_u:staff_r:staff_systemd_tmpfiles_t:s0-s0:c0.c1023
tclass=process
permissive=1
```

### DIFF
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -155,6 +155,7 @@ template(`systemd_role_template',`
 	userdom_exec_user_bin_files($1_systemd_t)
 
 	# user systemd-tmpfiles rules
+	allow $1_systemd_tmpfiles_t self:process setfscreate;
 	allow $1_systemd_tmpfiles_t $1_systemd_t:unix_stream_socket rw_socket_perms;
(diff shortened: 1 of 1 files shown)


policy/modules/system/systemd.if
```diff
@@ -156,6 +156,7 @@ template(`systemd_role_template',`
 
 	# user systemd-tmpfiles rules
 	allow $1_systemd_tmpfiles_t $1_systemd_t:unix_stream_socket rw_socket_perms;
+	allow $1_systemd_tmpfiles_t self:process setfscreate;
```
0xC0ncord, line 158: This should swap places with the line above so it's consistent with the rest of refpolicy style. Other than that LGTM.

thesamesam: Done, thanks.

pebenito (approved):

PR #1123, opened 2026-05-17
authlogin: allow pam_domain to read /usr/share/pam

thesamesam:
.. by allowing /usr overall. pam installs these configuration files there as its `vendordir`. It also does this since aea30822e2fa6f87faf2541da140f69dfdfee728.

(I think that commit message isn't quite right: it seems to do this regardless of libeconf support.)

This comes from auth_use_pam which sudo and su use:
```
AVC avc:  denied  { getattr } for  pid=239941 comm="su" path="/usr/share/pam/security/faillock.conf" dev="dm-0" ino=3227904748
scontext=staff_u:sysadm_r:sysadm_su_t:s0-s0:c0.c1023
tcontext=system_u:object_r:usr_t:s0
tclass=file

AVC avc:  denied  { read } for  pid=239938 comm="sudo" name="pam_env.conf" dev="dm-0" ino=3227906195
scontext=staff_u:staff_r:staff_sudo_t:s0-s0:c0.c1023
tcontext=system_u:object_r:usr_t:s0
tclass=file
```

This is a followup to 169f725b3ff9988de5ab9b628fc45efb8b29d0b2.

### DIFF
--- a/policy/modules/kernel/corecommands.fc
+++ b/policy/modules/kernel/corecommands.fc
@@ -355,6 +355,7 @@ ifdef(`distro_debian',`
 /usr/share/Modules/init(/.*)?		gen_context(system_u:object_r:bin_t,s0)
 /usr/share/org\.gnome\.Weather/org\.gnome\.Weather\.Application	--	gen_context(system_u:object_r:bin_t,s0)
 /usr/share/org\.gnome\.Weather/org\.gnome\.Weather\.BackgroundService	--	gen_context(system_u:object_r:bin_t,s0)
+/usr/share/pam/security/namespace\.init	--	gen_context(system_u:object_r:bin_t,s0)
 /usr/share/printconf/util/print\.py --	gen_context(system_u:object_r:bin_t,s0)
 /usr/share/PackageKit/pk-upgrade-distro\.sh -- 	gen_context(system_u:object_r:bin_t,s0)
 /usr/share/PackageKit/helpers(/.*)?	gen_context(system_u:object_r:bin_t,s0)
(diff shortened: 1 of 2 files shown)


0xC0ncord:
I also hit this issue. I don't know if this is what is intended by upstream, but the PAM ebuild is now installing files that used to be in `/etc/security` to `/usr/share/pam/security`. I think the correct fix is to use filecons rather than to allow reading `usr_t` files. We must also make sure that `/usr/share/pam/security/namespace.init` is labeled `bin_t` so that `pam_namespace` continues to work.

EDIT: A file equivalence rule will do exactly what we need.

thesamesam:
I think the change was intentional but the commit message didn't get updated. If you look at the bugs/pr the Pam commit is from, it seems they were OK with it.

Please feel free to have a go: I'm happy to try but I'm wading through a bunch of avcs and learning as I go so I don't mind if you take it over!

WavyEbuilder:
> I also hit this issue. I don't know if this is what is intended by upstream, but the PAM ebuild is now installing files that used to be in `/etc/security` to `/usr/share/pam/security`. I think the correct fix is to use filecons rather than to allow reading `usr_t` files. We must also make sure that `/usr/share/pam/security/namespace.init` is labeled `bin_t` so that `pam_namespace` continues to work.
> 
> EDIT: A file equivalence rule will do exactly what we need.

> /usr/share/pam/security/namespace.init

Oh hm, that makes it a little tricky.

> EDIT: A file equivalence rule will do exactly what we need.

subs_dist sounds good to me however:

> I think the correct fix is to use filecons rather than to allow reading usr_t files.

I still think (in the future) ideally we should move to a private type. In my opinion, given the nature of PAM configuration files, they should not be etc_t as that's a bit too broad. So a private type with `files_security_file()` in my opinion

thesamesam:
The `pam_namespace` thing makes sense, let's go with https://github.com/SELinuxProject/refpolicy/pull/1125. Thank you!

pebenito (approved):

PR #1125, opened 2026-05-17
file_contexts: make /usr/share/pam equivalent to /etc

0xC0ncord:
PAM 1.7.2 makes the vendordir build option the default. This puts PAM's config files in `/usr/share/pam` rather than `/etc`. Unfortunately this makes the default file contexts for these files `usr_t` which breaks many things.

Fix this by adding a file equivalence rule for `/usr/share/pam`.

Closes: https://github.com/SELinuxProject/refpolicy/pull/1123

### DIFF
--- a/config/file_contexts.subs_dist
+++ b/config/file_contexts.subs_dist
@@ -25,6 +25,7 @@
 /usr/local/lib32 /usr/lib
 /usr/local/lib64 /usr/lib
 /usr/local/lib /usr/lib
+/usr/share/pam /etc
 /var/lib/private /var/lib
 /var/cache/private /var/cache
 /var/log/private /var/log


pebenito:
Actually, I think we want #1123 instead. My understanding is the vendordir stuff is supposed to be the defaults, and if you want to change an option, then you place the modified configuration file in /etc. In that scenario, we would want /usr/share/pam to be read-only except to only to the package manager and sysadm.

thesamesam:
> My understanding is the vendordir stuff is supposed to be the defaults, and if you want to change an option, then you place the modified configuration file in /etc. In that scenario, we would want /usr/share/pam to be read-only except to only to the package manager and sysadm.

Yeah, that's right. More projects are moving to that because `libeconf` makes it easy to do so.

0xC0ncord:
> Actually, I think we want #1123 instead. My understanding is the vendordir stuff is supposed to be the defaults, and if you want to change an option, then you place the modified configuration file in /etc. In that scenario, we would want /usr/share/pam to be read-only except to only to the package manager and sysadm.

How should this apply for `pam_namespace`? In my testing, PAM tries to execute `/usr/share/pam/security/namespace.init` if `/etc/security/namespace.init` doesn't exist so this should be at least `bin_t` there too, unless a new type is deemed appropriate.

pebenito:
Yes, that sounds right.

thesamesam:
I've reopened https://github.com/SELinuxProject/refpolicy/pull/1123 and added the label there. I don't use `pam_namespace` so I can't test that easily.

PR #1127, opened 2026-05-18
git: add type transitions for user home content

thesamesam:
My `~/.lesshst` was getting relabeled as `user_home_dir_t` after running `git log`:

```
avc:  denied  { read } for  pid=2374509 comm="less" name=".lesshst" dev="dm-2" ino=78844139
scontext=staff_u:staff_r:staff_git_t:s0-s0:c0.c1023
tcontext=staff_u:object_r:user_home_dir_t:s0
tclass=file
```

Fix that by adding a type transition for *_git_t where we change user_home_dir_t -> user_home_t.

Note: I'm not sure which types to allow in the second argument. Various git tooling may perform a variety of actions in the user home directory. As such, I've included several types as to maintain labelling robustness. We could debate those.


pebenito:
I think this is probably ok, but I will leave this open for a while and encourage others to comment or upvote.

pebenito (approved):

PR #1128, opened 2026-05-23
auth_provide_userdb

etbe:


pebenito (changes requested):
I've been mulling about this for a while. I think it's good to have this, but since we're adding it, we should fully cover the known cases that are already handled by explicit rules. This page outlines which systemd services supply a db: https://systemd.io/USER_GROUP_API/

Furthermore, I'd like to move this to the systemd module, as this is their service. I think the interface names should be `systemd_userdb_service` and `systemd_stream_connect_userdb_services`.

thesamesam:
Yes I think I've been piecemeal noticing a bunch as well

PR #1129, opened 2026-05-23
fsadm

etbe:


pebenito:
> rw non auth dirs is for btrfs operations on dirs

Can you elaborate on this ^

etbe:
> > rw non auth dirs is for btrfs operations on dirs
> 
> Can you elaborate on this ^

To do things like create a subvol snapshot the BTRFS utility needs read/write access to the directory.

The sysadmin may create a subvol almost anywhere based on their own ideas of where data should be separated. Maybe a subvol for /etc, maybe subvol for ~/.cache on a personal workstation, maybe for ~/mysourcecode.

If you have a system run with the unconfined module and root as unconfined_t then cron jobs just work for all variants like that, but if you have root as sysadm_t then they don't work.

I want to reduce the pain of removing unconfined.

pebenito (approved):

PR #1132, opened 2026-05-23
userdomains

etbe:


dsugar100:
What is the use case where an unprivileged user needs to ptrace their own process?  I know it is wrapped in a boolean, but I would like to understand. 
I just removed ptrace - https://github.com/SELinuxProject/refpolicy/pull/1068 as a policy analysis showed too much permission.
Or is there a case where you should be making an admin user who does have ptrace permission https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/system/userdomain.if#L1374

etbe:
> What is the use case where an unprivileged user needs to ptrace their own process? I know it is wrapped in a boolean, but I would like to understand. I just removed ptrace - #1068 as a policy analysis showed too much permission. Or is there a case where you should be making an admin user who does have ptrace permission https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/system/userdomain.if#L1374

All the obvious uses of investigating problems and running a debugger when developing software.

thesamesam:
It's definitely excessive to allow unconditional ptrace but I think a tunable is reasonable. You need it for gdb or strace to be able to attach.

You may find the discussion at https://fedoraproject.org/wiki/Changes/Restrict_ptrace_for_unprivileged_users_to_child_processes_to_match_kernel_default interesting.

dsugar100:
Sure, a users does need ptrace to gdb or strace a process.  And maybe my use cases are different.  But when I have a system locked down a user debugging a process wouldn't be something that would be allowed.  I know that a boolean solves that.  Still seems lik

> > What is the use case where an unprivileged user needs to ptrace their own process? I know it is wrapped in a boolean, but I would like to understand. I just removed ptrace - #1068 as a policy analysis showed too much permission. Or is there a case where you should be making an admin user who does have ptrace permission https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/system/userdomain.if#L1374
> 
> All the obvious uses of investigating problems and running a debugger when developing software.

So you are suggesting that users on a locked system are debugging software?  That seems like it would be done on a permissive system.  Maybe my use cases for an SELinux enforced system are different than yours.  Use cases that I have would not even allow a debugger (or compiler) to be installed.  Again, I know there is the boolean around it, so it is what it is.  I guess I'm more commenting on the general use cases we (as a community) are using refpolicy for than anything else.  

Also, userdom_admin_user_template already has ptrace access.  Maybe the user you are allowing to strace and gdb should be an admin user?

etbe:
> > All the obvious uses of investigating problems and running a debugger when developing software.
> 
> So you are suggesting that users on a locked system are debugging software?

Users on a system that they don't want to be vulnerable to all the recent kernel exploits want to debug software.

> That seems like it would be done on a permissive system.

Pwning software developers is an important part of many attacks.

> Maybe my use cases for an SELinux enforced system are different than yours.

Not everything is about you.

> Also, userdom_admin_user_template already has ptrace access. Maybe the user you are allowing to strace and gdb should be an admin user?

Having non-root users debug software has been a thing for the entire history of Unix.

Having people employed as software developers who are not permitted admin access is a standard thing at almost every company with more than about 10 employees.

Having a FOSS developer on their own system at home not use an admin account for software development is hopefully becoming a standard practice now that malware in repositories is becoming common.

thesamesam:
> Sure, a users does need ptrace to gdb or strace a process. And maybe my use cases are different. But when I have a system locked down a user debugging a process wouldn't be something that would be allowed. I know that a boolean solves that.

I'm not following what your query is then..

>  I guess I'm more commenting on the general use cases we (as a community) are using refpolicy for than anything else.

refpol aims to be usable for regular use on Debian, Gentoo, etc. Gentoo is a source-based distribution. Not having a compiler available would be quite alien. But the discussion here isn't even about default behaviour as there's a tunable?

Your usecase sounds interesting but I would not consider it the norm.

WavyEbuilder:
In my opinion, this is a good patch, much appreciated @etbe.

> I know that a boolean solves that.

How so? For all intents and purposes, this is a purely a userspace construct. If we follow the line of reasoning, do we also prevent loading a module allowing ptrace on user domains as an admin may allow that locally? I don't see how a boolean adds any harm here beyond convenience for users for a quite common experience.

It's also disabled by default, so the defaults are not being changed.

> Maybe my use cases for an SELinux enforced system are different than yours.

I don't see how this would harm your usecase, considering that any domain that may reasonably set an SELinux boolean should also be able to load SELinux policy. If that's a concern, I do feel like the onus is on the person looking to lock down to take various preventative measures.

Quite literally, nothing should change for you. The next release of refpolicy should have the _exact_ same permissions with regard to ptrace for users outside of the box (or well, at least from the scope of this patch), merely with an option that now makes it more convenient for end users to configure their system to their liking.

> I guess I'm more commenting on the general use cases we (as a community) are using refpolicy for than anything else.

SELinux is gaining traction in the Gentoo side of things, and this is a feature which which is desirable for a significant chunk of users.

dsugar100:
I'm not against the merge request either, just interesting in the various use cases that refpolicy is filling.
Is it worth removing the ptrace permission from here https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/system/userdomain.if#L1374 so that the boolean fully controls ptrace for all users?

policy/modules/system/userdomain.if
```diff
@@ -2435,6 +2439,24 @@ interface(`userdom_dontaudit_append_user_home_content_files',`
...
+##	</summary>
+## </param>
+#
+interface(`userdom_rw_inherited_user_home_content_files',`
+	gen_require(`
+		type user_home_t;
+	')
+
+	allow $1 user_home_t:file rw_inherited_file_perms;
+')
```
pebenito, line 2458: Did this get in the PR by accident? I don't see a call for it.

pebenito (commented):
I'm ok with the ptrace change.

policy/modules/system/userdomain.if
etbe: This is explained in the git comment:

    Add interface userdom_rw_inherited_user_home_content_files() because it's
    really good for custom policy.  It's useful enough to go against the custom
    of only adding interfaces when the refpolicy uses them.

pebenito (approved):

PR #1139, opened 2026-05-25
authlogin: guard dbus_system_bus_client calls with optional_policy

thesamesam:
The dbus policy may not be installed (yet) when installing the systemd
module (for the first time).

Note that it is possible I believe to use systemd w/o DBus at all but
it's quite rare.

Bug: https://bugs.gentoo.org/975996
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/policy/modules/services/ssh.te
+++ b/policy/modules/services/ssh.te
@@ -286,12 +286,15 @@ ifdef(`distro_debian',`
 ')
 
 ifdef(`init_systemd',`
-	auth_use_pam_systemd(sshd_t)
-	init_dbus_chat(sshd_t)
 	# dynamic users
 	init_stream_connect(sshd_t)
 	init_rw_stream_sockets(sshd_t)
 	systemd_write_inherited_logind_sessions_pipes(sshd_t)
+
+	optional_policy(`
+		auth_use_pam_systemd(sshd_t)
+		init_dbus_chat(sshd_t)
+	')
 ')
 
 tunable_policy(`ssh_sysadm_login',`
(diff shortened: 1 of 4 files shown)


thesamesam:
I don't yet understand why the sudo commit failed CI. The first commit is the only important one though. Feel free to cherry-pick that.

pebenito:
> I don't yet understand why the sudo commit failed CI. The first commit is the only important one though. Feel free to cherry-pick that.

If init_systemd is disabled, the optional block is empty. I'd say move the rules out of the ifdef. If you have a systemd module, that's a good enough indicator to enable systemd rules.

thesamesam:
Thanks, duh..

policy/modules/system/authlogin.if
```diff
@@ -87,7 +87,9 @@ interface(`auth_use_pam',`
 ## </param>
 #
 interface(`auth_use_pam_systemd',`
-	dbus_system_bus_client($1)
+	optional_policy(`
+		dbus_system_bus_client($1)
+	')
```
pebenito: Nit: this should be at the bottom of the block.

policy/modules/services/ssh.te
```diff
@@ -286,8 +286,10 @@ ifdef(`distro_debian',`
 ')
 
 ifdef(`init_systemd',`
-	auth_use_pam_systemd(sshd_t)
-	init_dbus_chat(sshd_t)
+	optional_policy(`
+		auth_use_pam_systemd(sshd_t)
+		init_dbus_chat(sshd_t)
+	')
```
pebenito: Nit: this should be at the bottom of the block.

pebenito (changes requested):

pebenito (approved):

PR #1142, opened 2026-06-02
thunderbird

etbe:


policy/modules/apps/thunderbird.te
```diff
@@ -26,6 +26,8 @@ xdg_cache_content(thunderbird_xdg_cache_t)
 
 optional_policy(`
 	wm_application_domain(thunderbird_t, thunderbird_exec_t)
+	wm_mmap_rw_tmpfs_files(thunderbird_t)
+	wm_send_fd(thunderbird_t)
```
pebenito, line 30: Can you clarify what this is used for? It feels like it might be something that might make more sense in `wm_application_domain()`.

etbe: It could be thunderbird receiving a paste buffer from the window manager. But it seems like a fairly standard thing for the wm to receive fds from Wayland clients:

# sesearch -A -s user_wm_t -c fd
allow application_domain_type init_t:fd use;
allow nsswitch_domain nscd_t:fd use; [ nscd_use_shm ]:True
allow nsswitch_domain nscd_t:fd use; [ nscd_use_shm ]:True
allow pulseaudio_client pulseaudio_t:fd use;
allow user_application_exec_domain mozilla_t:fd use;
allow user_application_exec_domain pulseaudio_t:fd use;
allow user_application_exec_domain user_ssh_agent_t:fd use;
allow user_application_exec_domain user_su_t:fd use; [ su_allow_user_exec_domains ]:True
allow user_application_exec_domain user_sudo_t:fd use; [ sudo_allow_user_exec_domains ]:True
allow user_application_exec_domain user_wm_t:fd use;
allow user_wm_t chromium_t:fd use;
allow user_wm_t privfd:fd use;
allow user_wm_t systemd_logind_t:fd use;
allow user_wm_t user_application_exec_domain:fd use;
allow user_wm_t user_dbusd_t:fd use;
allow user_wm_t xdm_t:fd use;
allow user_wm_t xserver_t:fd use;
allow wm_domain evolution_alarm_t:fd use;
allow wm_domain thunderbird_t:fd use;
allow wm_domain xscreensaver_t:fd use;

pebenito: I'm not much of a fan of the interface name, but I also don't have a better name. It seems we can put the allow rule in `wm_application_domain()` and then we won't need the interface 😄

PR #1143, opened 2026-06-02
systemd: systemd-sysctl needs locales

thesamesam:
```
AVC avc:  denied  { search } for  pid=1117 comm="systemd-sysctl" name="locale" dev="dm-0" ino=1080894036
scontext=system_u:system_r:systemd_sysctl_t:s0
tcontext=system_u:object_r:locale_t:s0 tclass=dir
```

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -491,6 +491,8 @@ fs_getattr_cgroup(systemd_binfmt_t)
 fs_search_cgroup_dirs(systemd_binfmt_t)
 fs_getattr_nsfs_files(systemd_binfmt_t)
 
+miscfiles_read_localization(systemd_binfmt_t)
+
 ######################################
 #
 # Cgroups local policy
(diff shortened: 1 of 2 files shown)


policy/modules/system/systemd.te
```diff
@@ -2252,7 +2254,7 @@ logging_setattr_syslogd_tmp_dirs(systemd_tmpfiles_t)
 
 miscfiles_manage_man_pages(systemd_tmpfiles_t)
 miscfiles_relabel_man_cache(systemd_tmpfiles_t)
-miscfiles_getattr_localization(systemd_tmpfiles_t)
+miscfiles_read_localization(systemd_tmpfiles_t)
```
thesamesam, line 2263: The `getattr` was added in c66fefcbf157cd1f76d2af726351e480b48dece5. All of the denials in this PR went on to try to read the actual locales (sometimes a while later so maybe a different execution of the program though).

pebenito:
Looks ok to me, though I'd prefer to squash to 2 commits. One for the localization and another for the typo.

thesamesam:
Fine with me, I just discovered a bunch incrementally so committed as I went. Fixing..

PR #1146, opened 2026-06-06
mail domains

etbe:


pebenito:
This is a large change and it's taking me a while to digest it.

etbe:
> This is a large change and it's taking me a while to digest it.

How is it going?

pebenito:
> sysadm_mail_t and unconfined_mail_t compile /etc/aliases

I'm not clear why drastic changes like these when the permissions could simply be added to `admin_mail_t`?

PR #1148, opened 2026-06-06
init: adapt to OpenRC changes

thesamesam:
* Cache has moved to /var/cache/rc, per b71db4053651662a37b42f0b1273a92a870a85aa in openrc.git (note that it says /var/cache/openrc but the commit actually uses /var/cache/rc).

* keymaps are now stored in /usr/libexec/rc/console/keymap. This is since 696e684c722732f0ec0eddf7bd17cf85ca07aeaa in openrc.git.

Bug: https://bugs.gentoo.org/974290


pebenito:
> note that it says /var/cache/openrc but the commit actually uses /var/cache/rc

How confident are you that the commit message is the wrong one and not the other way around?

navi-desu:
> > note that it says /var/cache/openrc but the commit actually uses /var/cache/rc
> 
> How confident are you that the commit message is the wrong one and not the other way around?

the intention was /var/cache/rc since we use ~/.config/rc and /usr/libexec/rc (and on master, /etc/rc too)

/usr/libexec/rc has had some issues on gentoo/hurd specifically due to a conflict with hurd's upstream rc script, but we're still not set on changing it or not as it'd be quite a few paths to change (and fallbacks to include)

PR #1157, opened 2026-06-11
boot

etbe:


policy/modules/admin/bootloader.if
```diff
@@ -13,10 +13,12 @@
 interface(`bootloader_domtrans',`
 	gen_require(`
 		type bootloader_t, bootloader_exec_t;
+		type mkinitramfs_t, mkinitramfs_exec_t;
 	')
 
 	corecmd_search_bin($1)
 	domtrans_pattern($1, bootloader_exec_t, bootloader_t)
+	domtrans_pattern($1, mkinitramfs_exec_t, mkinitramfs_t)
 ')
```
pebenito, line 22: I don't think these should be overloaded like this. Additionally, I think mkinitramfs should have its own module.

policy/modules/admin/bootloader.te
```diff
@@ -8,19 +8,26 @@ policy_module(bootloader)
...
 
+ifdef(`distro_redhat',`
 #
 # boot_runtime_t is the type for /boot/kernel.h,
 # which is automatically generated at boot time.
 # only for Red Hat
 #
 type boot_runtime_t;
 files_type(boot_runtime_t)
+')
```
pebenito, line 19: I don't see this file on any of my Fedora or AlmaLinux systems, so I'd be fine with removing this, with an alias to `boot_t`.

pebenito (changes requested):

PR #1164, opened 2026-06-16
Enable policy capabilities that require libsepol 3.9, except netlink_xperm.

pebenito:
This sets the `init` initial sid to `kernel_t` to maintain current behavior with userspace_initial_context enabled.

### DIFF
--- a/policy/modules/kernel/kernel.te
+++ b/policy/modules/kernel/kernel.te
@@ -50,6 +50,7 @@ type kernel_t, can_load_kernmodule;
 domain_base_type(kernel_t)
 role system_r types kernel_t;
 sid kernel gen_context(system_u:system_r:kernel_t,mls_systemhigh)
+sid init gen_context(system_u:system_r:kernel_t,mls_systemhigh)
 
 #
 # DebugFS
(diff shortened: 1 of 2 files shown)


thesamesam:
Nit: s/capabiliies/capabilities/

WavyEbuilder:
This may be a suggestion for another patch, but I think now we have userspace_initial_context, it could very well be ideal to look to create an `earlyinit_t` or `userspace_init_t` instead of reusing `kernel_t`. I've seen quite a few early boot denials lately, for example:

```
avc:  denied  { read write } for  pid=1183 comm="systemd-udevd" path="socket:[3032]" dev="sockfs" ino=3032 scontext=system_u:system_r:udev_t:s0 tcontext=system_u:system_r:kernel_t:s0 tclass=netlink_kobject_uevent_socket permissive=1
```

This chiefly seems to be caused by the fact that systemd current and continually seems to creates more in very early boot, for example the initramfs, _before_ policy has started. Given the sensitive nature of the kernel_t domain—or to be more accurate, access control to the kernel_t domain given that kernel threads run with ring 0 privilege and therefore are high-priority targets—I think it would be desirable to use a separate domain for this.

But, I'm not too sure if this is something to be raised in this patch for a different issue as whilst the patch does introduce userspace_initial_context, I suppose it is keeping the current status quo?

pebenito:
> But, I'm not too sure if this is something to be raised in this patch for a different issue as whilst the patch does introduce userspace_initial_context, I suppose it is keeping the current status quo?

I'm trying to keep the status quo for this PR, since I'm turning on a bunch of policycaps.

PR #1166, opened 2026-06-19
wayland: Add wayland_stream_connect interface

kemalrasimsh:
Add a new interface, wayland_stream_connect(), to allow domains to connect to a Wayland compositor via an UNIX stream socket.

This interface grants the necessary permissions to search the user runtime directory and establish a stream connection to Wayland compositor sockets labeled with wayland_runtime_t.

Typical usage includes enabling confined domains (such as container runtimes or applications) to communicate with the Wayland display server.

### DIFF
--- a/policy/modules/session/wayland.fc
+++ b/policy/modules/session/wayland.fc
@@ -1,1 +1,2 @@
 /run/user/%{USERID}/wayland-.*	-s	gen_context(system_u:object_r:wayland_runtime_t,s0)
+/run/wayland-.*		-s	gen_context(system_u:object_r:wayland_runtime_t,s0)
--- a/policy/modules/session/wayland.if
+++ b/policy/modules/session/wayland.if
@@ -99,3 +99,25 @@ interface(`wayland_client_sandboxed_domain',`
 
 	typeattribute $1 wayland_client_sandboxed;
 ')
+
+#########################################
+## <summary>
+##	Connect to the Wayland compositor
+##	using a UNIX domain stream socket.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`wayland_stream_connect',`
+	gen_require(`
+		type wayland_runtime_t;
+	')
+
+	files_search_runtime($1)
+	userdom_search_user_runtime($1)
+	allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
+	allow $1 wayland_runtime_t:unix_stream_socket connectto;
+')


thesamesam:
cc @WavyEbuilder

policy/modules/session/wayland.fc
```diff
@@ -1 +1,2 @@
 /run/user/%{USERID}/wayland-.*	-s	gen_context(system_u:object_r:wayland_runtime_t,s0)
+/run/wayland-[0-9]+		-s	gen_context(system_u:object_r:wayland_runtime_t,s0)
```
WavyEbuilder: I'm guessing this is for some system wide Wayland compositor or? I'd much rather prefer you split adding this file context into a _separate_ commit with a descriptive commit message explaining what it's for.

Beyond that, I don't think `[0-9]+` is necessarily correct, namely because of the `.lock` files. I would personally suggest just using `wayland-.*` (as seen above) as that's more robust anyway, for this exact reason.

policy/modules/session/wayland.if
```diff
@@ -99,3 +99,30 @@ interface(`wayland_client_sandboxed_domain',`
...
+#
+interface(`wayland_stream_connect',`
+	gen_require(`
+		attribute wayland_compositor;
+		type wayland_runtime_t;
+	')
+
+	files_search_runtime($1)
+
+	stream_connect_pattern(
```
WavyEbuilder: I don't think the formatting for this is correct, but Chris is better suited for comment.

policy/modules/session/wayland.if
```diff
@@ -99,3 +99,30 @@ interface(`wayland_client_sandboxed_domain',`
...
+		$1,
+		wayland_runtime_t,
```
WavyEbuilder: I don't think this parameter is correct. I'm aware `search_dir_perms` is granted to `$1` for `/run`, but I don't think there are any parent container directories of the socket with the type `wayland_runtime_t`? Also, what about searching `/run/user/%{USERID}`? Considering there are "multiple" possible parent directories I'm not even sure if `stream_connect_pattern` is preferable to writing out manually like so:

```
files_search_runtime($1)
userdom_search_user_runtime($1)
allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
allow $1 wayland_runtime_t:unix_stream_socket connectto;
```

But I will defer to @pebenito for his preference here.

WavyEbuilder (changes requested):

policy/modules/session/wayland.fc
kemalrasimsh: Thanks, I’ve split this into a separate change with a proper commit message explaining the use case.

policy/modules/session/wayland.if
pebenito: Please put on one line.

policy/modules/session/wayland.if
pebenito: > ```
> files_search_runtime($1)
> userdom_search_user_runtime($1)
> allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
> allow $1 wayland_runtime_t:unix_stream_socket connectto;
> ```

Writing out the individual rules as above would be preferred if the socket is at the top level of `/run`.

kemalrasimsh: I simplified the change to follow a minimal, least-privilege approach that covers just what’s needed for our use case. Does this look better now?

pebenito: This is confusing to me, as it doesn't allow a connection to wayland.

kemalrasimsh: Yes, but it works. Would you prefer that I revert to the previous version?

`files_search_runtime($1)
userdom_search_user_runtime($1)
allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
allow $1 wayland_runtime_t:unix_stream_socket connectto;`

pebenito: It makes me concerned that this patch was not tested with the system in a clean state. For example, in permissive only, or the connection was already established when going to enforcing.

kemalrasimsh: No, I tested this on a clean setup using a freshly flashed device with SELinux always in enforcing mode. When I use Docker CDI to inject the socket, the only required permissions are:
`sock_file { getattr ioctl }`

However, if I don't use CDI and instead mount the socket directly with:
`-v /run/wayland-0:/run/wayland-0`

then an additional permission is required:
`sock_file { mounton }`

Without adding this permission, I encounter the following AVC denial:
`avc:  denied  { mounton } for  pid=3488 comm="runc:[2:INIT]" path="/var/lib/docker/rootfs/overlayfs/e37787c455093839bfa36de97a4fd5871c42d48e1044f265824e7f6275119d8a/run/wayland-0" dev="tmpfs" ino=5756 scontext=system_u:system_r:dockerd_t:s0 tcontext=system_u:object_r:wayland_runtime_t:s0 tclass=sock_file permissive=0`

There may be differences in OCI runtimes between CDI and non-CDI setups. It's also possible that another process opens the socket and provides the necessary connection permissions. Alternatively, once the socket is mounted into the container, control might effectively shift to the container_d runtime, limiting SELinux enforcement on that socket - but I'm not entirely certain on this.
In any case, I have never switched SELinux to permissive mode.

So, what would you prefer - keeping the current minimal, least‑privilege approach as is and waiting others to contribute additional permissions if needed for their use cases, or reverting to this previous more comprehensive version?
```
files_search_runtime($1)
userdom_search_user_runtime($1)
allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
allow $1 wayland_runtime_t:unix_stream_socket connectto;
```

pebenito: > So, what would you prefer - keeping the current minimal, least‑privilege approach as is and waiting others to contribute additional permissions if needed for their use cases, or reverting to this previous more comprehensive version?
> 
> ```
> files_search_runtime($1)
> userdom_search_user_runtime($1)
> allow $1 wayland_runtime_t:sock_file write_sock_file_perms;
> allow $1 wayland_runtime_t:unix_stream_socket connectto;
> ```

Yes, this please.

policy/modules/session/wayland.if
kemalrasimsh: Done.

policy/modules/session/wayland.if
kemalrasimsh: Done.

pebenito (approved):

PR #1167, opened 2026-06-19
docker: Allow dockerd to connect to pulseaudio and wayland sockets

kemalrasimsh:
Add optional policy support to allow the Docker daemon (dockerd_t) to connect to user session services over UNIX stream sockets.

This introduces:
  - pulseaudio_stream_connect(dockerd_t)
  - wayland_stream_connect(dockerd_t)

These interfaces enable dockerd to communicate with PulseAudio and Wayland compositor sockets located under /run/user/$UID, which is required for certain container workloads that need access to the host audio or graphical display.

Both permissions are wrapped in optional_policy blocks to ensure they are only enabled when the corresponding PulseAudio and Wayland policy modules are present.

This change improves compatibility with containerized desktop applications while preserving modular policy design.

### DIFF
--- a/policy/modules/services/docker.te
+++ b/policy/modules/services/docker.te
@@ -5,6 +5,15 @@ policy_module(docker)
 # Declarations
 #
 
+## <desc>
+##	<p>
+##	Determine whether the Docker daemon can connect to user
+##	session services such as PulseAudio and Wayland over
+##	UNIX stream sockets.
+##	</p>
+## </desc>
+gen_tunable(dockerd_connect_user_services, false)
+
 container_engine_domain_template(dockerd)
 container_system_engine(dockerd_t)
 optional_policy(`
@@ -75,6 +84,18 @@ ifdef(`init_systemd',`
 	init_stop_system(dockerd_t)
 	init_get_system_status(dockerd_t)
 	init_stop_generic_units(dockerd_t)
+')
+
+optional_policy(`
+	tunable_policy(`dockerd_connect_user_services',`
+		pulseaudio_stream_connect(dockerd_t)
+	')
+')
+
+optional_policy(`
+	tunable_policy(`dockerd_connect_user_services',`
+		wayland_stream_connect(dockerd_t)
+	')
 ')
 
 ########################################


thesamesam (commented):
I think these may be a tunable candidate.

ievlogie:
Why tunable? qimsdk docker always needs access to these 2 sockets

thesamesam:
I don't know what qimsdk docker is. I suggested a tunable because I don't think it's a given that people will want containers to be able to access these.

Please revisit your internal policy as to whether changes like this get upstreamed automatically if only your internal tooling needs it or with some very specific configuration. If it requires some specific use of Docker you rely on, please say that, and we can discuss it.

kemalrasimsh:
I’ve made this both optional and tunable.

policy/modules/services/docker.te
```diff
@@ -61,6 +70,18 @@ container_stream_connect_system_containers(dockerd_t)
...
+	tunable_policy(`dockerd_connect_user_services',`
+		pulseaudio_stream_connect(dockerd_t)
+	')
+')
+
+optional_policy(`
+	tunable_policy(`dockerd_connect_user_services',`
+		wayland_access_runtime_sock(dockerd_t)
+	')
+')
```
pebenito: This should move after the `init_systemd` block below

pebenito (changes requested):
The wayland interface doesn't exist.

policy/modules/services/docker.te
kemalrasimsh: Done.

kemalrasimsh:
The Wayland interface is available in the following pull request:
https://github.com/SELinuxProject/refpolicy/pull/1166
I'm unsure how to declare a dependency on this PR.

pebenito:
> The Wayland interface is available in the following pull request: #1166 I'm unsure how to declare a dependency on this PR.

You'll need to rebase this PR.

For future reference, the wayland interface commit could have been included in this PR.

aerusso:
Just saw this---I'm not sure why we would want to use `wayland_access_runtime_sock` over `wayland_client_domain`.  Generically, I would think that we should get `wayland_client` working better if for some reason it's inadequate and/or narrow  its permissions if they are too broad.

kemalrasimsh:
You're right that wayland_client_domain() should be the primary mechanism, and I'm not trying to replace it.

The issue I ran into is that wayland_client_domain() only assigns the wayland_client attribute, but it doesn't grant access to the Wayland runtime socket (wayland_runtime_t:sock_file). That seems intentional, since the policy assumes clients interact with Wayland over already-established FDs rather than directly touching the socket inode.

In the container/CDI case, though, we're bind-mounting /run/wayland-0 into the container, so the runtime (runc under dockerd_t) ends up needing filesystem-level access to the socket. That results in AVCs like { ioctl } on wayland_runtime_t:sock_file, which aren't covered by the existing client attribute rules.

So instead of broadening wayland_client_domain() (which would affect all clients), this keeps the extra permissions explicit and opt-in for domains that actually need to interact with the socket at the file level (like container runtimes).

If the preference is to improve upstream policy instead, another option would be:
- either extend wayland_client_domain() to optionally cover socket file access
- or introduce a more narrowly scoped upstream interface for runtime socket access

but given the Wayland model, I thought keeping this separate was safer and more precise.

pebenito:
kemalrasimsh commented:
> You're right that wayland_client_domain() should be the primary mechanism, and I'm not trying to replace it.
> 
> The issue I ran into is that wayland_client_domain() only assigns the wayland_client attribute, but it doesn't grant access to the Wayland runtime socket (wayland_runtime_t:sock_file). That seems intentional, since the policy assumes clients interact with Wayland over already-established FDs rather than directly touching the socket inode.
> 
> In the container/CDI case, though, we're bind-mounting /run/wayland-0 into the container, so the runtime (runc under dockerd_t) ends up needing filesystem-level access to the socket. That results in AVCs like { ioctl } on wayland_runtime_t:sock_file, which aren't covered by the existing client attribute rules.
> 
> So instead of broadening wayland_client_domain() (which would affect all clients), this keeps the extra permissions explicit and opt-in for domains that actually need to interact with the socket at the file level (like container runtimes).
> 
> If the preference is to improve upstream policy instead, another option would be:
> 
> * either extend wayland_client_domain() to optionally cover socket file access
> * or introduce a more narrowly scoped upstream interface for runtime socket access
> 
> but given the Wayland model, I thought keeping this separate was safer and more precise.

@aerusso any further response to this?

aerusso:
@pebenito I need to look at this more closely.  I had not originally appreciated the already open distinction (and I want to look at what is going on in my own policies). My first thought is we may want to have an "open new connections" and wayland_client_domain interface.  I do feel like everything that's a client should at *least* be wayland_client_domain, but again I need to think about this more carefully.

@kemalrasimsh Thanks for clarifying!

ievlogie:
> @pebenito I need to look at this more closely. I had not originally appreciated the already open distinction (and I want to look at what is going on in my own policies). My first thought is we may want to have an "open new connections" and wayland_client_domain interface. I do feel like everything that's a client should at _least_ be wayland_client_domain, but again I need to think about this more carefully.
> 
> @kemalrasimsh Thanks for clarifying!

@pebenito , @aerusso  any update? is there conclusion what needs to be final implementation?

pebenito:
> @pebenito , @aerusso any update? is there conclusion what needs to be final implementation?

I'm waiting for feedback from @aerusso .

aerusso:
Sorry for the delay!  I'm just still a bit confused.  `pulseaudio_client` has been granted `pulseaudio_stream_connect` (line 291 of pulseaudio.te).  Is the idea that `dockerd_t` needs less access than a regular client?  Is the same thing true of wayland?  It doesn't sounds like the `unix_stream_socket connectto` is a genuine additional need.

To complicated things, the wayland module, as it currents sits in main, doesn't work for me---it needs (as far as I can tell and among other things)

```
allow wayland_client wayland_runtime_t:sock_file rw_sock_file_perms
```

I haven't cleaned up that changeset, so it's not an PR yet.  With that additional change, does dockerd_t actually use less permission than a normal wayland client, or is this just a deficiency of the wayland module? Similarly, is dockerd_t really using less than a regular pipewire client?

I'm thinking dockerd_t should be a regular client for both of these (and the extra `unix_stream_socket connectto` is not necessary, right?).

pebenito:
> I'm thinking dockerd_t should be a regular client for both of these (and the extra `unix_stream_socket connectto` is not necessary, right?).

I would tend to agree with that, as it makes the intent clearer. That is, as long as the client interfaces don't give worrying extra unnecessary access that wouldn't be used here.

kemalrasimsh:
@aerusso, `pulseaudio_client` is already granted `pulseaudio_stream_connect` which expands to:
```
interface(`pulseaudio_stream_connect',`
	gen_require(`
		type pulseaudio_t, pulseaudio_runtime_t, pulseaudio_tmp_t;
	')

	files_search_runtime($1)
	stream_connect_pattern($1, { pulseaudio_tmp_t pulseaudio_runtime_t }, { pulseaudio_tmp_t pulseaudio_runtime_t }, pulseaudio_t)
')
```
and `stream_connect_pattern` is defined as:
```
define(`stream_connect_pattern',`
	allow $1 $2:dir search_dir_perms;
	allow $1 $3:sock_file write_sock_file_perms;
	allow $1 $4:unix_stream_socket connectto;
')
```
As a result, clients receive `write_sock_file_perms ({ getattr write open append })` on the relevant `sock_file`.
Based on that, I would expect `dockerd_t` to have the same access pattern as a regular client.
With the Wayland module as it currently exists in main, I can get it working by simply adding:
`allow dockerd_t wayland_runtime_t:sock_file { getattr };`
From what I can see, there are three possible approaches:

1. Treat dockerd_t as a Wayland compositor domain
  Add:
  `wayland_compositor_domain(dockerd_t)` to docker.te.
  This macro already grants:
  `allow wayland_compositor wayland_runtime_t:sock_file manage_sock_file_perms;`

2. Extend `wayland_client_sandboxed`
  Add the following to wayland.te:
  ```
# Local Policy
  allow wayland_compositor wayland_client_sandboxed:sock_file getattr;
```
and then use: `wayland_client_sandboxed(dockerd_t)` in docker.te.

3. Reduce permissions in wayland_stream_connect
  Remove `allow $1 wayland_runtime_t:unix_stream_socket connectto;` and instead of:
  `allow $1 wayland_runtime_t:sock_file write_sock_file_perms;`
  change the socket file permission to:
  `allow $1 wayland_runtime_t:sock_file getattr_sock_file_perms;`
  while keeping the rest of `wayland_stream_connect(dockerd_t)` unchanged.
  
  
  Do you have a preference on which direction we should take?
  Also, @aerusso, are you planning to clean up and push the Wayland-related PR you mentioned earlier?

aerusso:
I think I have some misunderstanding about the way that docker is working---I would have thought that full wayland/pulseaudio client permission is needed for a docker-contained process to be a wayland/pulseaudio client.  Given how limited these additional permissions are, and how immature the rest of the wayland confinement is, I think it makes sense to just merge this, and fix any problems that develop as the Wayland confinement develops.

pebenito (approved):

PR #1169, opened 2026-06-22
qemu

etbe:


policy/modules/services/virt.te
```diff
@@ -1136,13 +1136,24 @@ allow virt_bridgehelper_t virt_etc_t:file read_file_perms;
...
 corenet_rw_tun_tap_dev(virt_bridgehelper_t)
 
+domain_use_interactive_fds(virt_bridgehelper_t)
+
 userdom_search_user_home_dirs(virt_bridgehelper_t)
 userdom_use_user_ptys(virt_bridgehelper_t)
 
+optional_policy(`
+	qemu_domain(virt_bridgehelper_t)
+')
```
pebenito, line 1155: This should instead be a run interface provided by the virt module.

policy/modules/services/virt.te
```diff
@@ -1136,13 +1136,24 @@ allow virt_bridgehelper_t virt_etc_t:file read_file_perms;
 
 manage_files_pattern(virt_bridgehelper_t, svirt_home_t, svirt_home_t)
 
+kernel_getattr_proc(virt_bridgehelper_t)
+kernel_read_kernel_sysctls(virt_bridgehelper_t)
 kernel_read_network_state(virt_bridgehelper_t)
+kernel_read_system_state(virt_bridgehelper_t)
+
+dev_read_sysfs(virt_bridgehelper_t)
```
pebenito, line 1144: What entries are read? Please add a comment in the policy with that info.

policy/modules/apps/qemu.te
```diff
@@ -43,9 +43,15 @@ init_unit_file(qemu_unit_t)
 
 dev_read_sysfs(qemu_t)
 
+allow qemu_t self:anon_inode { create map read write };
```
pebenito, line 46: Probably can fully upgrade to `mmap_manage_file_perms`.

policy/modules/roles/sysadm.te
```diff
@@ -903,6 +903,10 @@ optional_policy(`
 	pyzor_role(sysadm, sysadm_t, sysadm_application_exec_domain, sysadm_r)
 ')
 
+optional_policy(`
+	qemu_role(sysadm_r, sysadm_t)
+')
```
pebenito, line 908: What's the use case here?

pebenito (changes requested):

policy/modules/apps/qemu.te
kcinimod-defensec, line 46: I would not use "self" for anon_inode because then you lose control over the various kinds like uffd, io_uring, perf_event, secretmem etc.

In the case of qemu you would allow excessive permissions because afaik qemu only leverages io_uring and uffd anon_inodes

Here is how i deal with anon_inodes:

https://git.defensec.nl/forge/defensec/dssp5/src/branch/dssp5-debian/src/anoninode.cil
https://git.defensec.nl/forge/defensec/dssp5/src/branch/dssp5-debian/src/anoninode

https://git.defensec.nl/forge/defensec/dssp5/src/branch/dssp5-debian/src/anoninode/iouringanoninode.cil#L32

kcinimod-defensec:
```
root@nimbus:~# sesearch -A -s qemu.guest.subj -c anon_inode
allow qemu.guest.subj qemu.guest.iouring.anon_inode:anon_inode { append create getattr ioctl link lock map open read rename setattr unlink write };
allow qemu.guest.subj qemu.guest.uffd.anon_inode:anon_inode { append create getattr ioctl link lock open read rename setattr unlink write };
```

policy/modules/services/virt.te
kcinimod-defensec: I don't think the role matters. If a role is not authorized to associate with qemu_t then it should also not be authorized to associate with virt_bridgehelper_t because let's face it: it only makes sense for qemu to ever run that helper.

pebenito: I'm not sure I understand this comment. It sounds like you're agreeing with me?. The point of the run interface is to pass the qemu roles down to virt_bridgehelper_t.

policy/modules/apps/qemu.te
kcinimod-defensec:
```
root@nimbus:~# sesearch -T -s qemu.guest.subj -c anon_inode
type_transition qemu.guest.subj qemu.guest.subj:anon_inode qemu.guest.iouring.anon_inode [io_uring];
type_transition qemu.guest.subj qemu.guest.subj:anon_inode qemu.guest.uffd.anon_inode [userfaultfd];
```

policy/modules/services/virt.te
kcinimod-defensec: > I'm not sure I understand this comment. It sounds like you're agreeing with me?. The point of the run interface is to pass the qemu roles down to virt_bridgehelper_t.

Yes. Sorry.

PR #1170, opened 2026-06-23
tpm2: fix tpm2-abrmd D-Bus communication under enforcing mode

khalidfaisal17:
tpm2-abrmd was crashing on every D-Bus activation attempt under SELinux enforcing mode, causing all tpm2-tools commands to fall back to direct TPM device access with a GDBus warning.

AVC log confirmed the root denial:

```
avc: denied { read write } for comm="dbus-daemon"
scontext=system_u:system_r:system_dbusd_t:s0
tcontext=system_u:system_r:tpm2_abrmd_t:s0
tclass=unix_stream_socket permissive=0
```

tpm2-abrmd uses socketpair(PF_LOCAL, SOCK_STREAM) to create a connected fd pair per client, then passes one end to dbus-daemon via SCM_RIGHTS (D-Bus fd-passing). dbus-daemon needs { read write } on that inherited anonymous socket to relay TPM commands. This is not a connectto — the socket is already connected when the fd is received.

**Files changed:**

- `policy/modules/services/tpm2.if` — add tpm2_rw_abrmd_stream_sockets() granting rw_socket_perms on tpm2_abrmd_t:unix_stream_socket
- `policy/modules/services/dbus.te` — call tpm2_rw_abrmd_stream_sockets(system_dbusd_t) inside the existing optional_policy block alongside tpm2_rw_abrmd_pipes()
- `policy/modules/services/tpm2.te` — add files_read_etc_files (D-Bus config), miscfiles_read_localization (GLib init), and init_dbus_chat (systemd Type=dbus activation) for tpm2_abrmd_t
- `policy/modules/system/unconfined.te` — add tpm2_dbus_chat_abrmd(unconfined_t) in an optional_policy block for interactive root sessions

### DIFF
--- a/policy/modules/roles/sysadm.te
+++ b/policy/modules/roles/sysadm.te
@@ -1148,6 +1148,11 @@ optional_policy(`
 ')
 
 optional_policy(`
+	tpm2_run(sysadm_t, sysadm_r)
+	tpm2_dbus_chat_abrmd(sysadm_t)
+')
+
+optional_policy(`
 	transproxy_admin(sysadm_t, sysadm_r)
 ')
 
--- a/policy/modules/services/dbus.te
+++ b/policy/modules/services/dbus.te
@@ -305,6 +305,7 @@ optional_policy(`
 
 optional_policy(`
 	tpm2_rw_abrmd_pipes(system_dbusd_t)
+	tpm2_rw_abrmd_stream_sockets(system_dbusd_t)
 ')
 
 optional_policy(`
--- a/policy/modules/services/tpm2.if
+++ b/policy/modules/services/tpm2.if
@@ -224,3 +224,23 @@ interface(`tpm2_rw_abrmd_pipes',`
 	allow $1 tpm2_abrmd_t:fd use;
 	allow $1 tpm2_abrmd_t:fifo_file rw_fifo_file_perms;
 ')
+
+########################################
+## <summary>
+##	Read and write tpm2-abrmd unix stream sockets.
+##	This is used by the D-Bus system daemon to communicate
+##	over the socketpair fds passed by tpm2-abrmd to clients.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`tpm2_rw_abrmd_stream_sockets',`
+	gen_require(`
+		type tpm2_abrmd_t;
+	')
+	allow $1 tpm2_abrmd_t:fd use;
+	allow $1 tpm2_abrmd_t:unix_stream_socket rw_socket_perms;
+')
--- a/policy/modules/services/tpm2.te
+++ b/policy/modules/services/tpm2.te
@@ -33,8 +33,13 @@ logging_send_syslog_msg(tpm2_abrmd_t)
 
 dev_read_urand(tpm2_abrmd_t)
 
+files_read_etc_files(tpm2_abrmd_t)
+
+miscfiles_read_localization(tpm2_abrmd_t)
+
 optional_policy(`
 	dbus_system_domain(tpm2_abrmd_t, tpm2_abrmd_exec_t)
+	init_dbus_chat(tpm2_abrmd_t)
 ')
 
 
--- a/policy/modules/system/unconfined.te
+++ b/policy/modules/system/unconfined.te
@@ -275,6 +275,10 @@ allow unconfined_execmem_t self:process { execmem execstack };
 unconfined_domain_noaudit(unconfined_execmem_t)
 
 optional_policy(`
+	tpm2_dbus_chat_abrmd(unconfined_t)
+')
+
+optional_policy(`
 	unconfined_dbus_chat(unconfined_execmem_t)
 ')
 


khalidfaisal17:
Hi Reviewers,

Can you please approve the workflow checks so that this change go through it.

policy/modules/services/dbus.te
```diff
@@ -422,3 +422,4 @@ optional_policy(`
 
 allow dbusd_unconfined { dbusd_session_bus_client dbusd_system_bus_client }:dbus send_msg;
 allow dbusd_unconfined { system_dbusd_t session_bus_type }:dbus all_dbus_perms;
+tpm2_dbus_stream_connect(system_dbusd_t)
```
pebenito: This seems strange to me. Why would dbusd be doing the connection instead of the reverse direction?

policy/modules/system/init.te
```diff
@@ -403,6 +403,7 @@ ifdef(`init_systemd',`
 	dev_manage_sysfs_dirs(init_t)
 	dev_relabel_sysfs_dirs(init_t)
 	dev_read_usbfs(initrc_t)
+	tpm2_dbus_chat_abrmd(initrc_t)
```
pebenito: What processes specifically are doing this access? If this is daemons that don't have their own domain, then it's not appropriate.

policy/modules/system/unconfined.te
```diff
@@ -277,6 +277,7 @@ unconfined_domain_noaudit(unconfined_execmem_t)
 optional_policy(`
 	unconfined_dbus_chat(unconfined_execmem_t)
 ')
+tpm2_dbus_chat_abrmd(unconfined_t)
```
pebenito: This should be in an optional block.

pebenito (changes requested):
cc @dsugar100 for additional thoughts.

policy/modules/services/dbus.te
khalidfaisal17: You are right that the name and permission were wrong in the original patch. The original tpm2_dbus_stream_connect used rw_stream_socket_perms which implies a connectto style operation — but that is not what is happening here.

The AVC log from the device confirmed the exact denial:

```
avc: denied { read write } for comm="dbus-daemon"
scontext=system_u:system_r:system_dbusd_t:s0
tcontext=system_u:system_r:tpm2_abrmd_t:s0
tclass=unix_stream_socket permissive=0
```

tpm2-abrmd uses socketpair(PF_LOCAL, SOCK_STREAM) to create a pre-connected anonymous fd pair for each client. It keeps one end and passes the other to dbus-daemon via SCM_RIGHTS (D-Bus fd-passing). dbus-daemon is not making a new connection — it is receiving an already-connected socket fd and needs { read write } on it to relay TPM commands between the client and tpm2-abrmd.

The updated patch replaces the original interface with tpm2_rw_abrmd_stream_sockets() which grants exactly rw_socket_perms on tpm2_abrmd_t:unix_stream_socket — no connectto, no listen/accept. The interface name now accurately describes what is happening: dbus-daemon is reading and writing an abrmd-owned socket fd it received, not initiating a connection.

policy/modules/system/init.te
khalidfaisal17: You are correct — initrc_t was the wrong domain. That line has been removed entirely from init.te.

The process that actually needs D-Bus chat with tpm2-abrmd is systemd itself (init_t), not SysV init scripts (initrc_t). tpm2-abrmd is a Type=dbus systemd service. When systemd activates it, init_t needs to exchange D-Bus messages with tpm2_abrmd_t to monitor readiness (waiting for the D-Bus name com.intel.tss2.Tabrmd to be acquired before marking the service as started).

The correct fix is init_dbus_chat(tpm2_abrmd_t) added inside the dbus_system_domain optional block in tpm2.te -consistent with how other Type=dbus daemons like NetworkManager and devicekit handle this. This is a targeted, per-daemon permission, not a broad grant to all init scripts.

policy/modules/system/unconfined.te
khalidfaisal17: Already addressed. The change is wrapped in an optional_policy block

policy/modules/services/tpm2.te
```diff
@@ -33,8 +33,16 @@ logging_send_syslog_msg(tpm2_abrmd_t)
...
 
+files_read_etc_files(tpm2_abrmd_t)
+
+miscfiles_read_localization(tpm2_abrmd_t)
+
 optional_policy(`
 	dbus_system_domain(tpm2_abrmd_t, tpm2_abrmd_exec_t)
+
+	optional_policy(`
+		init_dbus_chat(tpm2_abrmd_t)
```
pebenito: This is already unconditionally required by the module (lines 10 and 13), so this extra optional isn't necessary. It should remain in the dbus optional.

policy/modules/system/unconfined.te
```diff
@@ -278,6 +278,10 @@ optional_policy(`
 	unconfined_dbus_chat(unconfined_execmem_t)
 ')
 
+optional_policy(`
+	tpm2_dbus_chat_abrmd(unconfined_t)
+')
```
pebenito, line 283: This should go above the unconfined_dbus_chat() call.

pebenito (changes requested):

policy/modules/system/unconfined.te
khalidfaisal17: Thanks for review;
updated the code as suggested

policy/modules/services/tpm2.te
khalidfaisal17: Resolved in latest commit; Thanks

policy/modules/roles/sysadm.te
```diff
@@ -1131,6 +1131,11 @@ optional_policy(`
 	tcsd_admin(sysadm_t, sysadm_r)
 ')
 
+optional_policy(`
+	tpm2_run(sysadm_t, sysadm_r)
+	tpm2_dbus_chat_abrmd(sysadm_t)
+')
```
pebenito: This block should move after the `tor_admin` optional block.

policy/modules/services/tpm2.if
```diff
@@ -224,3 +224,23 @@ interface(`tpm2_rw_abrmd_pipes',`
...
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`tpm2_rw_abrmd_stream_sockets',`
+	gen_require(`
+		type tpm2_abrmd_t;
+	')
+
```
pebenito: Please add `allow $1 tpm2_abrmd_t:fd use;`.

pebenito (changes requested):

policy/modules/services/tpm2.if
khalidfaisal17: updated, Thanks

policy/modules/roles/sysadm.te
khalidfaisal17: Thanks for input. Updated;
Thanks

khalidfaisal17:
Hi Reviewers;

Kindly approve the changes; Its already validated.

policy/modules/system/init.te
```diff
@@ -1334,6 +1334,10 @@ optional_policy(`
 	optional_policy(`
 		policykit_dbus_chat(initrc_t)
 	')
+
+	optional_policy(`
+		tpm2_dbus_chat_abrmd(initrc_t)
```
pebenito: This was supposed to be removed, it has reappeared. See previous discussion in init.te.

khalidfaisal17: Re-added with justification. On this platform adbd (Android Debug Bridge daemon) runs as initrc_t, confirmed on device:

```
# ps -eZ | grep adbd
system_u:system_r:initrc_t:s0  1649  ?  00:00:00 adbd

```
Without this, tpm2_abrmd_t cannot send the D-Bus reply back to initrc_t — the reply is silently dropped and the caller times out. No AVC is logged for the dropped reply, making it non-obvious.

The placement follows the existing pattern in this block (firewalld, networkmanager, modemmanager, policykit) and the nested optional_policy ensures it only applies when the tpm2 module is loaded.

pebenito: This was discussed in other PRs, but we are not handling this in refpolicy. The rule will need to remain in you downstream policy until an adbd domain is added here.

khalidfaisal17: Updated the patch as discussed here!!
Kindly review and approve!

policy/modules/system/init.te
```diff
@@ -1334,6 +1334,7 @@ optional_policy(`
 	optional_policy(`
 		policykit_dbus_chat(initrc_t)
 	')
+
```
pebenito: nit: please remove this residual whitespace change

khalidfaisal17: Done!

pebenito (approved):

PR #1171, opened 2026-06-29
Pipewire/wireplumber

aerusso:
The new pipewire module needs some changes to work on the desktop.  I've also included @tkanfade's new wireplumber module, and changes atop it that I needed to get everything working.

I'm happy to divide these changes however makes everyone happy, but the current state of refpolicy main is nonfunctional.

There are also some comments about consolidating al of pulseaudio, pipewire, alsa, et. al.  I take no position on that issue beyond also including wireplumber in that discussion.

### DIFF
--- a/policy/modules/apps/pipewire.fc
+++ b/policy/modules/apps/pipewire.fc
@@ -3,10 +3,11 @@
 
 # System-service runtime directory (/run/pipewire).
 /run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
-/run/user/%{USERID}/pipewire-.*	-s	gen_context(system_u:object_r:pipewire_runtime_t,s0)
 
 # User-session runtime sockets live under $XDG_RUNTIME_DIR/pipewire and are
-# labelled at runtime via userdom_user_runtime_filetrans; no static entry needed.
+# labelled at runtime via userdom_user_runtime_filetrans
+# Static entry added for restorecon
+/run/user/%{USERID}/pipewire-.*		gen_context(system_u:object_r:pipewire_runtime_t,s0)
 
 # Persistent system-service state.
 /var/lib/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_var_lib_t,s0)
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -197,6 +197,6 @@ ifndef(`pipewire_system_service',`
 allow pipewire_client_domain pipewire_runtime_t:dir list_dir_perms;
 pipewire_stream_connect(pipewire_client_domain)
 pipewire_use_fds(pipewire_client_domain)
-pipewire_mmap_rw_tmpfs_files(pipewire_client_domain)
+pipewire_mmap_rw_inherited_tmpfs_files(pipewire_client_domain)
 pipewire_rw_stream_sockets(pipewire_client_domain)
 pipewire_dbus_chat(pipewire_client_domain)
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -33,6 +33,10 @@ type pipewire_runtime_t;
 files_runtime_file(pipewire_runtime_t)
 userdom_user_runtime_content(pipewire_runtime_t)
 
+optional_policy(`
+	systemd_user_runtime_dir_unlink(pipewire_runtime_t, sock_file)
+')
+
 type pipewire_tmpfs_t;
 userdom_user_tmpfs_file(pipewire_tmpfs_t)
 
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -153,6 +153,7 @@ ifndef(`pipewire_system_service',`
 	# Runtime socket directory transition under $XDG_RUNTIME_DIR/pipewire.
 	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, dir, "pipewire")
 	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, file)
+	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, sock_file)
 
 	# Session D-Bus (register org.PipeWire.* on the per-user session bus).
 	optional_policy(`
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -7,6 +7,8 @@ policy_module(pipewire)
 
 attribute_role pipewire_roles;
 
+attribute pipewire_client_domain;
+
 #
 # pipewire_t — the PipeWire multimedia daemon.
 #
@@ -168,3 +170,15 @@ ifndef(`pipewire_system_service',`
 		systemd_user_daemon_domain(staff, pipewire_exec_t, pipewire_t)
 	')
 ')
+
+########################################
+#
+# pipewire_client domain
+#
+# domain that captures requirements to use the pipewire service
+#
+
+allow pipewire_client_domain pipewire_runtime_t:dir list_dir_perms;
+pipewire_stream_connect(pipewire_client_domain)
+pipewire_use_fds(pipewire_client_domain)
+pipewire_mmap_rw_tmpfs_files(pipewire_client_domain)
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -94,6 +94,8 @@ xdg_search_config_dirs(pipewire_t)
 # ALSA configuration files (/etc/alsa/*, ~/.asoundrc).
 optional_policy(`
 	alsa_read_config(pipewire_t)
+	alsa_read_home_files(pipewire_t)
+	alsa_read_lib(pipewire_t)
 ')
 
 # PulseAudio compatibility layer (pipewire-pulse connects to PulseAudio clients).
@@ -109,6 +111,12 @@ optional_policy(`
 # Read systemd journal for log correlation.
 optional_policy(`
 	systemd_read_journal_files(pipewire_t)
+')
+
+optional_policy(`
+	dbus_list_system_bus_runtime(pipewire_t)
+	dbus_send_system_bus(pipewire_t)
+	dbus_system_bus_client(pipewire_t)
 ')
 
 ########################################
@@ -161,9 +169,6 @@ ifndef(`pipewire_system_service',`
 	optional_policy(`
 		dbus_all_session_bus_client(pipewire_t)
 		dbus_connect_all_session_bus(pipewire_t)
-		dbus_list_system_bus_runtime(pipewire_t)
-		dbus_send_system_bus(pipewire_t)
-		dbus_system_bus_client(pipewire_t)
 		dbus_write_session_runtime_socket(pipewire_t)
 	')
 
@@ -171,6 +176,10 @@ ifndef(`pipewire_system_service',`
 	optional_policy(`
 		systemd_user_daemon_domain(user, pipewire_exec_t, pipewire_t)
 		systemd_user_daemon_domain(staff, pipewire_exec_t, pipewire_t)
+	')
+
+	optional_policy(`
+		xdg_search_config_dirs(pipewire_t)
 	')
 ')
 
--- a/policy/modules/apps/pipewire.te
+++ b/policy/modules/apps/pipewire.te
@@ -56,6 +56,8 @@ manage_sock_files_pattern(pipewire_t, pipewire_runtime_t, pipewire_runtime_t)
 mmap_manage_files_pattern(pipewire_t, pipewire_tmpfs_t, pipewire_tmpfs_t)
 fs_tmpfs_filetrans(pipewire_t, pipewire_tmpfs_t, file)
 
+allow pipewire_t pipewire_client_domain:fd use;
+
 # Audio hardware access.
 dev_read_sound(pipewire_t)
 dev_write_sound(pipewire_t)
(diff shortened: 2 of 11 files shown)


policy/modules/apps/pipewire.te
```diff
@@ -170,6 +172,23 @@ ifndef(`pipewire_system_service',`
...
+########################################
+#
+# pipewire_client domain
+#
+# domain that captures requirements to use the pipewire service
+#
+
+pipewire_stream_connect(pipewire_client)
+pipewire_rw_stream_sockets(pipewire_client)
+pipewire_mmap_rw_tmpfs_files(pipewire_client)
```
kcinimod-defensec: Should probably ideally be pipewire_mmap_rw_inherited_tmpfs_files(pipewire_client)

With these memfd/shm aspects excluding open can be a real win.

Here are my pipewire_client direct rules (in case youre interested)

```
root@nimbus:~# sesearch -A -s pipewire.client.typeattr -ds
allow pipewire.client.typeattr pipewire.conf.file:dir { getattr ioctl lock open read search };
allow pipewire.client.typeattr pipewire.conf.file:file { getattr ioctl lock map open read };
allow pipewire.client.typeattr pipewire.data.file:dir { getattr ioctl lock open read search };
allow pipewire.client.typeattr pipewire.data.file:file { getattr ioctl lock map open read };
allow pipewire.client.typeattr pipewire.home.conf.file:dir { getattr ioctl lock open read search };
allow pipewire.client.typeattr pipewire.home.conf.file:file { getattr ioctl lock map open read };
allow pipewire.client.typeattr pipewire.memfd_file:memfd_file { append getattr ioctl lock map read write };
allow pipewire.client.typeattr pipewire.run.file:sock_file { append getattr ioctl lock open write };
allow pipewire.client.typeattr pipewire.subj:fd use;
allow pipewire.client.typeattr pipewire.subj:unix_stream_socket connectto;
allow pipewire.client.typeattr pipewire.tmpfs.file:file { append getattr ioctl lock map read write };
```

```
root@nimbus:~# sesearch -A -t pipewire.client.typeattr -dt
allow pipewire.subj pipewire.client.typeattr:dir { getattr ioctl lock open read search };
allow pipewire.subj pipewire.client.typeattr:fd use;
allow pipewire.subj pipewire.client.typeattr:file { getattr ioctl lock open read };
allow pipewire.subj pipewire.client.typeattr:lnk_file { getattr lock read };
```

```
root@nimbus:~# sesearch -A -t pipewire.client.memfd.typeattr -dt
allow pipewire.subj pipewire.client.memfd.typeattr:memfd_file { append getattr ioctl lock read write };
```

```
root@nimbus:~# sesearch -A -t pipewire.client.tmpfs.typeattr -dt
allow pipewire.subj pipewire.client.tmpfs.typeattr:file { append getattr ioctl lock read write };
```

The pipewire.client.tmpfs.typeattr is for compatibility with pre-mem_fd.

policy/modules/apps/pipewire.te
```diff
@@ -170,6 +172,23 @@ ifndef(`pipewire_system_service',`
...
 
+########################################
+#
+# pipewire_client domain
+#
+# domain that captures requirements to use the pipewire service
+#
+
+pipewire_stream_connect(pipewire_client)
+pipewire_rw_stream_sockets(pipewire_client)
```
kcinimod-defensec: This is questionable. AVC denial of the event that prompted you to add this would be interesting to interpret. Could just be a leak unrelated to pipewire_client.

kcinimod-defensec: I also noticed that you allow pw clients to dbus chat with pw. I actually don't have these rules (and see these accesses) in my policy but pw indeed maintains a name on the user dbus ... I wonder whether that pipewire_rw_stream_sockets(pipewire_client) might be due to that socket being passed through dbus in your scenario. I am just comparing your rules to what I have as per my sesearch analysis.

policy/modules/apps/pipewire.te
```diff
@@ -170,6 +172,23 @@ ifndef(`pipewire_system_service',`
...
+pipewire_mmap_rw_tmpfs_files(pipewire_client)
+pipewire_use_fds(pipewire_client)
+allow pipewire_t pipewire_client:fd use;
+
+optional_policy(`
+	pipewire_dbus_chat(pipewire_client)
```
kcinimod-defensec: I guess this is pipewire-pulse/org.pulseaudio.Server. Maybe I am overlooking something. Would be nice if there is a way to separate the pulseaudio compat rules because pulseaudio is on its way out. I have a sepearate module for pulseaudio compat specific rules. So that I can just delete that when I decide to get rid of my pipewire-pulse support.

policy/modules/apps/pipewire.te
kcinimod-defensec: o wait ...

```
root@nimbus:~# sesearch --dontaudit -t pipewire.subj -dt
dontaudit pautils.subj pipewire.subj:unix_stream_socket { append bind connect getattr getopt ioctl read setopt shutdown write };
dontaudit pipewire.subj pipewire.subj:capability net_admin;
```

So yes I am pretty sure this is related to pipewire-pulse one way or another. Should probably be dontaudited.

policy/modules/apps/pipewire.fc
```diff
@@ -1,5 +1,6 @@
 # PipeWire daemon executable.
 /usr/bin/pipewire		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
+/usr/bin/wpctl		--	gen_context(system_u:object_r:pipewire_exec_t,s0)
```
kcinimod-defensec: This does not belong here as its part of wireplumber and even if then pipewire_client_exec_t would probably be much more appropriate than pipewire_exec_t.

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+#
+
+allow wireplumber_t self:process { getsched setsched signal signull };
+allow wireplumber_t self:capability sys_nice;
+allow wireplumber_t self:fifo_file rw_fifo_file_perms;
+allow wireplumber_t self:unix_stream_socket { create_stream_socket_perms connectto };
+allow wireplumber_t self:unix_dgram_socket create_socket_perms;
+
+# Bluetooth socket (for BlueZ audio routing)
+allow wireplumber_t self:bluetooth_socket { create bind setopt listen };
```
kcinimod-defensec: create_stream_socket_perms

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+
+# netlink kobject uevent socket (for udev device monitoring)
+allow wireplumber_t self:netlink_kobject_uevent_socket { create bind setopt getattr getopt };
```
kcinimod-defensec: create_socket_perms

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+
+dev_read_sysfs(wireplumber_t)
```
kcinimod-defensec, line 75: Eventually would be nice if we could create some order in /sys and then it would be useful to know what exactly wireplumber_t is accessing below /sys. Then again it access a lot there so maybe best to not go there yet.

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+
+# /dev/snd inotify watch for device monitoring
+dev_watch_dev_dirs(wireplumber_t)
+
+# filesystem getattr
+dev_getattr_fs(wireplumber_t)
+fs_getattr_xattr_fs(wireplumber_t)
+
+# v4l device getattr (video device enumeration)
+dev_dontaudit_getattr_video_dev(wireplumber_t)
```
kcinimod-defensec: Probably best to just let it do its thing?

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+')
+
+# Use file descriptors passed from PipeWire (e.g. memfd)
+optional_policy(`
+	pipewire_use_fds(wireplumber_t)
+')
+
+# Access PipeWire tmpfs shared memory (memfd zero-copy buffers)
+optional_policy(`
+	pipewire_rw_tmpfs_files(wireplumber_t)
```
kcinimod-defensec: I would argue that wireplumber_t should probably be associated with pipewire_client instead.

wireplumber_tmpfs_t would then ideally also be associated with pipewire_client_tmpfs_type

Also again. These accesses are inherited. Open permission should not be needed. I believe that if one just consistently omits open when its not needed that this is worth the trouble in the longer run.

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+optional_policy(`
+	ofono_dbus_chat(wireplumber_t)
+')
+
+optional_policy(`
+	rtkit_scheduled(wireplumber_t)
+')
+
+optional_policy(`
+	systemd_read_journal_files(wireplumber_t)
```
kcinimod-defensec, line 160: questionable. Have AVC denials?

policy/modules/apps/wireplumber.fc
```diff
@@ -0,0 +1,9 @@
+HOME_DIR/\.config/wireplumber(/.*)?	gen_context(system_u:object_r:wireplumber_home_t,s0)
+HOME_DIR/\.local/share/wireplumber(/.*)?	gen_context(system_u:object_r:wireplumber_home_t,s0)
+
+/run/wireplumber(/.*)?		gen_context(system_u:object_r:wireplumber_runtime_t,s0)
+/run/user/%{USERID}/wireplumber(/.*)?	gen_context(system_u:object_r:wireplumber_runtime_t,s0)
```
kcinimod-defensec: Why did you add support for wireplumber_runtime_t sockets? Who calls wireplumber_stream_connect() and why?

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+	pipewire_rw_tmpfs_files(wireplumber_t)
+')
+
+# Runtime directory management - differs between system and user service
+ifdef(`wireplumber_system_service',`
+	files_var_lib_filetrans(wireplumber_t, wireplumber_home_t, { dir file })
+	files_runtime_filetrans(wireplumber_t, wireplumber_runtime_t, { dir file sock_file })
+',`
+	userdom_user_runtime_filetrans(wireplumber_t, wireplumber_runtime_t, dir)
+	userdom_user_home_dir_filetrans(wireplumber_t, wireplumber_home_t, dir)
```
kcinimod-defensec: There are no wireplumber_home_t dirs directly under user_home_dir_t dirs. This rule does not correspond to the wireplumber_home_t fc specs above.

policy/modules/apps/wireplumber.te
```diff
@@ -0,0 +1,138 @@
...
+type wireplumber_t;
+type wireplumber_exec_t;
+ifdef(`wireplumber_system_service',`
+	init_daemon_domain(wireplumber_t, wireplumber_exec_t)
+',`
+	userdom_user_application_domain(wireplumber_t, wireplumber_exec_t)
+	role wireplumber_roles types wireplumber_t;
+')
+
+type wireplumber_home_t alias wireplumber_var_lib_t;
```
kcinimod-defensec: what is the point of that typealias? wireplumber_var_lib_t is not referenced anywhere?

policy/modules/apps/wireplumber.te
kcinimod-defensec: So one could have a macro like:

pipewire_client(wireplumber_t, wireplumber_tmpfs_t)

Something like (untested):

```
interface pipewire_client(`,'
gen_require(` attribute pipewire_client_type, pipewire_client_tmpfs_type; ')
typeattribute $1 pipewire_client_type;
typeattribute $2 pipewire_client_tmpfs_type;
')
```

policy/modules/apps/wireplumber.te
kcinimod-defensec:
```
root@nimbus:~# sesearch -A -s wireplumber.subj -t v4l.nodedev -ds
allow wireplumber.subj v4l.nodedev:chr_file { append getattr ioctl lock open read watch write };
```

aerusso:
@kcinimod-defensec Thank you for the review! I don't have time right now to respond to all of this (in particular, I'll want to demonstrate specific actions that break if I remove anything and/or figure out what wasn't actually necessary).  This will take a minute, but I will work on it.

policy/modules/apps/pipewire.fc
etbe: This doesn't belong because it's a client of the standard interface that desktop environments will use. user_t needs to be able to do all this stuff. So there's no point in having a program doing this running in a different domain.

policy/modules/apps/pipewire.te
aerusso: Preliminary testing with the inherited permissions seems to be working.

policy/modules/apps/pipewire.fc
aerusso: This is gone.

policy/modules/apps/wireplumber.te
aerusso: I agree (and fixed in the parent MR).

policy/modules/apps/wireplumber.fc
aerusso: Parent PR issue (and also resolved, I think).

policy/modules/apps/wireplumber.te
aerusso: This will have to be handled separately, I don't understand this either --- and did not investigate.

aerusso:
I had to rebase this after 3ee9b9517.  I haven't fully addressed all your points, @kcinimod-defensec, so I'm marking this draft and investigating cutting down the dbus permissions.  I'll ping again when this is better organized.

PR #1174, opened 2026-07-06
wayland

etbe:


pebenito:
Rule-wise, I don't see any issues. However, it seems like most of these rules should be folded into `wm_application_domain()` and then any GUI app should likely call the interface. Then the remaining receive fd interface would be renamed to `wm_use_application_fds()` for consistent naming.

etbe:
I'll make this one again from the newest git

PR #1175, opened 2026-07-07
lvm: label dmeventd on all distros, not only gentoo

apryor-owl:
dmeventd is shipped by the device-mapper-event package on Fedora, RHEL and Debian (as /usr/sbin/dmeventd), not just Gentoo, but lvm.fc only labeled it lvm_exec_t under ifdef(`distro_gentoo'). Elsewhere it is left bin_t, so systemd's exec of it for dm-event.service does not fire the lvm_exec_t -> lvm_t domain transition; dmeventd then runs in initrc_t and is denied its lvm_runtime_t control fifos, so dm-event.service/.socket fail and lvm2-monitor exits under enforcing.

file_contexts.subs_dist already maps /sbin and /usr/sbin to /usr/bin, so the existing /usr/bin/dmeventd entry covers every layout once it is no longer gated to Gentoo. Drop the ifdef.

### DIFF
--- a/policy/modules/system/lvm.fc
+++ b/policy/modules/system/lvm.fc
@@ -21,6 +21,7 @@
 # /usr
 #
 /usr/bin/cryptsetup		--	gen_context(system_u:object_r:lvm_exec_t,s0)
+/usr/bin/dmeventd		--	gen_context(system_u:object_r:lvm_exec_t,s0)
 /usr/bin/dmraid			--	gen_context(system_u:object_r:lvm_exec_t,s0)
 /usr/bin/dmsetup		--	gen_context(system_u:object_r:lvm_exec_t,s0)
 /usr/bin/dmsetup\.static	--	gen_context(system_u:object_r:lvm_exec_t,s0)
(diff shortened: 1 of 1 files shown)


policy/modules/system/lvm.fc
```diff
@@ -76,9 +76,7 @@
 /usr/bin/vgsplit		--	gen_context(system_u:object_r:lvm_exec_t,s0)
 /usr/bin/vgwrapper		--	gen_context(system_u:object_r:lvm_exec_t,s0)
 
-ifdef(`distro_gentoo',`
 /usr/bin/dmeventd		--	gen_context(system_u:object_r:lvm_exec_t,s0)
```
pebenito: Please move up to line 24 so the lines are in order.

pebenito (changes requested):

policy/modules/system/lvm.fc
apryor-owl: Fixed

pebenito (approved):

PR #1176, opened 2026-07-08
adb: add SELinux policy for adbd and interactive adb shell session

jaihindy:
This patch introduces SELinux policy support for adbd and interactive
adb shell sessions.

It defines the adbd_t and adb_shell_t domains, establishes the
transition between them, and provides the permissions required for
interactive shell operation.

The module also introduces the adb_unconfined_shell tunables,
allowing interactive adb shell sessions to transition to
the unconfined_t domains instead of the default adb_shell_t domain
when enabled.

Additional debugging-related permissions are isolated behind the
adb_shell_extended_access tunable, which remains disabled by default
to preserve the default security posture while allowing developers to
enable extended debugging capabilities when needed.

### DIFF
--- a/policy/modules/services/adb.te
+++ b/policy/modules/services/adb.te
@@ -1,0 +1,145 @@
+policy_module(adb)
+
+########################################
+#
+# Declarations
+#
+########################################
+
+## <desc>
+## <p>
+## Allow extended adb shell functionality for debugging and interactive shell access.
+## </p>
+## </desc>
+gen_tunable(adb_shell_extended_access, false)
+
+## <desc>
+## <p>
+## Allow adb shell sessions to run in unconfined_t instead of the default adb_shell_t
+## domain.
+## </p>
+## </desc>
+gen_tunable(adb_unconfined_shell, false)
+
+type adbd_t;
+type adbd_exec_t;
+init_daemon_domain(adbd_t, adbd_exec_t)
+
+type adb_shell_t;
+domain_type(adb_shell_t)
+
+########################################
+#
+# adbd local policy
+#
+########################################
+
+allow adbd_t self:capability sys_resource;
+
+corecmd_search_bin(adbd_t)
+dev_rw_usbfs(adbd_t)
+files_manage_generic_tmp_dirs(adbd_t)
+files_manage_generic_tmp_files(adbd_t)
+files_read_etc_files(adbd_t)
+logging_read_audit_log(adbd_t)
+logging_read_generic_logs(adbd_t)
+term_use_generic_ptys(adbd_t)
+term_use_ptmx(adbd_t)
+
+optional_policy(`
+	tunable_policy(`adb_unconfined_shell',`
+		unconfined_shell_domtrans(adbd_t)
+	',`
+		corecmd_shell_domtrans(adbd_t, adb_shell_t)
+	')
+')
+
+########################################
+#
+# adb_shell local policy
+#
+########################################
+role system_r types adb_shell_t;
+
+allow adb_shell_t self:process { getcap setpgid signal };
+allow adb_shell_t self:fifo_file rw_inherited_fifo_file_perms;
+
+allow adb_shell_t adbd_t:fd use;
+
+corecmd_exec_bin(adb_shell_t)
+corecmd_shell_entry_type(adb_shell_t)
+
+dev_search_sysfs(adb_shell_t)
+
+files_getattr_var_lib_dirs(adb_shell_t)
+files_list_default(adb_shell_t)
+files_manage_generic_tmp_dirs(adb_shell_t)
+files_manage_generic_tmp_files(adb_shell_t)
+files_read_default_files(adb_shell_t)
+files_read_etc_files(adb_shell_t)
+files_read_mnt_files(adb_shell_t)
+files_read_mnt_symlinks(adb_shell_t)
+files_read_usr_symlinks(adb_shell_t)
+files_read_var_symlinks(adb_shell_t)
+files_search_spool(adb_shell_t)
+files_search_tmp(adb_shell_t)
+files_search_var(adb_shell_t)
+fs_getattr_xattr_fs(adb_shell_t)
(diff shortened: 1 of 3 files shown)


policy/modules/services/adb_shell.te
```diff
@@ -0,0 +1,111 @@
+policy_module(adb_shell, 1.0)
```
pebenito: The version field is deprecated.

pebenito (changes requested):
How much testing has this received? For example, adb can operate over IP, but I see nothing about that in `adbd_t`. What about adb operations other than the shell?

There should probably also be tunables controlling which domain you end up in for the shell. It seems like we should be able to toggle between unconfined (if available), sysadm, and the restricted adb_shell_t.

Please rename the module to simply "adb".

Please refer to [StyleGuide](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide) as this needs style cleanup.

policy/modules/services/adb_shell.te
```diff
@@ -0,0 +1,111 @@
...
+files_read_usr_symlinks(adb_shell_t)
+files_getattr_var_lib_dirs(adb_shell_t)
+seutil_read_config(adb_shell_t)
+files_read_mnt_files(adb_shell_t)
+files_search_default(adb_shell_t)
+fs_list_auto_mountpoints(adb_shell_t)
+files_read_var_symlinks(adb_shell_t)
+files_read_mnt_symlinks(adb_shell_t)
+
+dbus_system_bus_client(adb_shell_t)
```
thesamesam: Is dbus mandatory for adb? If not, this should be an `optional_policy`.

jaihindy:
> How much testing has this received? For example, adb can operate over IP, but I see nothing about that in `adbd_t`. What about adb operations other than the shell?
> 
> There should probably also be tunables controlling which domain you end up in for the shell. It seems like we should be able to toggle between unconfined (if available), sysadm, and the restricted adb_shell_t.
> 
> Please rename the module to simply "adb".
> 
> Please refer to [StyleGuide](https://github.com/SELinuxProject/refpolicy/wiki/StyleGuide) as this needs style cleanup.

Thanks for the review and suggestions.

I have renamed the module from adb_shell to adb and started cleaning up the policy according to the refpolicy StyleGuide.

For testing, the current testing has mainly covered adbd startup and interactive adb shell over the local/USB path. I will expand the test coverage and document the results in the PR description, including:

- adbd daemon startup and domain transition
- adb shell
- adb shell dmesg 
- adb shell mount 
- adb push / adb pull
- adb shell (interactive)
- adb shell <command>
I am running  more adb/adb shell testcases and add sepolicy rules for them.

Regarding the shell domain tunables, I attempted to model this as tunable-controlled transitions from adbd_t to adb_shell_t, sysadm_t, and unconfined_t using shell_exec_t as the entrypoint. However, this results in conflicting type_transition rules for the same tuple as below.
| Conflicting type rules (scontext=adbd_t tcontext=shell_exec_t tclass=process result=unconfined_t), existing=adb_shell_t
| Expanded from type rule (scontext=adbd_t tcontext=shell_exec_t tclass=process result=unconfined_t)
| Failed to insert typerule into avtab at /refpolicy-targeted/2.20260312+git/image/etc/selinux/targeted/tmp/modules/100/adb/cil:245
| Failure while walking true conditional block at /refpolicy-targeted/2.20260312+git/image/etc/selinux/targeted/tmp/modules/100/adb/cil:241
| Binary policy creation failed at /refpolicy-targeted/2.20260312+git/image/etc/selinux/targeted/tmp/modules/100/adb/cil:240
| Failed to generate binary
| semodule:  Failed!
For example, selecting between adb_shell_t and unconfined_t through conditionals still causes semodule/libsepol to reject the policy because multiple transition results exist for the same source/entrypoint/class.
please let me know if i am missing here something?

Can you please help to review the change.

policy/modules/services/adb_shell.te
jaihindy: done

pebenito:
> Regarding the shell domain tunables, I attempted to model this as tunable-controlled transitions from adbd_t to adb_shell_t, sysadm_t, and unconfined_t using shell_exec_t as the entrypoint.

You need to make sure the transitions are in opposite if-then-else blocks, e.g.:

```
if (adb_sysadm_shell) {
sysadm_domtrans(...)
} else {
corecmd_shell_domtrans(...)
}
```

jaihindy:
> > Regarding the shell domain tunables, I attempted to model this as tunable-controlled transitions from adbd_t to adb_shell_t, sysadm_t, and unconfined_t using shell_exec_t as the entrypoint.
> 
> You need to make sure the transitions are in opposite if-then-else blocks, e.g.:
> 
> ```
> if (adb_sysadm_shell) {
> sysadm_domtrans(...)
> } else {
> corecmd_shell_domtrans(...)
> }
> ```

thanks for the suggestion and I have updated the change .
could you please review the change.

policy/modules/services/adb.te
```diff
@@ -0,0 +1,173 @@
...
+files_manage_generic_tmp_files(adb_shell_t)
+files_manage_generic_tmp_dirs(adb_shell_t)
+
+seutil_read_config(adb_shell_t)
+
+userdom_search_user_home_dirs(adb_shell_t)
+
+allow adb_shell_t self:fifo_file rw_inherited_fifo_file_perms;
+allow adb_shell_t self:process { setpgid getcap signal };
+allow adb_shell_t self:capability { net_admin sys_resource };
```
pebenito: Why are these capabilities needed by default? There should be transitions to other domains for tools that need them.

pebenito (changes requested):
Theres still style issues.

policy/modules/services/adb.te
jaihindy: thanks , removed these capabilities.
can you please review it again.

pebenito (approved):

PR #1177, opened 2026-07-11
Fix file_contexts for systemd runtime generator output directories

7-RED:
## Summary
Fixes incorrect `init_runtime_t` labeling for systemd generator runtime directories.

## Problem
The wildcard rule `/run/systemd(/.*)? -> init_runtime_t` in `init.fc` causes generator output directories (`/run/systemd/generator`, `generator.early`, `generator.late`, `system`, `units`) to be labeled with `init_runtime_t` instead of the semantically correct `systemd_unit_t`.

This creates two issues:
1. **Relabel operations break generator outputs**: When `restorecon` or `fixfiles` runs, generator-produced unit files are incorrectly relabeled from `systemd_unit_t` to `init_runtime_t`, breaking tools that expect proper unit file labels
2.  **Attribute-based permissions fail**: Domains using interfaces like `init_manage_all_unit_files() ` (which grants access to the `systemdunit` attribute) cannot access generator outputs without additional workarounds

This bug has existed unnoticed because generator outputs regenerate on every boot from tmpfs (/run), rarely triggering relabel operations.

## Solutions
- **policy/modules/system/init.fc**: Add 5 explicit file_contexts rules for runtime generator paths to match systemd_unit_t before the fallback
- **policy/modules/system/systemd.te**: Grant systemd_generator_t permission to manage all unit file types via init_manage_all_unit_files()

## Testing recommendation
```bash
# Before patch: generator outputs get init_runtime_t
ls -Zd /run/systemd/generator

restorecon -Rv /run/systemd/generator # No change (wrong label persists)

# After patch: proper systemd_unit_t labeling
ls -Zd /run/systemd/generator
# drwxr-xr-x. root root system_u:object_r:systemd_unit_t:s0 /run/systemd/generator

Related: #1173 
```

### DIFF
--- a/policy/modules/system/systemd.te
+++ b/policy/modules/system/systemd.te
@@ -647,6 +647,7 @@ fs_getattr_nsfs_files(systemd_generator_t)
 
 init_create_runtime_files(systemd_generator_t)
 init_read_all_script_files(systemd_generator_t)
+init_manage_generic_unit_files(systemd_generator_t)
 init_manage_runtime_dirs(systemd_generator_t)
 init_manage_runtime_symlinks(systemd_generator_t)
 init_read_runtime_files(systemd_generator_t)
--- a/policy/modules/system/init.fc
+++ b/policy/modules/system/init.fc
@@ -43,6 +43,11 @@ ifdef(`distro_gentoo',`
 /usr/lib/systemd/ntp-units\.d -d gen_context(system_u:object_r:systemd_unit_t,s0)
 /usr/lib/systemd/system(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
 /usr/share/containers/systemd(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
+/run/systemd/generator(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
+/run/systemd/generator\.early(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
+/run/systemd/generator\.late(/.*)?	gen_context(system_u:object_r:systemd_unit_t,s0)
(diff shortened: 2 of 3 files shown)


policy/modules/system/systemd.te
```diff
@@ -646,6 +646,7 @@ fs_getattr_all_fs(systemd_generator_t)
 fs_getattr_nsfs_files(systemd_generator_t)
 
 init_create_runtime_files(systemd_generator_t)
+init_manage_all_unit_files(systemd_generator_t)
```
pebenito: I'm not clear why this is needed. I would only expect a generator to work with `systemd_unit_t`.

pebenito:
also, please rebase so there isn't a merge commit in the PR.

policy/modules/system/systemd.te
7-RED: @pebenito Thanks for the feedback! I've updated the PR:

  Replaced `init_manage_all_unit_files()` with explicit permissions for
  `systemd_unit_t` only:
     allow systemd_generator_t systemd_unit_t:dir manage_dir_perms;
     allow systemd_generator_t systemd_unit_t:file manage_file_perms;
     allow systemd_generator_t systemd_unit_t:lnk_file manage_lnk_file_perms;
  The generator now only has permissions for `systemd_unit_t`, not all unit types.

7-RED:
Rebased onto latest `main` to remove merge commits.

  Please let me know if this looks better!

7-RED:
There's no existing interface for managing only systemd_unit_t (not the systemd unit attribute). Should I create a new interface like init_manage_generic_unit_files(), or is there a preferred approach?

pebenito:
Yes please.

pebenito (approved):

PR #1189, opened 2026-07-28
pipewire and wireplumber

etbe:


policy/modules/apps/pipewire.if
```diff
@@ -111,6 +111,46 @@ interface(`pipewire_domtrans_client',`
...
+## <summary>
+##	unlink pipewire user socket /run/user/$UID/pipewire-0
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_socket_user_unlink',`
```
pebenito: `pipewire_delete_user_sockets()`

policy/modules/apps/pipewire.te
```diff
@@ -154,6 +154,8 @@ ifndef(`pipewire_system_service',`
 
 	# Runtime socket directory transition under $XDG_RUNTIME_DIR/pipewire.
 	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, dir, "pipewire")
+	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, sock_file)
+	userdom_user_runtime_filetrans(pipewire_t, pipewire_runtime_t, file)
```
pebenito: I'm a little confused. Are these (sock) files not created under `$XDG_RUNTIME_DIR/pipewire`?

policy/modules/apps/wireplumber.te
```diff
@@ -42,13 +42,13 @@ userdom_user_tmpfs_file(wireplumber_tmpfs_t)
...
 allow wireplumber_t self:capability sys_nice;
-allow wireplumber_t self:process { getsched setsched signal signull };
+allow wireplumber_t self:process { getsched setsched signal signull setrlimit };
 allow wireplumber_t self:fifo_file rw_fifo_file_perms;
 allow wireplumber_t self:unix_stream_socket { create_stream_socket_perms connectto };
 allow wireplumber_t self:unix_dgram_socket create_socket_perms;
 
 # Bluetooth socket for BlueZ audio routing (A2DP/HFP profile management).
-allow wireplumber_t self:bluetooth_socket create_socket_perms;
+allow wireplumber_t self:bluetooth_socket { listen create_socket_perms };
```
pebenito: If `listen` is needed, then eventually I'd expect to see `accept`, so I think it should go up to `create_stream_socket_perms`.

policy/modules/apps/wireplumber.te
```diff
@@ -93,11 +96,26 @@ miscfiles_read_localization(wireplumber_t)
 udev_read_runtime_files(wireplumber_t)
 udev_search_runtime(wireplumber_t)
 
+xdg_search_config_dirs(wireplumber_t)
+xdg_search_data_dirs(wireplumber_t)
+xdg_exec_data(wireplumber_t)
```
pebenito: The exec is concerning for arbitrary code execution since it can also write the data. What is executed?

policy/modules/apps/wireplumber.te
```diff
@@ -93,11 +96,26 @@ miscfiles_read_localization(wireplumber_t)
...
+xdg_manage_data(wireplumber_t)
+
 # D-Bus communication with BlueZ (Bluetooth audio routing).
 optional_policy(`
 	bluetooth_dbus_chat(wireplumber_t)
 ')
 
+optional_policy(`
+	dbus_list_system_bus_runtime(wireplumber_t)
+	dbus_write_session_runtime_socket(wireplumber_t)
```
pebenito, line 111: This seems like it would be part of connecting to the session bus. I see some changes in `*_systemd_t` , so I think we should keep an eye on this.

policy/modules/system/userdomain.if
```diff
@@ -3674,6 +3674,25 @@ interface(`userdom_search_user_runtime',`
...
+## <summary>
+##	rw users runtime directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`userdom_rw_user_runtime',`
```
pebenito: userdom_rw_user_runtime_dirs

pebenito (changes requested):

policy/modules/apps/pipewire.te
etbe: Not in Debian/Testing:
etbe@dojacat:\~$ ls -l $XDG_RUNTIME_DIR/pipewire*
ls: cannot access '/run/user/1000/pipewire-0.lock': Permission denied
ls: cannot access '/run/user/1000/pipewire-0-manager.lock': Permission denied
srw-rw-rw-. 1 etbe etbe 0 Jul 29 00:44 /run/user/1000/pipewire-0
srw-rw-rw-. 1 etbe etbe 0 Jul 29 00:44 /run/user/1000/pipewire-0-manager
etbe@dojacat:~$ dpkg -l pipewire
┌── Desired=Unknown/Install/Remove/Purge/Hold
│┌─ Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
││┌ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
│││ Name           Version      Architecture Description
├┼┼─══════════════─════════════─════════════─═══════════════════════════════════════════════════
ii  pipewire:amd64 1.6.8-1      amd64        audio and video processing engine multimedia server

policy/modules/apps/wireplumber.te
etbe: Maybe, but we can go back and do a bulk change later, it is better not connected to this change.

policy/modules/apps/wireplumber.te
etbe: I can't reproduce this so I removed it.

etbe:
@pebenito I think my changes addressed every issue you raised.

policy/modules/apps/pipewire.if
```diff
@@ -111,6 +111,46 @@ interface(`pipewire_domtrans_client',`
...
+##	</summary>
+## </param>
+#
+interface(`pipewire_delete_user_sockets',`
+	gen_require(`
+		type pipewire_runtime_t;
+	')
+
+	userdom_rw_user_runtime_dirs($1)
+	allow $1 pipewire_runtime_t:sock_file unlink;
```
pebenito, line 151: Considering we have other distros that may have a `/run/user/$UID/pipewire/*`, we should upgrade this to use `delete_sock_files_pattern()` to get the appropriate dir access on `pipewire_runtime_t`.

policy/modules/apps/pipewire.if
```diff
@@ -111,6 +111,46 @@ interface(`pipewire_domtrans_client',`
...
+## </param>
+#
+interface(`pipewire_socket_user_create',`
+	gen_require(`
+		type pipewire_runtime_t;
+	')
+
+	userdom_user_runtime_filetrans($1, pipewire_runtime_t, sock_file, "pipewire-0")
+	userdom_user_runtime_filetrans($1, pipewire_runtime_t, sock_file, "pipewire-0-manager")
+	userdom_user_runtime_filetrans($1, pipewire_runtime_t, dir, "pulse")
```
pebenito, line 131: We should have .fc entries for these. Can you try with that and see if systemd will create them correctly without these filetrans?

If the interface is still needed, then it should be named `pipewire_create_user_sockets` and include corresponding create_sock_file_perms/create_dir_perms for these filetrans.

pebenito (changes requested):

policy/modules/apps/pipewire.if
etbe: ok I'll do that in my next version.

policy/modules/apps/pipewire.if
etbe: ok the sock_file creation obeys .fc directive and dir creation doesn't. I've adjusted it accordingly.

etbe:
I'll create a new PR with the latest stuff

PR #1193, opened 2026-08-05
Overhaul Documentation

pebenito:
- Move documentation from wiki to this repository
- Split README into different documents
- Add a markdown conversion of the original refpolicy whitepaper
- Add a security goals document
- Add a security architecture document
- Add AGENTS.md for AI/LLM use


pcmoore:
Happy to see the conversion to Markdown for a lot of these docs, thanks Chris!

doc/BUILD_INSTALL.md
```diff
@@ -0,0 +1,125 @@
...
+
+Running an installed policy requires a Linux kernel with SELinux support and
+the SELinux userspace tools used to install, load, manage, and label the policy.
+
+## General Make Targets
+
+- **`install-src`**: Install policy sources into `/etc/selinux/NAME/src/policy`, where `NAME` is
+  defined in the Makefile. If unset, `TYPE` is used. The default `NAME` is `refpolicy`. A
+  pre-existing source policy is moved to `/etc/selinux/NAME/src/policy.bak`.
+- **`conf`**: Regenerate `policy.xml` and update or create `modules.conf` and `booleans.conf`. Run
```
dburgener: nit: I think it would be clearer to mention modules.conf first, since IMO that's the most important thing built by this.

doc/BUILD_INSTALL.md
```diff
@@ -0,0 +1,125 @@
...
+  this after adding or removing modules, or after running `bare`. Existing configuration settings
+  are preserved. This must be run on policy sources checked out from the CVS repository before
+  they can be used.
+- **`clean`**: Delete temporary files, compiled policies, and `file_contexts`. Configuration files
```
dburgener, line 25: I'm trying to think about this with my policy developer hat off and wondering if the mention of `file_contexts` might make someone incorrectly think this will delete `.fc` files?  I'm wondering if something like "the generated `file_contexts` file" would be a sufficient clarity improvement to overcome the added verbosity?

Pushing against that, I see that this just directly converts existing text from the README to markdown, and I'm not aware of anyone being confused by this text in the years its existed, so maybe the existing text is fine.

doc/BUILD_INSTALL.md
```diff
@@ -0,0 +1,125 @@
...
+  are left intact.
+- **`bare`**: Run `clean` and also delete configuration files, web documentation, and `policy.xml`.
```
dburgener: I'm not sure exactly what "configuration files" is supposed to refer to, but to make this description match the behavior, you have to interpret it very broadly (e.g. including generated policy files).  And excluding policy.xml from the "configuration files"category even though it's built by `make conf` is weird.

According to the Makefile, this is clean, the items generated by `make conf`, documentation files, ctags file (if enabled), pycache python files, and generated policy files.

I suggest that rather than enumerating and trying to keep in sync, this says something like "Run `clean` and delete additional generated files (e.g. the output of `make conf` and `make html`)".  That lists those as examples and doesn't get bogged down in implementation details, but I think is more accurate and less confusing than the text as written.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,249 @@
...
+the module that declares them. This means that it is not possible, for
+example, to directly refer to the domain attribute. Instead, macros in
+other modules are used to declare that a type will be used for a certain
+purpose. These macros will likely use attributes (but not necessarily),
+but it allows the module that declared the attribute to strictly control
+how it can be used. In this example interfaces are used to transform the
+types into a domain, entry file, log file, and temporary file.
+
+Let's expand this example further by allowing some access for these
+types. My application needs access between its own types and access to
```
dburgener, line 57: Maybe I'm missing it, but I don't see anywhere that you have sample policy for this reading random numbers part.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,249 @@
...
+This allows myapp\_t to write to its private types, but it needs to be
+able to create its temporary files in /tmp. This requires a call to the
+files module.
+
+```sepolicy
+files_tmp_filetrans(myapp_t,myapp_tmp_t,file)
+```
+
+This call to the files module allows myapp\_t to create myapp\_tmp\_t
+files in the /tmp directory.
```
dburgener: And it also sets up the automatic type transition so that all files my_app_t creates in /tmp get the myapp_tmp_t label.

doc/GETTING_STARTED.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Module FC Policy {#ModuleFCPolicy}
```
dburgener, line 89: I always find it helpful to mention that file_contexts updates don't change policy labels on a policy load.  The user needs to run setfiles, restorecon or `make relabel` to apply these labels to files.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Derived Types
+
+Reference Policy uses strong naming conventions to aid policy management and understanding. For
+objects related to a domain, common type names are derived by adding a suffix to the domain's base
+name:
+
+| Generic type       | Derived type                              | Access | Purpose                             |
+| ------------------ | ----------------------------------------- | ------ | ----------------------------------- |
+| `bin_t`            | `*_exec_t`                                | RO     | Entry-point executable              |
```
dburgener, line 76: I'm not 100% sure how to make it clean in the table, but `*_bin_t` is probably worth a mention as well.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,245 @@
...
+is primarily determined by type, while roles constrain process-domain selection. The SELinux user
+identity and MLS or MCS ranges can further constrain object access.
+
+## User Architecture
+
+An SELinux user identity is a field of a security context and is distinct from a Linux account.
+Login configuration maps Linux accounts to SELinux users. Each SELinux user declaration specifies
+the roles it may enter, its default MLS or MCS level, and its authorized range or categories.
+
+The core policy declares a small set of SELinux users. `system_u` is reserved for system processes
```
dburgener, line 170: I think it might be helpful to call out explicitly that the typical pattern of Linux users to SELinux users is a many to one pattern.

doc/SECURITY_GOALS.md
```diff
@@ -0,0 +1,244 @@
...
+- executables, libraries, devices, and persistent state required by the trusted computing base.
+
+Authority to load policy, change enforcement state, relabel protected objects, administer security
+services, or modify their configuration should be confined to explicitly authorized administrative
+and system domains.
+
+In particular, authority to load policy, change Boolean values, or set enforcement mode should be
+limited to the domains identified for those functions. Direct kernel module loading and powerful
+capabilities such as `CAP_SYS_MODULE`, `CAP_SYS_ADMIN`, `CAP_SYS_RAWIO`, and `CAP_NET_ADMIN` should
+be restricted to domains with a documented need.
```
dburgener: I feel like the standard here could be clearer.  In general, refpolicy is applying the principle of least privilege, so all access should be restricted to domains with a documented need.  The point being made here is that some heightened standard is applied to these permissions that can compromise self protection, but I'm not clear exactly what that standard is.

AGENTS.md
```diff
@@ -0,0 +1,163 @@
...
+  `doc/policy.xml` data.
+- `make xml` regenerates `doc/policy.xml`; `make html` regenerates `doc/html/`.
+- Generated policy, configuration, XML, HTML, and `tmp/` outputs are ignored build artifacts. Do
+  not commit them.
+
+### Semantic Review
+
+- Review `sediff` output between the base and changed compiled policies for policy behavior
+  changes when SETools and both policy artifacts are available.
+- Run `sechecker testing/sechecker.ini` against the compiled policy when SETools is available.
```
thesamesam: Isn't `sediff` from the previous point also from SETools? I'd just require it to be available.

doc/HOW_TO_CONTRIBUTE.md
```diff
@@ -0,0 +1,97 @@
...
+If you are interested in working on Reference Policy, feel free to
+contact the developers on the Reference Policy mailing
+list. To join the mail list, send a plaintext email to <majordomo@vger.kernel.org>
+containing "subscribe selinux-refpolicy" in the body. All public
+development related discussion happens on this mailing list. The IRC
+channel, \#selinux on irc.libera.chat, is also appropriate for
+development discussions.
+
+Reference Policy changes should be submitted by pull requests on GitHub or by sending patches to
+the mail list. Pull requests on GitHub are preferred.
```
thesamesam: We should say somewhere that commit messages should include the denials/avcs.

AGENTS.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Change Hygiene
+
+- Review generated policy differences when a policy build changes behavior.
+- Include related `.te`, `.if`, and `.fc` updates in the same change when they form one policy
+  feature.
+- Keep each commit to one logical change and ensure the tree builds after each commit.
+- Include a `Signed-off-by` line using the contributor's real name in every commit.
+- Do not commit generated build artifacts.
+- Submit changes through a GitHub pull request as described in `doc/HOW_TO_CONTRIBUTE.md`.
```
thesamesam, line 157: I'm not really keen on the idea of seeing LLM-authored PRs. We already get people submitting PRs where the obvious solution is not the right one. It's more valuable to have a detailed bug report or a PR where someone states they're unsure rather than an LLM authoritatively saying something.

I try to monitor refpolicy PRs (and I know others do too) and we don't want people to be discouraged from doing that.

doc/HOW_TO_CONTRIBUTE.md
21sw-clayton: A reproducer is also helpful for context if possible - i.e. the command that triggered it (#1102).

Including the distro (and relevant customizations) would also potentially help.

21sw-clayton:
I like the documentation move to a typical repo; I've previously wanted to contribute to the documentation but didn't have an obvious way to do it via the Wiki.

The Wiki _could_ still be published (as a reader-friendly version) via an Action.

doc/BUILD_INSTALL.md
pebenito: The quoting (and format in rendered doc) is meant to emphasize that is this exact file name deleted.

doc/BUILD_INSTALL.md
pebenito: > I'm not sure exactly what "configuration files" is supposed to refer to

modules.conf, booleans.conf

doc/SECURITY_ARCHITECTURE.md
pebenito: `*_bin_t` isn't a pattern, at least not one to use normally.

AGENTS.md
pebenito: > I'm not really keen on the idea of seeing LLM-authored PRs. 

Removing this statement or even this entire file won't stop someone from having an LLM generate a PR, if they really feel they need to.

> I try to monitor refpolicy PRs (and I know others do too) and we don't want people to be discouraged from doing that.

The reviews are much appreciated. Feel free to call out low-quality PRs or even me.

doc/SECURITY_GOALS.md
pebenito: I may just strike this paragraph and make it clear this section lists examples, but is not limited to what is listed. That it's not an enumeration of everything.

pebenito:
> I've previously wanted to contribute to the documentation but didn't have an obvious way to do it via the Wiki.

I didn't realize there were permissions problems. I thought anyone with a github account could edit...

AGENTS.md
thesamesam: OK, fair enough.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Type Architecture
+
+On an SELinux system, every process and object has an SELinux security context which identifies
+the object's security attributes. Policy rules authorize interactions between a source context and
+a target context for specific object classes and permissions. Access is denied unless the loaded
+policy allows it and all applicable constraints are satisfied.
+
+Type Enforcement (TE) is SELinux's primary access-control mechanism, and TE rules make up most
+of Reference Policy. Achieving the security goals depends on assigning and using policy types
```
stephensmalley, line 11: This is the first use of "type" but does not define it. I would typically say that a type in Type Enforcement is a security equivalence class, i.e. all entities with the same type have exactly the same security properties/accesses.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+carefully.
+
+### Domains
+
+A domain is a process type with the `domain` attribute. One or more processes may share a domain,
+which forms their logical security boundary. Separate domains allow policy to grant each domain
+the rules required for its function and limit the effect of compromise.
+
+Domain type names normally use the `_t` suffix and correspond to a service name or process name.
```
stephensmalley: All types in refpolicy use the _t suffix, so I'm not sure why this is stated as being specific to domain type names.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+For example, the `sshd` service runs in the `sshd_t` domain. The suffix is a naming convention;
+membership in the `domain` attribute is what makes a type a domain.
```
stephensmalley: Technically what makes a type a domain is that it is assigned to a process. The domain attribute is just a refpolicy convention.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+
+A typical service has a domain type for its process and an executable type for its entry-point
```
stephensmalley, line 24: First use of "executable type" without a definition. An executable type is merely a type assigned to an executable file.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+program. A domain transition changes the process type when an authorized caller executes that
+entry point. Interfaces such as `init_daemon_domain()` create the domain, associate its entry
+point, authorize it for `system_r`, and establish the expected transition from init.
+
+Process transitions must be explicit and constrained. The core policy uses `neverallow` rules to
+prevent transitions to types that are not domains.
```
stephensmalley: If by "constrained" you mean more than just these neverallow rules, you may wish to qualify this further, e.g. every transition requires an explicit allow of :process transition or dyntransition permission between the source and target domains along with the requisite entrypoint, execute, etc permissions for the executable.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
+# Reference Policy Security Architecture
+
+## Type Architecture
+
+On an SELinux system, every process and object has an SELinux security context which identifies
+the object's security attributes. Policy rules authorize interactions between a source context and
```
stephensmalley: Typically we think of a process as a subject (when acting) so may want to say the subject's or object's here.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+program. A domain transition changes the process type when an authorized caller executes that
+entry point. Interfaces such as `init_daemon_domain()` create the domain, associate its entry
+point, authorize it for `system_r`, and establish the expected transition from init.
+
+Process transitions must be explicit and constrained. The core policy uses `neverallow` rules to
+prevent transitions to types that are not domains.
+
+### Object Types
+
+Object types, not to be confused with object classes, label resources accessed by domains. They
```
stephensmalley: Object types are types assigned to objects, i.e. anything other than a process.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+identify security purpose rather than pathname, owner, or discretionary mode. Object classes
+distinguish files, directories, symbolic links, sockets, devices, filesystems, ports, nodes,
+packets, and IPC objects that have the same type.
+
+Types should separate objects with different readers, writers, trust levels, lifecycles, or
```
stephensmalley: or executers?

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+transition behavior. A service commonly has distinct types for executable code, configuration,
+persistent state, logs, temporary data, and runtime objects. Network, device, filesystem, and IPC
+types similarly distinguish resources that Linux discretionary access control may otherwise treat
+alike.
+
+### Type Attributes
+
+Attributes group types that share a security property. Core examples include `domain`,
```
stephensmalley: Attributes are type sets.

stephensmalley: or at least type attributes are type sets, not to be confused with role attributes.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+function requires it.
+
+## Role Architecture
+
+Reference Policy uses SELinux role-based access control to constrain which process domains an
+SELinux user may enter. The process's role limits which domain types are available to that user.
+
+The authorization path is:
+
+1. A Linux account is mapped to an SELinux user.
```
stephensmalley: Possibly not relevant in this context but technically seusers also maps a Linux user or group to an authorized MLS/MCS range, which could be a subset of the range authorized for the SELinux user.

stephensmalley: NB This is only enforced at login time not by the kernel policy.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+
+Specialized roles reduce the need to use `sysadm_r` for every administrative task. Their security
+value depends on keeping their domain permissions narrower than general system administration and
+avoiding transition paths to unrelated administrative domains.
+
+### Role Transitions
+
+A role transition requires all relevant authorization: the SELinux user must be permitted to use
+the target role, the target role must be permitted to use the target domain, and policy must allow
+the process and entry-point transition. Tools and domains for `newrole`, `sudo`, `su`, login, and
```
stephensmalley: and the role allow rule.

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+similar mechanisms provide controlled transition paths.
+
+Role templates for applications authorize role-specific application domains without making those
+domains available to every role. Optional modules and tunables may add transition paths, so each
+configured policy must be reviewed as a whole.
+
+Objects normally use `object_r` rather than an interactive or system process role. Object access
```
stephensmalley, line 160: Did RBACSEP go away?

doc/SECURITY_ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+Login configuration maps Linux accounts to SELinux users. Each SELinux user declaration specifies
+the roles it may enter, its default MLS or MCS level, and its authorized range or categories.
+
+The core policy declares a small set of SELinux users. `system_u` is reserved for system processes
+and objects and must not be assigned to an ordinary Linux user. `user_u` is the generic identity
+for Linux accounts without a specific SELinux user mapping. `staff_u`, `sysadm_u`, and `root`
+provide different sets of administrative roles and ranges. Deployments may define additional
+SELinux users and map Linux accounts according to local responsibilities. In many cases, the
+existing SELinux users are sufficient, and instead, the mapping of Linux user accounts to SELinux
+users in a many-to-one pattern is the preferred method for managing a Linux user's SELinux user.
```
stephensmalley: and allowed MCS/MLS ranges.

doc/SECURITY_ARCHITECTURE.md
pebenito: The intended audience should know a little about SELinux, so I don't want to define basics in the text here. However, I will add a glossary with an entry that has the definition.

doc/SECURITY_ARCHITECTURE.md
pebenito: I'm trying to say that the domain generally corresponds to the process name, not to say that _t is limited to services. e.g. you can safely infer that sshd_t is the domain type for the sshd process

doc/SECURITY_GOALS.md
```diff
@@ -0,0 +1,244 @@
...
+- authentication, authorization, audit, logging, and system initialization services; and
+- executables, libraries, devices, and persistent state required by the trusted computing base.
+
+Authority to load policy, change enforcement state, relabel protected objects, administer security
+services, or modify their configuration should be confined to explicitly authorized administrative
+and system domains.
+
+In particular, authority to load policy, change Boolean values, or set enforcement mode should be
+limited to the domains identified for those functions. Direct kernel module loading and powerful
+capabilities such as `CAP_SYS_MODULE`, `CAP_SYS_ADMIN`, `CAP_SYS_RAWIO`, and `CAP_NET_ADMIN` should
```
jbrindle: I think this should refer to the selinux permissions rather than the kernel cap names, e.g., capability:sys_rawio instead of CAP_SYS_RAWIO

doc/SECURITY_GOALS.md
```diff
@@ -0,0 +1,244 @@
...
+mandatory confidentiality constraints in addition to Type Enforcement. Its sensitivity levels and
+categories implement the Bell-LaPadula model with the strong star property for ordinary
+information flows: a subject may read at its level or lower, but may write only at the same level.
+This prevents reading information from a higher level and writing information to a different
+level.
+
+MLS constraints apply to files, processes, IPC, networking, and other supported object classes.
+Explicitly trusted domains and objects may receive narrowly scoped exceptions for cross-level
+operations such as logging, relabeling, administration, or trusted data transfer. These exceptions
+are part of the confidentiality boundary and should be limited to functions that require them.
```
jbrindle, line 157: Examples of mls exception attributes here?

jbrindle: Examples and explanations, e.g., mlsfileread is an extremely trusted attribute reserved for TCB components that humans cannot manipulate; mlsfilereadtoclr is for domains that are MLS unaware applications that need to read their entire range; mlsfilewritetoclr is for domains that are MLS _aware_ and can make a decision about writing at a different from current level.

There is a lot of nuance in the mls attributes that isn't well documented

doc/SECURITY_GOALS.md
```diff
@@ -0,0 +1,244 @@
...
+- Users authorized for administrative roles should use an unprivileged role for routine activity.
+- Administrative functions should be divided among purpose-specific domains and, where supported,
+  roles rather than concentrated in a single general-purpose domain.
+- Services should receive individual domains whose privileges are limited to their defined
+  responsibilities.
+- Capabilities, privileged device access, process control, relabeling, and policy administration
+  should be granted independently and only where required.
+- A process should change its own security context only when the transition is expected and
+  constrained. Dynamic transitions from the kernel during system initialization should not permit
+  targets other than the kernel or init domains.
```
jbrindle: I don't understand this. Dynamic transitions are for SELinux aware applications to bracket their privileges for specific operations, or further constrain their privileges in a one-way manner without an associated `execute`.

doc/SECURITY_GOALS.md
```diff
@@ -0,0 +1,244 @@
...
+execution.
+
+The automated checks cover only selected integrity and privilege-separation properties. Passing
+them does not establish all goals in this document or correctness of every allowed access.
+
+## Review Criteria
+
+Policy changes should be reviewed against these questions:
+
+1. Does the change grant a domain authority beyond its documented function?
```
jbrindle: authority is a strange word here, I'd say privilege

doc/SECURITY_GOALS.md
pebenito: I disagree. Since this is the goals doc, I think it should stay as-is, since the kernel cap names are more commonly known and understood.

doc/SECURITY_GOALS.md
pebenito: I will put those details in a (future) doc on MLS

doc/SECURITY_ARCHITECTURE.md
pebenito: It never really landed, because I was trying to make it work before the `default_*` support existed. Files weren't getting the user's role. I never revisited it because none of the multi-role systems I worked on wanted/needed role changes, so UBACSEP with 1:1 seuser:role mappings was sufficient.

I'd be open to revisiting it, but no one has really expressed interest in it.

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+## Derived Type
+
+A Reference Policy naming convention in which the name of an object [type](#type) is formed from a
+[domain's](#domain) base name and a suffix that identifies the object's purpose. For example,
+`sshd_exec_t` and `sshd_log_t` are types derived from the `sshd` base name. A derived type has no
+special meaning to SELinux solely because of its name.
+
+## Domain
+
+A process [type](#type) that an equivalence class or security boundary for one or more processes.
```
stephensmalley: s/that/is/?

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+Processes in the same domain are governed by the same
+[Type Enforcement](#type-enforcement-te) rules. In Reference Policy, domain types have the
+`domain` attribute and conventionally end in `_t`.
+
+## Domain Transition
+
+A change from one process [domain](#domain) to another when executing a file with an
+[executable type](#executable-type). A transition requires policy authorization for the
```
stephensmalley: or when explicitly switching contexts via setcon(3).

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+source-to-target transition, execution of the [entrypoint](#entrypoint-type), and use of the
+[executable](#executable-type) as an entrypoint to the target [domain](#domain).
```
stephensmalley, line 40: when performing an exec-based domain transition. For a setcon-based dynamic domain transition, the transition requires the new type to be bounded by the old type if the process is multi-threaded, policy authorization for the source-to-target dynamic transition (dyntransition permission), and, if the process is being ptraced at the time of the setcon(3) call, policy authorization for the tracer-to-target ptrace permission.

stephensmalley: Technically all of this discussion applies to the entire context, not just the domain/type, since any of these permissions might be denied by constraints on user/role/level.

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+## Label
+
+See [Security context](#security-context).
+
+## Multi-Category Security (MCS)
+
+An optional SELinux policy configuration that uses unordered categories to separate
+[subjects](#subject) and [objects](#object). A [subject](#subject) has access to an
+[object](#object) if its category set dominates the category set of the object.  In SELinux, MCS
+only restricts access that [Type Enforcement](#type-enforcement-te) allows. It cannot allow access
```
stephensmalley: further restricts

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+on its own. Reference Policy enables MCS with `TYPE=mcs`.
+
+## Multi-Level Security (MLS)
+
+An optional SELinux policy configuration that implements the
+[Bell-LaPadula model](#bell-lapadula-model) with the [strong star property](#strong-star-property).
+In SELinux, MLS only restricts access that [Type Enforcement](#type-enforcement-te) allows. It
```
stephensmalley: further restricts

doc/GLOSSARY.md
```diff
@@ -0,0 +1,133 @@
...
+## Role
+
+The role component of an SELinux [security context](#security-context). This is used in the
+[Role-Based Access Control](#role-based-access-control-rbac) decisions.
+
+## Role-Based Access Control (RBAC)
+
+An access control model in which authorizes [roles](#role) for sets of [domain](#domain) types. It
+controls which domains an [SELinux user](#selinux-user-seuser) may enter and controls role changes.
+[Objects](#object) normally use the special `object_r` role. In SELinux, RBAC only restricts access
```
stephensmalley: further restricts

doc/GLOSSARY.md
stephensmalley: Additionally, since these checks fire on any context transition, it is often necessary to allow transition and/or dyntransition permissions between a domain and itself just to allow a change to the user/role/level fields.

stephensmalley: Not sure if you care about capturing this here, but setcon(3) and setexeccon(3) further require the corresponding permission (setcurrent and setexec respectively) between a domain and itself in order to set a context for a transition. This is NOT required for policy-defined type transitions.

doc/GLOSSARY.md
pebenito: Added a paragraph about this below.

doc/GLOSSARY.md
pebenito: Since this is a glossary, I want to limit detail in the descriptions, so I'm redirecting to the SELinux Notebook.

doc/SECURITY_GOALS.md
dburgener: I like your rewrite of this section a lot better.  I think it addresses my comment.

jbrindle (approved):

PR #1194, opened 2026-08-09
Improve incus and container modules

mschiff:
This PR
- improves compatibility with recent incus versions
- adds a tunable to allow nested containers
- adds some minor rule tweaks

### DIFF
--- a/policy/modules/kernel/kernel.if
+++ b/policy/modules/kernel/kernel.if
@@ -1207,6 +1207,24 @@ interface(`kernel_mounton_proc_dirs',`
 
 ########################################
 ## <summary>
+##	Mount on the files in /proc.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`kernel_mounton_proc_files',`
+	gen_require(`
+		type proc_t;
+	')
+
+	allow $1 proc_t:file mounton_file_perms;
+')
+
+########################################
+## <summary>
 ##	Get the attributes of files in /proc.
 ## </summary>
 ## <param name="domain">
@@ -1455,6 +1473,25 @@ interface(`kernel_dontaudit_getattr_core_if',`
 
 ########################################
 ## <summary>
+##	Mount on files of core kernel interface.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`kernel_mounton_core_if',`
+	gen_require(`
+		type proc_kcore_t;
+	')
+
+	allow $1 proc_kcore_t:file mounton_file_perms;
+')
+
+########################################
+## <summary>
 ##	Allows caller to read the core kernel interface.
 ## </summary>
 ## <param name="domain">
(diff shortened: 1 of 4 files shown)


policy/modules/services/container.te
```diff
@@ -649,7 +657,7 @@ optional_policy(`
 # Containers with additional permissions
 # required to run an init system
 
-allow container_init_t self:process { getcap setrlimit };
+allow container_init_t self:process { execmem execstack getcap setrlimit };
```
pebenito, line 660: This needs to be tunable.

policy/modules/services/incus.te
```diff
@@ -45,7 +45,7 @@ ifdef(`enable_mls',`
 allow incusd_t self:process getattr;
 dontaudit incusd_t self:capability sys_module;
 allow incusd_t self:anon_inode { create map read write };
-allow incusd_t self:io_uring sqpoll;
+allow incusd_t self:io_uring allowed;
```
pebenito: Why is the sqpoll dropped?

policy/modules/services/container.if
```diff
@@ -475,6 +475,24 @@ interface(`container_rw_system_engine_bpf',`
...
+## <summary>
+##	Ptrace container_init.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`container_ptrace_container_init',`
```
pebenito: `container_ptrace_init_container`. Please update the summary too.

policy/modules/kernel/kernel.if
```diff
@@ -2331,6 +2386,25 @@ interface(`kernel_dontaudit_read_kernel_sysctl',`
...
+##	Mount on files of core kernel interface.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+## <rolecap/>
+#
+interface(`kernel_mounton_core_files_if',`
```
pebenito: This should be with the other `proc_kcore_t` interfaces. Also drop the `files` from the interface name, for consistency with the existing interfaces.

policy/modules/kernel/kernel.if
```diff
@@ -1205,6 +1205,42 @@ interface(`kernel_mounton_proc_dirs',`
...
+## </param>
+#
+interface(`kernel_mounton_proc_psi_dirs',`
+	gen_require(`
+		type proc_psi_t;
+	')
+
+	allow $1 proc_psi_t:dir mounton_dir_perms;
+')
+
```
pebenito: This should also go with the other proc_psi_t interfaces.

pebenito (changes requested):

policy/modules/services/container.te
mschiff: I think `execmem` would be enough for now. The other container types like `container_t`, `container_kvm_t`, `spc_t`, `spc_user_t` all already have that rule so I thought it would be ok.

You still think should better be tunable?

policy/modules/services/container.if
mschiff: Fixed

policy/modules/services/container.te
pebenito: No, I forgot the other containers have it.

pebenito (approved):

PR #1196, opened 2026-08-10
pipe2

etbe:


policy/modules/apps/pipewire.fc
```diff
@@ -9,6 +9,8 @@
 
 # System-service runtime directory (/run/pipewire).
 /run/pipewire(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
+/run/user/%{USERID}/pipewire-.*	-s	gen_context(system_u:object_r:pipewire_runtime_t,s0)
+#/run/user/%{USERID}/pulse(/.*)?		gen_context(system_u:object_r:pipewire_runtime_t,s0)
```
pebenito: Please uncomment or remove (based on what the intent is)

policy/modules/apps/pipewire.fc
```diff
@@ -18,4 +20,5 @@
 
 # User home configuration and state.
 HOME_DIR/\.config/pipewire(/.*)?	gen_context(system_u:object_r:pipewire_home_t,s0)
+#HOME_DIR/\.config/pulse(/.*)?		gen_context(system_u:object_r:pipewire_home_t,s0)
```
pebenito: Please uncomment or remove (based on what the intent is)

policy/modules/apps/pipewire.if
```diff
@@ -111,6 +111,44 @@ interface(`pipewire_domtrans_client',`
...
+##	</summary>
+## </param>
+#
+interface(`pipewire_create_user_dirs',`
+	gen_require(`
+		type pipewire_runtime_t;
+	')
+
+	userdom_user_runtime_filetrans($1, pipewire_runtime_t, dir, "pulse")
+	allow $1 pipewire_runtime_t:dir manage_dir_perms;
```
pebenito, line 106: The interface name has `create`, but `manage` is allowed here. Please update this or the interface name for consistency.

Also please split out the filetrans to its own interface.

policy/modules/apps/pipewire.if
```diff
@@ -111,6 +111,44 @@ interface(`pipewire_domtrans_client',`
...
+## <summary>
+##	unlink pipewire user socket /run/user/$UID/pipewire-0
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_socket_user_unlink',`
```
pebenito: `pipewire_delete_user_sockets

policy/modules/apps/pipewire.if
```diff
@@ -224,3 +262,24 @@ interface(`pipewire_read_home_files',`
...
+## <summary>
+##	Allow full pipewire access to the main client facilities
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`pipewire_full_client',`
```
pebenito: I'm a bit confused on how this is supposed to work vs `pipewire_client_t`. The "full" client may also be introducing confusion: what is a "partial" client?.

policy/modules/apps/wireplumber.te
```diff
@@ -42,13 +42,13 @@ userdom_user_tmpfs_file(wireplumber_tmpfs_t)
...
 allow wireplumber_t self:capability sys_nice;
-allow wireplumber_t self:process { getsched setsched signal signull };
+allow wireplumber_t self:process { getsched setsched signal signull setrlimit };
 allow wireplumber_t self:fifo_file rw_fifo_file_perms;
 allow wireplumber_t self:unix_stream_socket { create_stream_socket_perms connectto };
 allow wireplumber_t self:unix_dgram_socket create_socket_perms;
 
 # Bluetooth socket for BlueZ audio routing (A2DP/HFP profile management).
-allow wireplumber_t self:bluetooth_socket create_socket_perms;
+allow wireplumber_t self:bluetooth_socket { listen create_socket_perms };
```
pebenito: this should go up to `create_stream_socket_perms` to handle the probable `accept` perm too.

policy/modules/apps/wireplumber.te
```diff
@@ -93,11 +97,25 @@ miscfiles_read_localization(wireplumber_t)
 udev_read_runtime_files(wireplumber_t)
 udev_search_runtime(wireplumber_t)
 
+xdg_search_config_dirs(wireplumber_t)
+xdg_search_data_dirs(wireplumber_t)
+xdg_manage_data(wireplumber_t)
```
pebenito: The search call shouldn't be necessary with the manage

policy/modules/system/systemd.if
```diff
@@ -252,6 +252,11 @@ template(`systemd_role_template',`
 		pulseaudio_manage_tmp_dirs($1_systemd_t)
 	')
 
+	optional_policy(`
+		pipewire_create_user_dirs($1_systemd_t)
+		pipewire_socket_user_unlink($1_systemd_t)
+	')
```
pebenito, line 257: This might not be necessary after you made `pipewire_runtime_t` user runtime content. If the fc entries are in place, the filetrans may also be unnecessary.

policy/modules/system/systemd.te
```diff
@@ -2558,6 +2558,10 @@ optional_policy(`
 	gpg_delete_agent_named_sockets(systemd_user_runtime_dir_t)
 ')
 
+optional_policy(`
+	pipewire_socket_user_unlink(systemd_user_runtime_dir_t)
```
pebenito: May not be necessary with `pipewire_runtime_t` being user runtime content.

policy/modules/system/userdomain.if
```diff
@@ -3674,6 +3678,25 @@ interface(`userdom_search_user_runtime',`
...
+## <summary>
+##	rw users runtime directories.
+## </summary>
+## <param name="domain">
+##	<summary>
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`userdom_rw_user_runtime',`
```
pebenito: `userdom_rw_user_runtime_dirs()`

pebenito (changes requested):

pebenito:
May also overlap #1171

policy/modules/apps/pipewire.if
etbe: pipewire_client_t is a domain for certain specific pipewire test and utility programs. I don't think it deserves a domain but I didn't make that decision.

The pipewire_full_client() interface gives all the access needed for talking to pipewire.

policy/modules/system/systemd.if
etbe: type=AVC msg=audit(1786533550.791:64432): avc:  denied  { create } for  pid=323711 comm="systemd" name="pulse" scontext=user_u:user_r:user_systemd_t:s0 tcontext=user_u:object_r:pipewire_runtime_t:s0 tclass=dir permissive=0

I tested and allowing creation it is necessary but the filetrans and unlink aren't.

policy/modules/apps/pipewire.if
pebenito: What do you propose for those utilities instead of `pipewire_client_t`?

etbe: I think that labelling the files as bin_t and having them run as user_t would be fine. But that could be in a future patch.

pebenito: That sounds good. With that in mind, I'd prefer to group that change with this one so we can comfortably drop "full" from the interface name.

etbe:
@pebenito OK I made the client changes requested and also copied a few lines from #1171.

pebenito (approved):

PR #1198, opened 2026-08-13
docportal3

etbe:


pebenito:
I'd appreciate if you wouldn't close then open new PRs for the same changes. That makes it more difficult to review since the previous comments don't show here.

etbe:
> I'd appreciate if you wouldn't close then open new PRs for the same changes. That makes it more difficult to review since the previous comments don't show here.

ok I'll try to avoid that in future

policy/modules/apps/docportal.if
```diff
@@ -0,0 +1,79 @@
...
+
+		allow $1_docportal_t $2:dbus send_msg;
+		allow $2 $1_docportal_t:dbus send_msg;
+		dbus_all_session_bus_client($1_docportal_t)
+	')
+')
+
+########################################
+## <summary>
+##	Execute a domain transition to run WirePlumber.
```
pebenito: Copy-paste error

policy/modules/system/userdomain.if
```diff
@@ -3062,6 +3062,27 @@ interface(`userdom_list_user_tmp',`
...
+##	Domain allowed access.
+##	</summary>
+## </param>
+#
+interface(`userdom_mounton_user_tmp',`
+	gen_require(`
+		type user_tmp_t, user_runtime_t;
+	')
+
+	allow $1 user_tmp_t:dir { mounton write };
```
pebenito: This `write` is out of place. Is this for checking if the location is writable? If so, we should have a `userdom_check_writable_user_tmp` for this, since there is no `add_name` nor `remove_name`.

pebenito (changes requested):

policy/modules/system/userdomain.if
etbe: OK I'll remove it for now. I will add it back in a different interface later if needed.

pebenito (approved):

PR #1202, opened 2026-08-18
systemd-nspawn-passwd-agent

etbe:


pebenito (commented):
There isn't a transition from sysadm to systemd-nspawn, so this seems incomplete.

PR #1203, opened 2026-08-18
policy/modules/roles/sysadm.te: grant sysadm_t to check SELinux booleans

ChenQi1989:
For sysadm, it would be useful to use getsebool to check SELinux settings.

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -1513,6 +1513,8 @@ template(`userdom_admin_user_template',`
 	# But presently necessary for installing the file_contexts file.
 	seutil_manage_bin_policy($1_t)
 
+	selinux_get_all_booleans($1_t)
+
 	userdom_manage_user_home_content_dirs($1_t)
 	userdom_manage_user_home_content_files($1_t)
 	userdom_manage_user_home_content_symlinks($1_t)


pebenito (changes requested):
This should go in `userdom_security_admin_template()` as that's where the permission to set all Booleans is.

ChenQi1989:
@pebenito Thanks for your suggestion.

I checked the configurations.

The userdom_security_admin_template in policy/modules/system/userdomain.if has already got:
`selinux_set_all_booleans($1)`

The policy/modules/roles/sysadm.te has:
```
ifndef(`enable_mls',`
        userdom_security_admin_template(sysadm_t, sysadm_r)
')
```

This means if mls is not used, sysadm_t can get/set selinux booleans; but if mls is used, sysadm_t cannot do anything.

What I want is sysadm_t can get selinux booleans even in mls. 

What do you think?

pebenito:
Sorry, I meant `userdom_admin_user_template()`.

ChenQi1989:
> Sorry, I meant `userdom_admin_user_template()`.

Got it. Thanks. I've updated this PR according to your suggestion. @pebenito

pebenito (approved):

PR #1207, opened 2026-08-19
kernel: allow module loaders to use net_admin

wenwfu:
The qrtr and qrtr_smd kernel modules register the PF_QIPCRTR network protocol family during module initialization. This requires CAP_NET_ADMIN, but systemd-modules-load runs in systemd_modules_load_t without that capability.

When SELinux is enforcing, systemd-modules-load runs in Loading qrtr modules from modules-load.d fails with EPERM, and the modules are  qrtr modules from modules-load.d fails with EPERM, and the modules are loaded later through modalias autoloading.

Allow systemd-modules-load to use CAP_NET_ADMIN so qrtr modules can be preloaded during boot. This applies to any system using SELinux and building qrtr as loadable kernel modules. The denial is hidden by dontaudit rules and can be observed with semodule -DB:

```
  avc: denied { net_admin } for comm="systemd-modules" capability=12
  scontext=system_u:system_r:systemd_modules_load_t:s0
  tcontext=system_u:system_r:systemd_modules_load_t:s0 tclass=capability
```

### DIFF
--- a/policy/modules/kernel/kernel.te
+++ b/policy/modules/kernel/kernel.te
@@ -16,6 +16,14 @@ gen_require(`
 ## </desc>
 gen_bool(secure_mode_insmod, false)
 
+## <desc>
+## <p>
+## Allow domains that can load kernel modules to use the net_admin
+## capability during module initialization.
+## </p>
+## </desc>
+gen_tunable(kernel_module_load_net_admin, false)
+
 # assertion related attributes
 attribute can_load_kernmodule;
 attribute can_receive_kernel_messages;
@@ -589,6 +597,10 @@ if(secure_mode_insmod) {
 	files_load_kernel_modules(can_load_kernmodule)
 
 	kernel_search_key(can_load_kernmodule)
+}
+
+if (!secure_mode_insmod && kernel_module_load_net_admin) {
+	allow can_load_kernmodule self:capability net_admin;
 }
 
 ########################################
--- a/testing/sechecker.ini
+++ b/testing/sechecker.ini
@@ -245,6 +245,7 @@ exempt_source = arpwatch_t
                 blueman_t
                 bluetooth_t
                 brctl_t
+                can_load_kernmodule # Conditional access (kernel_module_load_net_admin)
                 cgred_t
                 chronyd_t           # Conditional access (chronyd_hwtimestamp)
                 condor_startd_t


dsugar100:
I will of course defer to @pebenito, but maybe this could be in a boolean?  Seems like most won't need to allow systemd-modules to have net_admin capability.

wenwfu:
> I will of course defer to @pebenito, but maybe this could be in a boolean? Seems like most won't need to allow systemd-modules to have net_admin capability.

Thanks, good point. I updated the PR to gate CAP_NET_ADMIN behind the disabled-by-default systemd_modules_load_net_admin tunable. Systems that do not need it retain the existing restrictions. Please take another look.

pebenito:
I agree it should be tunable; however, there are other potential module loading domains that would still break. I think we probably should put it here: https://github.com/SELinuxProject/refpolicy/blob/f2562d4884a4a8c2015c5436ea3b37adb6d8c43f/policy/modules/kernel/kernel.te#L578

I'd like to hear others thoughts on this.

wenwfu:
> I agree it should be tunable; however, there are other potential module loading domains that would still break. I think we probably should put it here:
> 
> https://github.com/SELinuxProject/refpolicy/blob/f2562d4884a4a8c2015c5436ea3b37adb6d8c43f/policy/modules/kernel/kernel.te#L578
> 
> I'd like to hear others thoughts on this.

That makes sense. Since can_load_kernmodule already centralizes the module-loading policy for init, udev, kmod, systemd-modules-load, and container domains, I’ll move this to kernel.te and rename the tunable to a generic kernel_module_load_net_admin, disabled by default.

policy/modules/kernel/kernel.te
```diff
@@ -586,6 +594,10 @@ if(secure_mode_insmod) {
 	allow can_load_kernmodule self:capability sys_module;
 	allow can_load_kernmodule self:system module_load;
 
+	tunable_policy(`kernel_module_load_net_admin',`
+		allow can_load_kernmodule self:capability net_admin;
+	')
```
pebenito: The compiler doesn't support nesting like this. You'll need to make an `if (!secure_mode_insmod && kernel_module_load_net_admin)` block.

testing/sechecker.ini
```diff
@@ -323,6 +323,7 @@ exempt_source = arpwatch_t
                 system_cronjob_t
                 system_munin_plugin_t
                 systemd_cgroups_t
+                systemd_modules_load_t # Conditional access (kernel_module_load_net_admin)
```
pebenito: This would change to `can_load_kernmodule` (with updated sorting)

policy/modules/kernel/kernel.te
wenwfu: Updated the rule to use a combined conditional: if (!secure_mode_insmod && kernel_module_load_net_admin).

testing/sechecker.ini
wenwfu: Replaced systemd_modules_load_t with can_load_kernmodule and updated the sorting.

pebenito (approved):

PR #1215, opened 2026-08-27
Fix login error about lacking anon_inodefs on kernel 6.6

ChenQi1989:
For kernel < 6.9, pidfs is not supported. login via pam_systemd chain will give us avc denial error.

Version info about the related softwares:
kernel : 6.6
systemd : 259.5
shadow : 4.19.4

### DIFF
--- a/policy/modules/system/locallogin.te
+++ b/policy/modules/system/locallogin.te
@@ -9,6 +9,7 @@ type local_login_t;
 domain_interactive_fd(local_login_t)
 auth_login_pgm_domain(local_login_t)
 auth_login_entry_type(local_login_t)
+fs_getattr_all_fs(local_login_t)
 
 type local_login_lock_t;
 files_lock_file(local_login_lock_t)
--- a/policy/modules/system/systemd.if
+++ b/policy/modules/system/systemd.if
@@ -60,6 +60,7 @@ template(`systemd_role_template',`
 	# This domain is per-role because of the below transitions.
 	# See the systemd --user section of systemd.te for the
 	# remainder of the rules.
+	fs_getattr_all_fs($1_systemd_t)
 	allow $1_systemd_t self:process { getsched signal };
 	allow $1_systemd_t self:netlink_kobject_uevent_socket create_socket_perms;
 	allow $1_systemd_t self:netlink_route_socket r_netlink_socket_perms;


pebenito:
I don't see any issues with the rules, but I'm reluctant to merge this because we already know they're no longer needed.

ChenQi1989:
@pebenito Sorry but I don't quite understand what you meant by "no longer needed". New version systemd running on kernel < 6.9 can reveal the error. Do you mean that "kernel < 6.9  + new version systemd" is not in the supporting scope of refpolicy? 

P.S.
I haven't tried using new kernel versions, so I don't know if pidfs will also give us a similar error. Looking at refpolicy rules, there doesn't seem to a rule granting pidfs specifically.

pebenito:
My understand of what you're saying was that this access isn't needed on kernel >= 6.9. If that's not the case, then I will reconsider.

ChenQi1989:
@pebenito I did not test kernel >= 6.9, so I had to say kernel < 6.9 in the commit message. That seems to cause some misunderstanding. 
I'll use a new kernel and do the test. I'll let you know the result ASAP.

ChenQi1989:
@pebenito I can confirm that new kernel does NOT have the problem. There's no AVC message.

My kernel version is 6.18.39. systemd version is 259.5
```text
root@qemuarm64:~# ausearch -m avc
<no matches>
root@qemuarm64:~# uname -a
Linux qemuarm64 6.18.39-yocto-standard #1 SMP PREEMPT Wed Jul 22 16:47:36 UTC 2026 aarch64 GNU/Linux
root@qemuarm64:~# systemctl --version
systemd 259 (259.5)
+PAM +AUDIT +SELINUX -APPARMOR +IMA +IPE -SMACK +SECCOMP -GCRYPT -GNUTLS +OPENSSL +ACL +BLKID -CURL -ELFUTILS -FIDO2 -IDN2 -IDN +KMOD -LIBCRYPTSETUP -LIB
CRYPTSETUP_PLUGINS +LIBFDISK -PCRE2 -PWQUALITY -P11KIT -QRENCODE -TPM2 -BZIP2 -LZ4 -XZ -ZLIB +ZSTD -BPF_FRAMEWORK -BTF +XKBCOMMON +UTMP -SYSVINIT -LIBARC
HIVE
```

The pidfs support seems to be there via 2ade34d9ba8b9c5697d6cfc0219a5801e56f1cd6

pebenito:
I will have to NAK this. If there is more demand for this in the future, I'll reconsider.

PR #1220, opened 2026-08-30
Add configfs_t and tracefs_t to mountpoint attribute

henke9600:
Noticed that these mount points aren't covered by `files_search_all_mountpoints()`. I assume this isn't intentional.

### DIFF
--- a/policy/modules/kernel/filesystem.te
+++ b/policy/modules/kernel/filesystem.te
@@ -108,6 +108,7 @@ dev_associate_sysfs(memory_pressure_t)
 
 type configfs_t;
 fs_type(configfs_t)
+files_mountpoint(configfs_t)
 genfscon configfs / gen_context(system_u:object_r:configfs_t,s0)
 
 type cpusetfs_t;
@@ -200,6 +201,7 @@ genfscon v7 / gen_context(system_u:object_r:sysv_t,s0)
 
 type tracefs_t;
 fs_type(tracefs_t)
+files_mountpoint(tracefs_t)
 genfscon tracefs / gen_context(system_u:object_r:tracefs_t,s0)
 
 optional_policy(`


pebenito:
What are some examples where you mount over configfs or tracefs? I'm not aware of any; if there are none, then they are not mountpoints.

henke9600:
Not sure i fully understand the question, but on my Debian system they are mount points:
```
$ mount | grep configfs
configfs on /sys/kernel/config type configfs (rw,nosuid,nodev,noexec,relatime)
$ mount | grep tracefs
tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
```

I ran into this when creating a custom policy for a program that searches all mount points in the system. Adding `files_search_all_mountpoints()` solved all of them except for these two mounts. Since debugfs is also under /sys/kernel/ and it had the mountpoint attribute, i assumed the same should apply to configfs and tracefs.

pebenito:
> Not sure i fully understand the question, but on my Debian system they are mount points:
> 
> ```
> $ mount | grep configfs
> configfs on /sys/kernel/config type configfs (rw,nosuid,nodev,noexec,relatime)
> $ mount | grep tracefs
> tracefs on /sys/kernel/tracing type tracefs (rw,nosuid,nodev,noexec,relatime)
> ```

The types that `files_mountpoint()` are ones that are mounted over. For example, the location ` /sys/kernel/config` is wheere configfs mounts over sysfs. The mountpoint type is `sysfs_t`, not `configfs_t`.

With that in mind, this still seems incorrect.

henke9600:
If this is incorrect, then there's an apparent inconsistency that i don't understand:
1. `/sys/kernel/debug` is a mount point, labeled `debugfs_t` and has `files_mountpoint(debugfs_t)`
2. `/sys/kernel/tracing` is a mount point, labeled `tracefs_t` but does not have `files_mountpoint(tracefs_t)`

Why would they be treated differently?

I also did some googling just now and found this similar commit:
https://github.com/fedora-selinux/selinux-policy/commit/185631685b46f283b4ab9d23e20b1d828f45a1b7

pebenito:
> If this is incorrect, then there's an apparent inconsistency that i don't understand:
> 
> 1. `/sys/kernel/debug` is a mount point, labeled `debugfs_t` and has `files_mountpoint(debugfs_t)`
> 2. `/sys/kernel/tracing` is a mount point, labeled `tracefs_t` but does not have `files_mountpoint(tracefs_t)`
> 
> Why would they be treated differently?

The first one is almost certainly a bug.

henke9600:
Ah, i understand what you meant now.

So `files_mountpoint(debugfs_t)` probably comes from the fact that tracefs used to be mounted at `/sys/kernel/debug/tracing`, before it was moved in 2015 to `/sys/kernel/tracing`. Seems like it can still be enabled with `CONFIG_TRACEFS_AUTOMOUNT_DEPRECATED`, but this is evidently deprecated.

If you want i can open a PR to remove `files_mountpoint(debugfs_t)`. Anyway, i'll close this one.
Thanks for the help!

pebenito:
> If you want i can open a PR to remove `files_mountpoint(debugfs_t)`.

Yes, please.

PR #1223, opened 2026-09-04
del1

etbe:


policy/modules/services/hddtemp.fc
```diff
@@ -1,6 +0,0 @@
-/etc/rc\.d/init\.d/hddtemp	--	gen_context(system_u:object_r:hddtemp_initrc_exec_t,s0)
-
-/etc/sysconfig/hddtemp	--	gen_context(system_u:object_r:hddtemp_etc_t,s0)
```
thesamesam, line 3: We still have this packaged in Gentoo as `app-admin/hddtemp`.

etbe: https://github.com/guzu/hddtemp

I can only find this for upstream source, which hasn't been updated in 16  years. Do you know of one that is being maintained?

thesamesam: It looks like we use https://savannah.nongnu.org/projects/hddtemp/ + Debian's old patches. Someone proposed we switch to a new upstream at https://github.com/vitlav/hddtemp but we haven't done that yet, and in https://bugs.gentoo.org/933749, someone else said we should just remove the package

etbe: If you change to a new upstream and keep supporting it then we could add it in. But for the moment I think we should just remove it. @pebenito ?

pebenito: See https://github.com/SELinuxProject/refpolicy/pull/1224#discussion_r3958337153.

The bug you mention says the Gentoo package doesn't even have a maintainer, so I'm much more inclined to push this module down to Gentoo.

PR #1224, opened 2026-09-04
del2

etbe:


policy/modules/services/openct.fc
```diff
@@ -1,7 +0,0 @@
-/etc/rc\.d/init\.d/openct	--	gen_context(system_u:object_r:openct_initrc_exec_t,s0)
-
-/usr/bin/ifdhandler	--	gen_context(system_u:object_r:openct_exec_t,s0)
```
thesamesam, line 3: We still have it packaged in Gentoo as `dev-libs/openct`.

etbe: What is the upstream you have for it?

etbe: Is there any hardware from the last 10 years that is supported by it and not by other software?

thesamesam: It just uses https://github.com/OpenSC/openct/ still. I have no idea, but I will ask a friend who knows about smartcards

thesamesam: ok, he says it's dead. personally I would rather policies only go when the main distros using the policy no longer use them, rather than the other way around though

etbe: As it's dead the best way of addressing this is to just remove it from Gentoo. Having obsolete software in Gentoo that's not usable isn't helping users.

pebenito: thesamesam said:
> personally I would rather policies only go when the main distros using the policy no longer use them, rather than the other way around though

etbe said:
> As it's dead the best way of addressing this is to just remove it from Gentoo. Having obsolete software in Gentoo that's not usable isn't helping users.

I think these are both very compelling points. However, I don't think it's the business of the upstream policy to try to force the hand of officially supported distros. I think we should keep it. For now.

@thesamesam please keep an eye on this and see if Gentoo will remove it. There is no official policy for Refpolicy's support of dead projects and this situation is a good reason to define one. Arguably, the module should go downstream to Gentoo's repo if they are keeping openct alive, despite the upstream project being dead.

PR #1226, opened 2026-09-06
Fix missing allow_ptrace and sysadm_allow_rw_inherited_fifo booleans

henke9600:
They were missing from `policy/booleans.conf` due to the following warning: support/segenxml.py: WARNING: cannot open file doc/tmp/iftemplates/userdom_security_admin_template.iftemplate.  Ignoring.

This is because userdom_security_admin_template has the `_template` suffix, but isn't actually a template.

### DIFF
--- a/policy/modules/system/userdomain.if
+++ b/policy/modules/system/userdomain.if
@@ -1561,7 +1561,7 @@ template(`userdom_admin_user_template',`
 ##	</summary>
 ## </param>
 #
-interface(`userdom_security_admin_template',`
+template(`userdom_security_admin_template',`
 	allow $1 self:capability { dac_override dac_read_search };
 	allow $1 self:capability2 mac_admin;
 


policy/modules/system/userdomain.if
```diff
@@ -1561,7 +1561,7 @@ template(`userdom_admin_user_template',`
 ##	</summary>
 ## </param>
 #
-interface(`userdom_security_admin_template',`
+interface(`userdom_security_admin',`
```
pebenito, line 1564: Do not change the existing names, as that's a breaking change.

pebenito (changes requested):
This does not result in a rule change. See the diff stages of the GitHub actions.

policy/modules/system/userdomain.if
henke9600: I see.

I changed it back to a template instead, is that appropriate?
It was previously changed from template to interface here: https://github.com/SELinuxProject/refpolicy/commit/6c5928d65a340fe423b9b163dfe1b312a1c5fd5b

pebenito: Sorry, I got myself confused since the XML docs are out of date. Basically, if a parameter is a prefix (e.g `sysadm`) then it's a template. The docs say it is in the `<desc>`, but that is wrong. It should be an interface since a type and a role are the parameters. We can also do the rename, but there will need to be a backward compatibility interface so it's not a breaking change. See `systemd_read_user_unit_files()` for an example of deprecating an interface.

henke9600: Done. I also tried to remove some misleading text for the docs, but wasn't fully sure what to write there instead.

pebenito (approved):

PR #1231, opened 2026-09-09
tee_supplicant: Allow qtee_supplicant to access the eMMC RPMB device

Pradeep-pvk:
Add SELinux allow rules for qtee_supplicant to access the eMMC RPMB device, enabling communication with the RPMB partition on eMMC-based systems.

Define a new storage_raw_rw_removable_device() interface to storage.if to allow read or write on removable chr_file interfaces, and use it in tee_supplicant.te to grant qtee_supplicant the necessary chr_file rw_chr_file_perms on removable_device_t.

### DIFF
--- a/policy/modules/services/tee_supplicant.te
+++ b/policy/modules/services/tee_supplicant.te
@@ -34,6 +34,13 @@ dev_rw_tee_priv(tee_supplicant_t)
 
 kernel_read_vm_overcommit_sysctl(tee_supplicant_t)
 
+# Allow qtee_supplicant to access the eMMC RPMB device (/dev/mmcblk0rpmb)
+storage_raw_rw_removable_device(tee_supplicant_t)
+
+# Allow qtee_supplicant to search directories labeled default_t
+# (e.g., the "data" directory on mmcblk0pxx)
+files_search_default(tee_supplicant_t)
+
 # Access qtee_supplicant to access UFS BSG device
 storage_read_scsi_generic_cond(tee_supplicant_t,tee_supplicant_qtee)
 storage_write_scsi_generic_cond(tee_supplicant_t,tee_supplicant_qtee)
--- a/policy/modules/kernel/storage.if
+++ b/policy/modules/kernel/storage.if
@@ -912,6 +912,24 @@ interface(`storage_dontaudit_raw_write_removable_device',`
 
 ########################################
 ## <summary>
+#	read or write on removable device interfaces.
+## </summary>
+## <param name="domain">
+##     <summary>
+##     Domain allowed access.
+##     </summary>
+## </param>
+#
+interface(`storage_raw_rw_removable_device',`
+	gen_require(`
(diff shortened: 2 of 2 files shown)


policy/modules/services/tee_supplicant.te
```diff
@@ -34,6 +34,13 @@ dev_rw_tee_priv(tee_supplicant_t)
 
 kernel_read_vm_overcommit_sysctl(tee_supplicant_t)
 
+# Allow qtee_supplicant to access the eMMC RPMB device (/dev/mmcblk0rpmb)
```
dburgener: Can we put all of these down in the qtee block? I don't think any of these are needed in the optee case.

policy/modules/services/tee_supplicant.te
```diff
@@ -34,6 +34,13 @@ dev_rw_tee_priv(tee_supplicant_t)
...
+storage_raw_rw_removable_device(tee_supplicant_t)
```
pebenito, line 38: This should go in the `tee_supplicant_qtee` block below.

policy/modules/services/tee_supplicant.te
```diff
@@ -34,6 +34,13 @@ dev_rw_tee_priv(tee_supplicant_t)
...
+
+# Allow qtee_supplicant to search directories labeled default_t
+# (e.g., the "data" directory on mmcblk0pxx)
+files_search_default(tee_supplicant_t)
```
pebenito, line 42: This indicates a labeling problem that should be fixed instead.

pebenito (changes requested):

policy/modules/services/tee_supplicant.te
Pradeep-pvk: sure, i will move it to block below.

policy/modules/services/tee_supplicant.te
Pradeep-pvk: ok, for now, i will remove this change and make it as a separate change as i don't see any functional failure without it.

policy/modules/services/tee_supplicant.te
```diff
@@ -63,4 +63,7 @@ tunable_policy(`tee_supplicant_qtee',`
 
         # Access qtee_supplicant to send logs to systemd journal
         logging_send_syslog_msg(tee_supplicant_t)
+
+	# Allow qtee_supplicant to access the eMMC RPMB device (/dev/mmcblk0rpmb)
+	storage_raw_rw_removable_device(tee_supplicant_t)
```
pebenito: nit: Please adjust the indentation.

Otherwise looks good.

pebenito (changes requested):

policy/modules/services/tee_supplicant.te
Pradeep-pvk: sure.

pebenito (approved):

PR #1248, opened 2026-09-18
del8

etbe:


policy/modules/admin/usbmodules.fc
```diff
@@ -1,2 +0,0 @@
-/usr/bin/usbmodules	--	gen_context(system_u:object_r:usbmodules_exec_t,s0)
```
thesamesam, line 1: It was part of usbutils a long time ago: https://github.com/gregkh/usbutils/commit/d7b2d878134f68f961b34c980cefeb77b9b8fb0c

Indeed obsolete.

policy/modules/roles/sysadm.te
```diff
@@ -1205,10 +1204,6 @@ optional_policy(`
 	watchdog_admin(sysadm_t, sysadm_r)
 ')
 
-optional_policy(`
-	wdmd_admin(sysadm_t, sysadm_r)
```
thesamesam, line 1209: wdmd is part of sanlock in gentoo, I dunno anything else about it though. @mschiff?

policy/modules/admin/usbmodules.fc
etbe: thanks

policy/modules/roles/sysadm.te
etbe: Is there an upstream repository for it?

thesamesam: https://codeberg.org/sanlock/sanlock/src/branch/master/wdmd

mschiff: > wdmd is part of sanlock in gentoo, I dunno anything else about it though. @mschiff?

unfortunately I am not running sanlock anywhere currently and never did on a box with selinux enable, so I fear I cannot help here

PR #1250, opened 2026-09-18
daemon-fc

etbe:


policy/modules/services/acpi.fc
```diff
@@ -1,8 +1,6 @@
 /etc/rc\.d/init\.d/acpid	--	gen_context(system_u:object_r:acpid_initrc_exec_t,s0)
 
 /usr/bin/acpid	--	gen_context(system_u:object_r:acpid_exec_t,s0)
-/usr/bin/apm	--	gen_context(system_u:object_r:acpi_exec_t,s0)
```
thesamesam, line 4: We still package it but I suspect nobody uses it, so fine.

policy/modules/services/apache.fc
```diff
@@ -5,7 +5,6 @@ HOME_DIR/((www)|(web)|(public_html))(/.*)?/logs(/.*)?		gen_context(system_u:obje
 
 /etc/apache(2)?(/.*)?						gen_context(system_u:object_r:httpd_config_t,s0)
 /etc/apache-ssl(2)?(/.*)?					gen_context(system_u:object_r:httpd_config_t,s0)
-/etc/cherokee(/.*)?						gen_context(system_u:object_r:httpd_config_t,s0)
```
thesamesam, line 8: Last release in 2014, we last-rited it in gentoo a few years ago too: ok

policy/modules/services/cpucontrol.fc
```diff
@@ -2,6 +2,7 @@
 
 /usr/bin/cpufreqd	--	gen_context(system_u:object_r:cpuspeed_exec_t,s0)
 /usr/bin/cpuspeed	--	gen_context(system_u:object_r:cpuspeed_exec_t,s0)
+/usr/bin/iucode_tool	--	gen_context(system_u:object_r:cpucontrol_exec_t,s0)
```
thesamesam, line 5: it's not a daemon :p

could you try split things more finely? `git add -p` makes it easy if not using that already

policy/modules/services/policykit.fc
```diff
@@ -23,4 +23,8 @@
 /var/lib/polkit-1(/.*)?	gen_context(system_u:object_r:policykit_var_lib_t,s0)
 /var/lib/PolicyKit-public(/.*)?	gen_context(system_u:object_r:policykit_var_lib_t,s0)
 
+ifdef(`distro_debian',`
```
thesamesam, line 26: This should be its own change with some explanation of how it's Debian specific

policy/modules/services/acpi.fc
etbe: It's been over 25 years since they stopped manufacturing laptops with APM support. APM was on the way out when SE Linux was first developed. A quick google suggests that apm driver code is no longer in the Linux kernel.

policy/modules/services/apache.fc
etbe: thanks

policy/modules/services/cpucontrol.fc
etbe: But still the correct context.

policy/modules/services/policykit.fc
etbe: Debian uses a different path to what other Refpolicy based distributions apparently use.

policy/modules/services/acpi.fc
thesamesam: ok, will kill the pkg too

